PR #27, opened 2020-09-04
Update Dockerfile to Airflow v1.10.10

johnclary:
@sergiogcx i ran into [this issue](https://github.com/puckel/docker-airflow/issues/535) just now while trying to run locally. Looks like we need to either use `SQLAlchemy==1.3.15` in the Dockerfile or use Airflow v1.10.10.

I did the latter here and it worked for me locally.


sergiogcx:
@johnclary sweet it works on my end! 👍

sergiogcx (approved):
Thanks for fixing it!

PR #32, opened 2020-11-30
Program to extract CR3 diagrams & OCR the narrative

frankhereford:
This PR proposes to merge the CR3 extraction tool into master. It currently detects scanned CR3 documents and skips them, as well as has the changes to address the thoughtful feedback provided by @sergiogcx.

### DIFF
--- a/dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
+++ b/dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
@@ -1,0 +1,136 @@
+#!/usr/bin/python3
+
+import io
+import os
+import re
+import sys
+import json
+import boto3
+import argparse
+import psycopg2
+from   pdf2image import convert_from_path, convert_from_bytes
+import pytesseract
+
+# Configure the available arguments to the program
+parser = argparse.ArgumentParser(description='Extract CR3 diagrams and narratives')
+parser.add_argument('-v', action='store_true', help='Be verbose')
+parser.add_argument('--cr3-source', metavar=('bucket', 'path'), nargs=2, required=True, help='Where can we hope to find CR3 files on S3?')
+parser.add_argument('--batch-size', metavar='int', required=True, help='How many cr3s to attempt to process?')
+parser.add_argument('--update-narrative', action='store_true', help='Update narrative in database')
+parser.add_argument('--update-timestamp', action='store_true', help='Update timestamp in database')
+parser.add_argument('--save-diagram', metavar=('bucket', 'path'), nargs=2, help='Save diagram PNG in a S3 bucket and path')
+parser.add_argument('--crash-id', metavar='int', type=int, nargs=1, help='Specific crash ID to operate on')
+parser.add_argument('--db-host', metavar='string', nargs=1, required=True, help='DB: host')
+parser.add_argument('--db-database', metavar='string', nargs=1, required=True, help='DB: database name')
+parser.add_argument('--db-username', metavar='string', nargs=1, required=True, help='DB: username')
+parser.add_argument('--db-password', metavar='string', nargs=1, required=True, help='DB: password')
+args = parser.parse_args()
+
+s3 = boto3.client('s3')
+
+# connect to a postgres database expecting atd_txdot_crashes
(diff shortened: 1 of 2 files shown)


dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
```diff
@@ -0,0 +1,247 @@
...
+parser = argparse.ArgumentParser(description='Extract CR3 diagrams and narratives')
+parser.add_argument('-v', action='store_true', help='Be verbose')
+parser.add_argument('-d', action='store_true', help='Check for digitally created PDFs')
+parser.add_argument('--cr3-source', metavar=('bucket', 'path'), nargs=2, required=True, help='Where can we hope to find CR3 files on S3?')
+parser.add_argument('--batch-size', metavar='int', default=1, help='How many cr3s to attempt to process?')
+parser.add_argument('--update-narrative', action='store_true', help='Update narrative in database')
+parser.add_argument('--update-timestamp', action='store_true', help='Update timestamp in database')
+parser.add_argument('--save-diagram', metavar=('bucket', 'path'), nargs=2, help='Save diagram PNG in a S3 bucket and path')
+parser.add_argument('--crash-id', metavar='int', type=int, nargs=1, default=[0], help='Specific crash ID to operate on')
+args = parser.parse_args()
```
sergiogcx, line 24: I dig all these arg options! ✅

dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
```diff
@@ -0,0 +1,247 @@
...
+
+s3 = boto3.client('s3')
+
+def update_crash_processed_date(crash_id: int) -> bool:
+    if (args.v):
```
sergiogcx, line 29: Thank you for providing type annotations in all your function implementations! 📜 

I'd also appreciate if you could add docstrings to very briefly describe what each function does using this pattern:

```
:param param_type param_name: parameter description (type description is on the same line as the parameter description).
```



Taken from: https://www.jetbrains.com/help/pycharm/using-docstrings-to-specify-types.html#example

dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
```diff
@@ -0,0 +1,247 @@
...
+        if (args.v):
+            print('Saving PNG of diagram to S3')
+        try:
+            # never touch the disk; store the image data in a few steps to get to a variable of binary data
+            buffer = io.BytesIO()
+            diagram_image.save(buffer, format='PNG')
+            output_diagram = s3.put_object(Body=buffer.getvalue(), Bucket=args.save_diagram[0], Key=args.save_diagram[1] + '/' + str(crash['crash_id']) + '.png')
+        except:
+            sys.stderr.write("Error: Faild setting s3 object containing the diagram PNG file\n")
+            continue
```
sergiogcx, line 240: Brilliant that it never saves to local disk first! This is gonna help Airflow a lot with space availability. 💾 👍

sergiogcx (approved):
Thank you @frankhereford I enjoy learning from you!

frankhereford:
@sergiogcx - thank you for taking the time to chat with me about this; I appreciate your time! 

To memorialize our thoughts:

- The program is inherently single threaded and should consume a fixed amount of RAM limited by the resolution of the raster image created by rendering the PDF page.
- The program is intended to operate on a single crash via the --crash-id argument, or to pull a batch of size _n_ based on the --batch-size argument.
- I will add the docstrings describing the arguments of the functions used in the program in the style you provided.
- Pending approval from Jace & Tilly, I will add a feature to the program where it stores the name (or not, should it not exist) or the crash diagram PNG file on S3. The intent is to provide a way for the VZE front-end to check to see if that file is available on S3 without having to actually use S3 to see if that object exists. This data will be stored in a new key => value pair in the PDF metadata field.
- I will offer diff to the Dockerfile to include the new dependencies on the python install needed by this program. 
- I will also provide a list of any external libraries which may be needed by the python modules which are acting as bindings.

mateoclarke:
Looks great Frank, thanks for contributing on this! Can't wait to see the end result in the VZE frontend

PR #33, opened 2020-12-28
Alter GraphQL query for batch processing of CR3s to newest first

frankhereford:
Based on a conversation in the dev-sync meeting and a suggestion from
@sergiogcx, the query which pulls a batch of crash records for OCR /
diagram extraction is modified to request more recent crashes first.
The intent is to be processing crashes most likely to be viewed by
VZE viewers.


sergiogcx (approved):
It works perfectly! Thanks 🙏 🥇 

🚢  Merge any time!

PR #41, opened 2021-01-22
Removes example dags

sergiogcx:
I don't know why they are being loaded (even with the .airflowignore rule), but we don't need them anymore.


johnclary:
thanks @sergiogcx! Nice to have the Airflow UI less cluttered.

PR #47, opened 2021-03-30
4962 - add knack digital messaging sign dag

chiaberry:
Issue: https://github.com/cityofaustin/atd-data-tech/issues/4962

Adds the dag for the scripts in knack-services
Also updates styling and typo in docs. 

Corresponding knack-services pr [https://github.com/cityofaustin/atd-knack-services/pull/66](https://github.com/cityofaustin/atd-knack-services/pull/66)
Corresponding atd-data-deploy PR [https://github.com/cityofaustin/atd-data-deploy/pull/82](https://github.com/cityofaustin/atd-data-deploy/pull/82)

### DIFF
--- a/dags/atd_knack_dms.py
+++ b/dags/atd_knack_dms.py
@@ -1,14 +1,14 @@
 from datetime import datetime, timedelta
 from airflow.models import DAG
 from airflow.models import Variable
-from airflow.operators.docker_operator import DockerOperator
+from airflow.operators.docker_operator import DockerOperator # deprecated module?
 from _slack_operators import task_fail_slack_alert
 
 default_args = {
     "owner": "airflow",
     "description": "Load dms (view_1564) records from Knack to Postgrest to AGOL and Socata",  # noqa:E501
     "depend_on_past": False,
-    "start_date": datetime(2020, 9, 1), # TODO: check
+    "start_date": datetime(2021, 3, 31), # TODO: check
     "email_on_failure": False,
     "email_on_retry": False,
     "retries": 0,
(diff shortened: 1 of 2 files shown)


dags/atd_knack_dms.py
```diff
@@ -0,0 +1,92 @@
...
+atd_knack_auth = Variable.get("atd_knack_auth", deserialize_json=True)
+env_vars["KNACK_APP_ID"] = atd_knack_auth[app_name][env]["app_id"]
+env_vars["KNACK_API_KEY"] = atd_knack_auth[app_name][env]["api_key"]
+env_vars["SOCRATA_API_KEY_ID"] = Variable.get("atd_service_bot_socrata_api_key_id")
+env_vars["SOCRATA_API_KEY_SECRET"] = Variable.get(
+    "atd_service_bot_socrata_api_key_secret"
+)
+env_vars["SOCRATA_APP_TOKEN"] = Variable.get("atd_service_bot_socrata_app_token")
+env_vars["AGOL_USERNAME"] = Variable.get("agol_username")
+env_vars["AGOL_PASSWORD"] = Variable.get("agol_password")
```
sergiogcx, line 40: This is fine, but for future reference every time run `Variable.get` we are actually making a SQL request. You can create a single variable entry as a JSON document and deserialize it, then pass it straight to the operator as environment variables:

https://github.com/cityofaustin/atd-airflow/blob/3d2bf9ee49c37c31bb4e0260622606b659da7f87/dags/atd_visionzero_cr3_production.py#L41

That reduces the amount of code being used, and the number of times we use `Variable.get`

sergiogcx (approved):
Thank you for working on this Chia!  🚢

dags/atd_knack_dms.py
sergiogcx: To mix two dictionaries you can always build-and-unpack, I would try something like this:
```
environment={
    **Variable.get("your_knack_creds", deserialize_json=True),
    **Variable.get("your_agol_creds", deserialize_json=True),
...
}
```

chiaberry: oh this is a great tip, I'm going to read more about this and update my dag, thank you!

PR #59, opened 2021-07-23
Add scoobi to the daily data recovery process

sergiogcx:
Scoobi was missing from the daily data recovery proces, just added it so we can run daily backfills.

### DIFF
--- a/dags/atd_mds_process_unfinished_production.py
+++ b/dags/atd_mds_process_unfinished_production.py
@@ -153,10 +153,10 @@ with DAG(
 
     # lyft >> \
     # jump >> \
+    # revel >> \
     lime >> \
     bird >> \
     wheels >> \
     spin >> \
     ojo >> \
-    revel >> \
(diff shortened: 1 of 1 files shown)


dags/atd_mds_process_unfinished_production.py
```diff
@@ -140,11 +140,23 @@
...
+
     # lyft >> \
     # jump >> \
     lime >> \
     bird >> \
     wheels >> \
     spin >> \
     ojo >> \
-    revel
+    revel >> \
```
mateoclarke, line 161: I don't think we want to add Revel back. I thought they were offline. Other than that all good.

mateoclarke (approved):

dags/atd_mds_process_unfinished_production.py
sergiogcx: Ah, good call. Let me correct and merge.

PR #60, opened 2021-07-27
6676 - add dags for Signs and Markings

chiaberry:
Corresponding knack services PR https://github.com/cityofaustin/atd-knack-services/pull/81

`atd_knack_work_order_markings`, `atd_knack_work_orders_markings_jobs` and `atd_knack_work_orders_signs` also have an arcgis process (`atd_knack_markings_work_orders` `atd_knack_markings_jobs` and `atd_knack_signs_work_orders.py`). 

The updates to socrata need to happen at 10a and 2p daily, so I created different dags to handle them. The other option I considered was adding a task to the existing dags and having them run all tasks twice a day, but chose to not run the arcgis updates unnecessarily. 

Time logs ( `view_3516`) is not currently being updated anywhere, this is its first dag.


johnclary (changes requested):
@chiaberry if you're using the same views as the existing signs/markings dags, you can modify those dags to add the socrata publishing task. That way you're not re-loading the same data from knack to postgrest. Time logs will need to be a new DAG.

R.e. schedule—you can tweak the schedule of those existing DAGs if not frequent if enough (too lazy to check rn). They should already be staggered. It's important that the Knack > Postgrest tasks do not all fire at the same time, because it's a big load on the Knack DB, particularly on the 15th of the month. 

I'll be offline for a while so feel free to merge when ready.

PR #62, opened 2021-08-08
Reference correct signs/markings time log view

johnclary:
I forgot this DAG was live when I modified the source view in atd-knack-services [#82](https://github.com/cityofaustin/atd-knack-services/pull/82)


johnclary:
@chiaberry @mateoclarke @sergiogcx FYI merging

PR #63, opened 2021-08-12
Deploy Signs and Markings Reimbursements DAGs

johnclary:
Towards fixing [#3114](https://github.com/cityofaustin/atd-data-tech/issues/3114).

- New DAG for reimbursement data
- Updates time logs DAG to source from separate views for signs vs markings

### DIFF
--- a/dags/atd_knack_signs_markings_reimbursements.py
+++ b/dags/atd_knack_signs_markings_reimbursements.py
@@ -26,7 +26,7 @@ env = "prod"
 container_signs = "view_3526"
 container_markings = "view_3527"
 
-# use pooling to throttel DB load
+# use pooling to throttle DB load
 POOL_KNACK = "knack_signs_markings"
 POOL_POSTGREST = "atd_knack_postgrest_pool"
 
--- a/dags/atd_knack_signs_markings_reimbursements.py
+++ b/dags/atd_knack_signs_markings_reimbursements.py
@@ -1,0 +1,111 @@
+from datetime import datetime, timedelta
+from airflow.models import DAG
+from airflow.models import Variable
+from airflow.operators.docker_operator import DockerOperator
+from _slack_operators import *
+
+default_args = {
+    "owner": "airflow",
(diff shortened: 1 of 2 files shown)


johnclary:
@sergiogcx @mateoclarke @chiaberry if you can please give this a quick review. I never get this right on the first try so may need to patch after merging 🤞

dags/atd_knack_signs_markings_reimbursements.py
```diff
@@ -0,0 +1,111 @@
...
+    dagrun_timeout=timedelta(minutes=60),
+    tags=["production", "knack"],
+    catchup=False,
+) as dag:
+    t1 = DockerOperator(
+        task_id="atd_knack_signs_reimbursements_to_postgrest",
+        image=docker_image,
+        api_version="auto",
+        auto_remove=True,
+        command=f"./atd-knack-services/services/{script_task_1}.py -a {app_name} -c {container_signs}",  # noqa:E501
```
sergiogcx, line 61: This is a clever way to re-use the DAG with a variable command 👍 ⚡

dags/atd_knack_signs_markings_reimbursements.py
```diff
@@ -0,0 +1,111 @@
...
+
+# command args
+script_task_1 = "records_to_postgrest"
+script_task_2 = "records_to_socrata"
+app_name = "signs-markings"
+env = "prod"
+container_signs = "view_3526"
+container_markings = "view_3527"
+
+# use pooling to throttel DB load
```
mateoclarke, line 29: throttle*

sergiogcx (approved):
Looks solid, I haven't tested syntax or anything but makes sense overall.

dags/atd_knack_signs_markings_reimbursements.py
johnclary: thx. fixed

johnclary:
that was fast. thank you both! here we gooooo

PR #67, opened 2021-09-13
Add dag for atd-knack-banner employee updates

chiaberry:
Knack banner PR: https://github.com/cityofaustin/atd-knack-banner/pull/1

This dag cannot be started until atd-knack-banner docker image is in atddocker hub.

I've added the new environment variables to the airflow variable list.

### DIFF
--- a/dags/atd_knack_banner.py
+++ b/dags/atd_knack_banner.py
@@ -40,7 +40,7 @@ env_vars["KNACK_API_KEY"] = atd_knack_auth[app_name][env]["api_key"]
 with DAG(
     dag_id="atd_knack_banner",
     default_args=default_args, #check this
-    schedule_interval="0 5 * * 0", #todo: update, weekly on sunday? when
+    schedule_interval="0 5 * * 0", # is a weekly on sunday update enough?
     dagrun_timeout=timedelta(minutes=60),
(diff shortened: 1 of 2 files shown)


dags/atd_knack_banner.py
```diff
@@ -0,0 +1,60 @@
...
+env_vars["KNACK_API_KEY"] = atd_knack_auth[app_name][env]["api_key"]
+env_vars["SHAREDDRIVE_USERNAME"] = atd_shared_drive["SHAREDDRIVE_USERNAME"]
+env_vars["SHAREDDRIVE_PASSWORD"] = atd_shared_drive["SHAREDDRIVE_PASSWORD"]
+env_vars["SHAREDDRIVE_FILEPATH"] = atd_shared_drive["SHAREDDRIVE_FILEPATH"]
+
+
+with DAG(
+    dag_id="atd_knack_banner",
+    default_args=default_args,
+    schedule_interval="0 5 * * 0", # is a weekly on sunday update enough?
```
johnclary, line 40: hmm—i think we should just do it daily. i'm not quite sure when the banner endpoint actually gets refreshed, so this would help us make sure we're catching updates early

johnclary (approved):
hooray!

PR #74, opened 2021-11-12
change knack banner run time

chiaberry:
The dag was running smoothly at 7am everyday, but since the time change, it now runs at 6am. And for some reason does not like running at 6am. 

Changing the time to 6:45am local time.


johnclary (approved):
🚢

PR #82, opened 2021-12-29
Add task to fetch payment records

johnclary:


Charlie-Henry:
Looks good. I don't think we'll lose any transactions through the cracks this way.

Charlie-Henry (approved):

PR #85, opened 2022-02-06
Move daily inventory snapshot to a separate DAG and restore previous inventory DAG

johnclary:
This PR reverts #84. It restores the original [`atd_knack_inventory_items.py`](https://github.com/cityofaustin/atd-airflow/blob/master/dags/atd_knack_inventory_items.py) DAG and moves the snapshot task to a separate DAG. It is a second attempt to complete https://github.com/cityofaustin/atd-data-tech/issues/7581.

The problem is that `atd_knack_inventory_items.py` is scheduled to run multiple times per day (it keeps inventory items in sync between the Finance & Purchasing System and the Data Tracker). The new daily inventory snapshot task should only run once a day, because it must only append one new set of inventory records **per day** to the [open dataset](https://data.austintexas.gov/Transportation-and-Mobility/Arterial-Management-Materials-Warehouse-Inventory/hcaw-evi2).

### DIFF
--- a/dags/atd_knack_inventory_daily_status.py
+++ b/dags/atd_knack_inventory_daily_status.py
@@ -1,0 +1,79 @@
+from copy import deepcopy
+from datetime import datetime, timedelta
+from airflow.models import DAG
+from airflow.models import Variable
+from airflow.operators.docker_operator import DockerOperator
+from _slack_operators import task_fail_slack_alert
+
+default_args = {
+    "owner": "airflow",
+    "description": "Append inventory item counts in the Data Tracker to Socrata open dataset",
+    "depend_on_past": False,
+    "start_date": datetime(2020, 9, 1),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=5),
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+docker_image = "atddocker/atd-knack-services:production"
+app_name = "data-tracker"
+container = "view_2863"
+env = "prod"
+
+# assemble env vars
+env_vars = Variable.get("atd_knack_services_postgrest", deserialize_json=True)
+atd_knack_auth = Variable.get("atd_knack_auth", deserialize_json=True)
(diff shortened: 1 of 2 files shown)


dags/atd_knack_inventory_daily_status.py
```diff
@@ -0,0 +1,79 @@
...
+    Push Data Tracker inventory items to PostgREST. We always push all records (by setting
+    data to 1970-01-01) because it is required to create a snapshot of the
+    inventory quantities on-hand.
+    """
+    t1 = DockerOperator(
+        task_id="atd_knack_data_tracker_inventory_items_to_postgrest",
+        image=docker_image,
+        api_version="auto",
+        auto_remove=True,
+        command=f'./atd-knack-services/services/records_to_postgrest.py -a {app_name} -c {container} -d "1970-01-01"',  # noqa
```
frankhereford, line 53: TIL #noqa as a IDE hint for python - thanks!

dags/atd_knack_inventory_daily_status.py
```diff
@@ -0,0 +1,79 @@
...
+env_vars["SOCRATA_API_KEY_SECRET"] = Variable.get(
+    "atd_service_bot_socrata_api_key_secret"
+)
+env_vars["KNACK_APP_ID"] = atd_knack_auth[app_name][env]["app_id"]
+env_vars["KNACK_API_KEY"] = atd_knack_auth[app_name][env]["api_key"]
+
+with DAG(
+    dag_id="atd_knack_inventory_items_nightly_snapshot",
+    default_args=default_args,
+    schedule_interval="13 4 * * *",  # at 4:13am UTC
```
frankhereford, line 38: Thanks for denoting that cron timing is specified in UTC in airflow. I have had to go and check to make sure it's not local time more than once!

frankhereford (approved):
Looks great - thank you @johnclary! 🚢 🚢

dags/atd_knack_inventory_daily_status.py
johnclary: indeed—it's much better practice to specify the specific lint rule e.g. `# noqa: E501`

johnclary:
thanks y'all!

PR #87, opened 2022-03-18
Add boto3 to airflow image

frankhereford:
## Overview
This is a minor change that came out of building the airflow image locally to investigate the OCR mechanism used on VZ CR3 files. Included are the following changes:

* Add boto3 to the airflow image. I imagine it was added by one of us who built and published the image to dockerhub, but the change to the dockerfile didn't make it back into this repo.
* Two corrections for spelling errors in comments
* The addition of an environment file name that is more descriptive to where it points and is not hidden as a system file. This gives us a place to keep environment variable files in our local checkout without fear of checking them in one day by mistake.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,6 +119,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+production_env
 
 # Spyder project settings
 .spyderproject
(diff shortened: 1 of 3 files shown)


mateoclarke (approved):
👍

chiaberry (approved):

PR #93, opened 2022-12-28
Update OCR script in anticipation of new CR3 format

chiaberry:
Related issue: 
https://github.com/cityofaustin/atd-data-tech/issues/9704

-- edited 4/4

Now that we have a copy of actual new cr3 forms, I reworked the script to check if the pixels are black on the newer cr3, as the pdfs are slightly different. If the pdf doesnt match up with the new pixels, then it goes to check in the old pixels. 

I tested this by downloading a new cr3 and checking the pixels and cropped image locally in a python console. I am not sure what the best way is to test the whole script, besides letting it run in production.

### DIFF
--- a/dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
+++ b/dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
@@ -5,27 +5,60 @@ import os
 import re
 import sys
 import json
-from   uuid import uuid4
+from uuid import uuid4
 import boto3
 import argparse
 import requests
-from   pdf2image import convert_from_path, convert_from_bytes
+from pdf2image import convert_from_path, convert_from_bytes
 import pytesseract
 
 # Configure the available arguments to the program
-parser = argparse.ArgumentParser(description='Extract CR3 diagrams and narratives')
-parser.add_argument('-v', action='store_true', help='Be verbose')
-parser.add_argument('-d', action='store_true', help='Check for digitally created PDFs')
-parser.add_argument('--cr3-source', metavar=('bucket', 'path'), nargs=2, required=True, help='Where can we hope to find CR3 files on S3?')
-parser.add_argument('--batch-size', metavar='int', default=1, help='How many cr3s to attempt to process?')
-parser.add_argument('--update-narrative', action='store_true', help='Update narrative in database')
-parser.add_argument('--update-timestamp', action='store_true', help='Update timestamp in database')
-parser.add_argument('--save-diagram-s3', metavar=('bucket', 'path'), nargs=2, help='Save diagram PNG in a S3 bucket and path')
-parser.add_argument('--save-diagram-disk', metavar=('path'), nargs=1, help='Save diagram PNG to disk in a certain directory')
-parser.add_argument('--crash-id', metavar='int', type=int, nargs=1, default=[0], help='Specific crash ID to operate on')
+parser = argparse.ArgumentParser(description="Extract CR3 diagrams and narratives")
+parser.add_argument("-v", action="store_true", help="Be verbose")
+parser.add_argument("-d", action="store_true", help="Check for digitally created PDFs")
+parser.add_argument(
+    "--cr3-source",
+    metavar=("bucket", "path"),
+    nargs=2,
+    required=True,
+    help="Where can we hope to find CR3 files on S3?",
+)
+parser.add_argument(
+    "--batch-size",
+    metavar="int",
+    default=1,
+    help="How many cr3s to attempt to process?",
+)
+parser.add_argument(
+    "--update-narrative", action="store_true", help="Update narrative in database"
+)
+parser.add_argument(
+    "--update-timestamp", action="store_true", help="Update timestamp in database"
+)
+parser.add_argument(
+    "--save-diagram-s3",
+    metavar=("bucket", "path"),
+    nargs=2,
+    help="Save diagram PNG in a S3 bucket and path",
+)
+parser.add_argument(
+    "--save-diagram-disk",
+    metavar=("path"),
+    nargs=1,
+    help="Save diagram PNG to disk in a certain directory",
+)
+parser.add_argument(
+    "--crash-id",
+    metavar="int",
+    type=int,
+    nargs=1,
+    default=[0],
+    help="Specific crash ID to operate on",
+)
 args = parser.parse_args()
 
-s3 = boto3.client('s3')
+s3 = boto3.client("s3")
+
 
 def update_crash_processed_date(crash_id: int) -> bool:
     """
@@ -34,7 +67,7 @@ def update_crash_processed_date(crash_id: int) -> bool:
     :param int: crash_id
     """
 
-    if (args.v):
+    if args.v:
         print("update_crash_processed_date(" + str(crash_id) + ")")
     query = """
     mutation update_crash_processsed_date($crash_id: Int) {
@@ -48,15 +81,10 @@ def update_crash_processed_date(crash_id: int) -> bool:
         headers={
             "Accept": "*/*",
             "content-type": "application/json",
-            "x-hasura-admin-secret": os.getenv("HASURA_ADMIN_KEY")
-            },
-        json={
-            "query": query,
-            "variables": {
-                "crash_id": crash_id
-                }
-            }
-        )
+            "x-hasura-admin-secret": os.getenv("HASURA_ADMIN_KEY"),
+        },
+        json={"query": query, "variables": {"crash_id": crash_id}},
+    )
     try:
         return response.json()["data"]["update_atd_txdot_crashes"]["affected_rows"] > 0
     except (KeyError, TypeError):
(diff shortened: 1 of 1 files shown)


dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
```diff
@@ -2,39 +2,70 @@
...
-parser.add_argument('-v', action='store_true', help='Be verbose')
-parser.add_argument('-d', action='store_true', help='Check for digitally created PDFs')
-parser.add_argument('--cr3-source', metavar=('bucket', 'path'), nargs=2, required=True, help='Where can we hope to find CR3 files on S3?')
-parser.add_argument('--batch-size', metavar='int', default=1, help='How many cr3s to attempt to process?')
-parser.add_argument('--update-narrative', action='store_true', help='Update narrative in database')
-parser.add_argument('--update-timestamp', action='store_true', help='Update timestamp in database')
-parser.add_argument('--save-diagram-s3', metavar=('bucket', 'path'), nargs=2, help='Save diagram PNG in a S3 bucket and path')
-parser.add_argument('--save-diagram-disk', metavar=('path'), nargs=1, help='Save diagram PNG to disk in a certain directory')
-parser.add_argument('--crash-id', metavar='int', type=int, nargs=1, default=[0], help='Specific crash ID to operate on')
+parser = argparse.ArgumentParser(description="Extract CR3 diagrams and narratives")
```
chiaberry, line 14: This looks like a lot of changes, but its mostly linting with Black

dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
```diff
@@ -153,8 +175,9 @@ def update_crash_metadata(crash_id: int, metadata: dict) -> bool:
   }
 }
 """
-batch_variables = { "limit": int(args.batch_size) }
+batch_variables = {"limit": int(args.batch_size)}
 
+# query for specific cr3, selected by crash_id
 get_single_cr3 = """
 query find_cr3s($crash_id: Int, $limit: Int) {
```
chiaberry, line 182: I was considering editing this query to remove the $limit var, since this will only ever return 1 record, but I decided against refactoring code that wasn't related to the OCR.

dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
```diff
@@ -167,151 +190,183 @@ def update_crash_metadata(crash_id: int, metadata: dict) -> bool:
...
+        )
         continue
 
-
-    if (args.d):
-        if (args.v):
-            print('Excuting a check for a digitally created PDF');
+    if args.d:
+        if args.v:
+            print("Executing a check for a digitally created PDF")
```
chiaberry, line 244: After this line there are actual code changes

dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
```diff
@@ -167,151 +190,183 @@ def update_crash_metadata(crash_id: int, metadata: dict) -> bool:
...
+            (288, 1364),
+            (1088, 1566),
+            (79, 1361),
+            (1216, 105),
+        ]
+        # new cr3 form size is 1275x1650
+        if pages[1].size[1] < 2000:
+            new_cr3_form = True
+            if args.v:
+                print("CR3 created with new 2023 version format")
```
chiaberry, line 269: When I ran `pages = convert_from_bytes(pdf["Body"].read(), 150)` on the new form that was linked in the issue, I got a smaller sized image. However, if the new forms are not actually going to be a smaller resolution, I will have to revisit this and find a different way to check if the form is new or not. 

If the files do end up being the same resolution, my backup idea was to check if the pixels are black and then if they arent, check if the new_pixels are black, and if they are then its a new CR3

dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
```diff
@@ -84,28 +110,26 @@ def update_crash_narrative(crash_id: int, narrative: str) -> bool:
...
         sys.stderr.write(response.json())
 
+
 def update_crash_metadata(crash_id: int, metadata: dict) -> bool:
     """
-    Store the extracted crash diagram in the metadata field, returning a boolean indicating success.
-
-    :param int: Crash ID
-    :param dict: Metadata
+    Store the extracted crash diagram in the metadata field
```
chiaberry, line 129: still linting, keep scrolling

dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
frankhereford: I wish I had read this comment more closely when I asked that very uninformed question yesterday about the pixel constellations. This is exact right way to handle this!

# 🐰 🕳️ 👇

I decided to see if I could dig more into the dimensions intrinsic to these two PDFs to see if they align or can inform more of your findings. I had to look it up to remind myself: 1 point is 1/72nd of an inch, because you know, reasons.

### The 2023 blank form linked in the related issue
image

### "Filled-in" CR3 PDF from the past
image

The PDFs from the past have a wacky-huge, non-standard poster-ish size to them.

## Prediction 🔮

This is just a guess, but I think the process that CRIS is using to burn data into a blank CR3 form, which happens on-demand, when a CR3 is downloaded -- I think that process is taking a vector PDF of nominal dimensions, filling in the "value" attributes in the defined fields, and then "burning" that PDF out by rasterizing it, but then re-wrapping that raster image into a PDF. The result is a jpg-like image in a PDF container -- which is larger in size and immutable -- no way to edit the XML to change a value, because now there isn't a XML DOM-like tree in the PDF, it's become a very tiny XML wrapper around a TIFF essentially. That immutability is a good thing for these "official" documents.

Anyway- all that said, I think the nice, tidy, well-formed PDF from which you derived the pixel coordinates (or its descendant) will be used as input into their rasterize-and-serve PDF generator, and come the 1st, we'll be getting large PDFs again in the same dimensions as the old, poster sized ones.

frankhereford (approved):
Wow, I wish I knew about `black` when I worked on this back then!! Thank you for cleaning up my mess! ✨🍰✨

This looks great. I put a comment on your note about the PDF sizes, but it's just my little conspiracy theory about TxDOT -- can't wait to see how this plays out on the first. 

Re testing: without actual input, I agree, we can't really test anything, so I think this is more than ready to be waiting in the wings for 2023. 🎆

🚢 🚀✨

dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
chiaberry: I agree with your prediction!

frankhereford: @chiaberry Do you think we should de-schedule the Prefect flow that does the rasterizing prior to the 1st (like maybe today at EOB) so we don't have any clean-up while we test, refine and deploy your new work after the change over?

mddilley (approved):
Very cool to see these constellations that I keep hearing about. 🌌 I enjoyed the discussion about the size of the completed CR3s that the script will actually be handling vs. the blank pdf. I agree with y'all that we'll see when the real deal comes in the new year. Thanks for the investigative work! 🕵️

roseeichelmann (approved):
This looks good to me! I'm interested to see how it will work come 2023 and if we will end up having to go the pixel route for checking if its the new form or if the page size thing will work. Nice investigating 🔬

patrickm02L (approved):
Can't wait to see this in action! 🤘

dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
```diff
@@ -167,151 +190,185 @@ def update_crash_metadata(crash_id: int, metadata: dict) -> bool:
...
-            if not(rgb_pixel[0] == 0 and rgb_pixel[1] == 0 and rgb_pixel[2] == 0):
-                digital_end_to_end = False
-            if (args.v):
-                print('Pixel' + "(%04d,%04d)" % pixel + ': ' + str(rgb_pixel))
-        if (args.v):
-            print('PDF Digital End to End?: ' + str(digital_end_to_end));
-        if not(digital_end_to_end):
-            if (args.v):
+            if not (rgb_pixel[0] == 0 and rgb_pixel[1] == 0 and rgb_pixel[2] == 0):
+                new_cr3_form = False
```
chiaberry, line 268: if these aren't black then we know its not the new cr3 form and we should check to see if its the old one

dags/python_scripts/cr3_extract_diagram_ocr_narrative.py
```diff
@@ -167,151 +190,185 @@ def update_crash_metadata(crash_id: int, metadata: dict) -> bool:
...
-        diagram_image = pages[1].crop((2589,3531,5001,6048))
+        if new_cr3_form:
+            narrative_image = pages[1].crop((90, 3026, 2496, 5466))
+            diagram_image = pages[1].crop((2496, 3036, 4836, 5464))
+        else:
+            narrative_image = pages[1].crop((96, 3683, 2580, 6049))
+            diagram_image = pages[1].crop((2589, 3531, 5001, 6048))
     except:
-        sys.stderr.write("Error: Failed to extract the image of the narative and diagram from image in memory\n")
+        sys.stderr.write(
```
chiaberry, line 298: everything below this is linting again

frankhereford (approved):
I've reviewed the code, and it looks right on to me. I have not attempted to run the program, but I feel that's OK since we can rerun with modifications if they are needed, but I can't imagine that will be the case.

Thank you so much for making these changes, especially on such short notice!! 🚢 🚢 🚀

mddilley (approved):
I read through the last two commits, and those changes look good to me! Thanks for getting to the bottom of this. I trust your testing on the new format, and I agree with Frank that we can adjust if we need to. 🚀🙏

PR #95, opened 2023-05-21
2️⃣ Changes to make primary airflow2 branch "local first"

frankhereford:
## Intent
The intent is that a user can pull the primary branch and have it out-of-the-box ready to run locally. 

## Considerations & `production` branch
This leads to the question: What about the changes that need to be made to make it run in production, namely SSL & listening port configurations?

A solution is to have a `production` branch, from which `main` will be merged into as part of the release cycle. Devs write features, PR them into `main` and when one or more features have accumulated in `main` and should be released, then a dev will PR `main` into `production`. This is very similar to the development cycle of our main web applications, however, in practice, a dev will frequently do the release from `main` into `production` autonomously after merging into `main`, piggy-packing the authority they received via reviews on their PR to take the feature all the way to production. This piggy-backing of authority is in lieu of having regular release parties, which don't fit our needs as well in this situation.

And finally, to answer the initial question: The `production` branch will have a few commits that it has which keep it ahead of main. These commits contain the changes to `main` which configure the application to run in production. These commits will be concerned with the ports the application listens on & the SSL related directives for HAProxy. 

### Important Note
The handful of commits which describe the changes to `main` required for production are not intended to ever make it back into `main`. They live in `production` only, as they are only germane to the production deployment. Letting them back into `main` will make the `main` branch no longer ready to run locally, out-of-the-box.

## Next Steps
When this has made it into `airflow-v2`, `production` should be created off of `airflow-v2` at that time and have the ports and SSL configuration adjusted for production. These changes are committed into `production`, pushed to GitHub, and then the production deployment should have its upstream source set to that production branch and pulled to sync to it.


.github/sanity-check.yml
frankhereford, line 1: This was a dup of `.github/actions/sanity-check.yml` and was removed.

README.md
```diff
@@ -56,6 +56,7 @@ _AIRFLOW_WWW_USER_PASSWORD=<pick your initial admin pw here>
 AIRFLOW_PROJ_DIR=<absolute path of airflow checkout>
 OP_API_TOKEN=<Get from 1Password here: 'name TBD'>
 OP_CONNECT=<URL of the 1Password Connect install>
+OP_VAULT_ID=<OP Vault ID>
```
frankhereford, line 59: Record the new requirement of the vault ID, such that it becomes available for us as an environment variable.

mddilley (approved):
Awesome - I agree with this plan to keep production configuration in its own branch and letting devs do solo release to production so the CI can handle updates on atd-data03. Like you mentioned earlier, the isolation of the configuration from DAG changes will make that task really straightforward and avoid any danger territory. 

Thanks for pushing the code back towards local dev so we will have a solid `main` branch to PR against. 🙏

frankhereford:
This is going into the main v2 branch, bringing with it the integration with smee.io as a webhook forwarding agent.

PR #96, opened 2023-05-21
1️⃣ Integrates Smee.io as a web hook relay and reworks system for tighter 1PW integration

frankhereford:
## Intent

* 🏴‍☠️🪝Integrate Smee.io as a zero-configuration web hook relay service
* Rework web hook for tighter 1PW integration and include additional safe guards
  * Only will operate when the environment is set to production (not locally)
  * Get web hook relay secrets out of 1PW
  * Remove concept of storing enabled keys in checkout, use 1PW instead
  * Use headers instead of JSON body to convey access token


webhook-relay/Dockerfile
frankhereford, line 1: Here's a good example of taking a debian based docker package and installing the `op` command, which is the 1PW CLI. Peek in the `entry_point.sh` to see how it's used to access secrets.

webhook-relay/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+https://downloads.1password.com/linux/debian/$(dpkg --print-architecture) \
+stable main" | \
+tee /etc/apt/sources.list.d/1password.list
+RUN mkdir -p /etc/debsig/policies/AC2D62742012EA22/
+RUN curl -sS https://downloads.1password.com/linux/debian/debsig/1password.pol | \
+tee /etc/debsig/policies/AC2D62742012EA22/1password.pol
+RUN mkdir -p /usr/share/debsig/keyrings/AC2D62742012EA22
+RUN curl -sS https://downloads.1password.com/linux/keys/1password.asc | \
+gpg --dearmor --output /usr/share/debsig/keyrings/AC2D62742012EA22/debsig.gpg
+RUN apt update && apt install 1password-cli
```
frankhereford, line 27: Here's an example of installing the `op` command, so you can use it (see `entry_point.sh`) to get secrets in a shell context.

webhook-relay/entry_point.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+WEBHOOK_PROXY_URL=$(op read op://$OP_VAULT_ID/Airflow\ Webhook/smee/address)
```
frankhereford, line 3: 1PW secret ➜ environment variable

webhook/webhook.py
```diff
@@ -1,23 +1,27 @@
...
+import shutil
 from onepasswordconnectsdk.client import Client, new_client
 import onepasswordconnectsdk
 
 ONEPASSWORD_CONNECT_TOKEN = os.getenv("OP_API_TOKEN")
 ONEPASSWORD_CONNECT_HOST = os.getenv("OP_CONNECT")
-VAULT_ID = "quvhrzaatbj2wotsjrumx3f62a"
-SECRET_NAME = "airflow.fyi webhook key"
+VAULT_ID = os.getenv("OP_VAULT_ID")
+OPERATING_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
```
frankhereford, line 12: While we have "leaked" our vault ID, it's not something we need to be concerned about. It's secured, and there isn't any changing it. ✅

webhook/webhook.py
frankhereford, line 1: I reworked this pretty heavily. It's much easier to read [here](https://github.com/cityofaustin/atd-airflow/blob/webhook-relay/webhook/webhook.py).

README.md
```diff
@@ -66,17 +66,6 @@ OP_VAULT_ID=<OP Vault ID>
...
-* GitHub key pair
-  * Public key installed on GitHub
-  * Private key in `private_key_for_github` at the top of the repo
-    * Starts and ends with 
-      * `-----BEGIN OPENSSH PRIVATE KEY-----`
-      * `-----END OPENSSH PRIVATE KEY-----` or similar
-  * We could change the checkout on the production machine to a `HTTPS` instead of `ssh` based git origin
-    * This change would eliminate the need for this file / key
-* `.env` file
-  * See above
```
frankhereford, line 79: The key used for deploying Airflow is now properly configured as a deploy key and is stored in 1PW. No longer need this messy business of keeping it in an ignored file.

.github/workflows/production_deployment.yml
frankhereford, line 1: This file is heavily reworked. It's easier to read [here](https://github.com/cityofaustin/atd-airflow/blob/webhook-relay/.github/workflows/production_deployment.yml).

frankhereford, line 1: This file is heavily reworked. It's easier to read [here](https://github.com/cityofaustin/atd-airflow/blob/webhook-relay/.github/workflows/production_deployment.yml).

mddilley (approved):
Thanks for demoing this earlier today. I think it is so smart to let 1PW take care of the secret handling here, and I'll be sure to follow that pattern for the renew script so we don't have any AWS secrets lying around. 🔐 This is super solid and will keep our DAGs fresh ✨

webhook-relay/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+# install 1PW cli client
+RUN curl -sS https://downloads.1password.com/linux/keys/1password.asc | \
+gpg --dearmor --output /usr/share/keyrings/1password-archive-keyring.gpg
+RUN echo "deb [arch=$(dpkg --print-architecture) \
+signed-by=/usr/share/keyrings/1password-archive-keyring.gpg] \
+https://downloads.1password.com/linux/debian/$(dpkg --print-architecture) \
+stable main" | \
+tee /etc/apt/sources.list.d/1password.list
+RUN mkdir -p /etc/debsig/policies/AC2D62742012EA22/
```
johnclary, line 21: can y'all clarify what this magic string is? `AC2D62742012EA22` ?

webhook-relay/entry_point.sh
johnclary: how is this instance authenticated with onepassword? ie how is this script able to read secrets from the one password vault?

edit: i see these env vars come in via docker compose.

webhook/webhook.py
johnclary: the packages in this app's `requirements.txt` need version pins.

webhook/webhook.py
```diff
@@ -26,24 +30,69 @@
...
+            # pull the latest changes from the git repository
+            data = subprocess.run(['git', 'pull'], env=environment, cwd=r'/opt/airflow', check=True, stdout=subprocess.PIPE).stdout
+            # print the output to the logger
+            app.logger.info(data.decode('utf-8'))
+            return 'git pull webhook received successfully'
+    # if the access token doesn't match, return an error
     else:
-        sys.stderr.write("failed to auth\n")
+        app.logger.info("failed webhook auth\n")
         return 'Failed to authenticate to webhook'
```
johnclary, line 69: just eyeballing it, but i think this returns a `200` response? you can use `abort` to end the request with an error code - [docs](https://flask.palletsprojects.com/en/1.1.x/api/#flask.abort).

johnclary: i'm also curious if this failure would make it's way back to Github so that we can see if the action failed...

webhook-relay/Dockerfile
frankhereford: Sure! Good question.

The installation of the 1Password ecosystem on a Debian machine is documented in these instructions:

https://support.1password.com/install-linux/

This particular string is the ID of `debsig-verify` policy published by 1Password. You can find more information about `debsig-verify` here: https://github.com/guillemj/debsig-verify. It's used to verify the authenticity of packages as having been published from 1Password.

webhook/webhook.py
```diff
@@ -26,24 +30,69 @@
...
-
-    if payload["data"]["webhook_key"] == SECRET_VALUES["secret_value"]:
-        sys.stderr.write("successful webhook auth\n")
-        os.chdir('/opt/airflow')
-        environment = dict(os.environ)
-        environment['GIT_SSH_COMMAND'] = 'ssh -i /opt/private_key_for_github -o IdentitiesOnly=yes -o "StrictHostKeyChecking=no"'
-        data = subprocess.run(['git', 'pull'], env=environment, cwd=r'/opt/airflow', check=True, stdout=subprocess.PIPE).stdout
-        return 'Received webhook payload'
+    # check if the the access token matches
+    if client_key == SECRET_VALUES["webhook_token"]:
```
johnclary, line 47: i think this is pretty low-risk, but the recommended way to evaluate a password is to use the `hmac` library to avoid [a timing attack](https://sqreen.github.io/DevelopersSecurityBestPractices/timing-attack/python). we came across this somewhere in the Moped API but i can't find it at the moment.

webhook/webhook.py
frankhereford: That's a great point. I don't think that it does, and that was something I ended up getting stuck on for a while when doing that initial dev. It should be able to do that too - I think this is a legitimate deficiency. I'd want to experiment some more to really know what works and what doesn't, but I'm not feeling great about this.

webhook/webhook.py
frankhereford: Huge TIL for me here. I strongly support this suggestion!

johnclary: one other thing: is there a chance the client key and secret both = `None`? i think you want to add an assertion that the `SECRET_VALUES["webhook_token"]` is not falsey.

webhook/webhook.py
```diff
@@ -26,24 +30,69 @@
...
 
 @app.route('/webhook', methods=['POST'])
 def handle_webhook():
-    payload = request.get_json()
+    # we only want to run this in production, so we check the environment variable
+    if OPERATING_ENVIRONMENT != "production":
+        app.logger.info("\nnot in production, skipping webhook\n")
+        return "Not in production, skipping webhook"
+    # get the client key from the request header
+    client_key = request.headers.get('X-Webhook-Key')
```
johnclary, line 43: related to my comment below, we should probably terminate the request right here if `client_key` is falsey

webhook/webhook.py
johnclary: a few more questions:
- do y'all think we need to stick this flask app between a WSGI server?

~in the webhook's dockerfile, i see this line~ edit: nm!

webhook/webhook.py
```diff
@@ -26,24 +30,69 @@
...
+            app.logger.info(data.decode('utf-8'))
+            return 'git pull webhook received successfully'
+    # if the access token doesn't match, return an error
     else:
-        sys.stderr.write("failed to auth\n")
+        app.logger.info("failed webhook auth\n")
         return 'Failed to authenticate to webhook'
 
+
+def write_to_file(path, content):
```
johnclary, line 72: this is minor, but it would probably be better to take the context manager approach here, so that the file gets closed automatically if something short-circuits.

```python
with os.open(path, os.O_WRONLY | os.O_CREAT, 0o600) as fd:
   # do stuff
```

johnclary: by the way, i _think_ the python file mode strings are OS independent? e.g. `os.open(path, "w+")` - [docs](https://docs.python.org/3.3/library/functions.html#open)

webhook/webhook.py
```diff
@@ -26,24 +30,69 @@
...
+class TempDir:
+    def __init__(self):
+        self.path = None
+
+    def __enter__(self):
+        self.path = tempfile.mkdtemp(dir='/tmp')
+        return self.path
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        shutil.rmtree(self.path)
```
johnclary, line 93: i think i mentioned this somewhere else, but this is cool to see in the wild! TIL for me.

PR #97, opened 2023-05-22
Have docker auto-remove containers when they finish

frankhereford:
This change is in response to a `docker system prune` that took ages earlier on data03. There is nothing inherently wrong with leaving the containers hanging around after they have quit, but this is just a little tidier.

With approval, this will merge into the main `airflow-v2` branch, which is standing as `main` until it gets itself into `main`.  I'll take that approval and PR `airflow-v2` into `production`, and then trigger the webhook via smee, so this will be a nice test of that system as well, as it stands now in production.


mddilley (approved):
Ready for smee! 🪝

PR #99, opened 2023-05-22
Update README to reflect current configuration & improve docs

frankhereford:
Just some housekeeping. Some of our data had gotten out of date. I think we might keep some of the links around in the feature list for a bit while our team adopts the new capabilities, but they are messy, and should probably go at some point.


mddilley (approved):
📚 🙏 ✨

PR #100, opened 2023-05-23
Migrate service bot flows to Airflow v2 on atd-data03

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12330

This migrates the three service bot flows to Airflow v2. I did not use the new decorators (`@dag` and `@task`) so that these could be directly compared to the Airflow v1 DAGs (linked below). I can update them if we feel strongly that all migrated DAGs should use the latest syntax.

The previous Airflow DAGs:
- https://github.com/cityofaustin/atd-airflow/blob/master/dags/atd_service_bot_production.py
- https://github.com/cityofaustin/atd-airflow/blob/master/dags/atd_service_bot_issues_to_socrata.py
- https://github.com/cityofaustin/atd-airflow/blob/master/dags/atd_service_bot_index_issues_to_dts_portal.py

## Testing
[Spin up the local stack
](https://github.com/cityofaustin/atd-airflow/blob/airflow-v2/README.md#local-setup)
**Steps to test:**
Either trust me that this works, or you can go through the following steps.

#### Intake
1. Submit a service request in the Knack app called DTS | Data and Technology Services Portal | PRODUCTION Copy which I've linked below but before that:
   - Be sure to name the request something obvious like "TEST Service Bot issue" because it will create a real issue [here](https://github.com/cityofaustin/atd-data-tech/issues)
   - Be sure to check that the url matches the one below and that you are not creating a new service request in the production app or else it will get picked up by the production bot instead of your local Airflow v2 stack (sometimes clicking links in Knack apps can redirect you to the production app without you realizing it)
   - https://atd.knack.com/dts--data-and-technology-services-portal--production-copy#new-service-request/
   - Also, be aware that this will end up triggering an email that goes to the transportation.data inbox so be sure to clear it out of that inbox or give the service desk primary a heads up.
2. In your [local Airflow dashboard](http://localhost:8080/home), switch on the DAG called **atd_service_bot_issue_intake_development**
3. You should see the DAG succeed, and your new issue will be viewable at https://github.com/cityofaustin/atd-data-tech/issues
4. Close out the issue and clean up the email in the transportation.data inbox

#### Github issues to Socrata
1. Go to the test service issues open dataset at https://datahub.austintexas.gov/Transportation-and-Mobility/TEST-ATD-Data-Tech-Services-Issues/93ru-6p6e/
2. Login with the password manager entry called **Socrata Admin | City Data Hub/EDP**
3. If you sort descending by `created_at` then you should see issues in the dataset from today (May 23rd) and that means that the script worked 🎉

#### Update index issues in DTS portal
1. Go to https://builder.knack.com/atd/dts-data-and-technology-services-portal-production-copy/records/objects/object_30 and login using either your personal login (if you have access to this app) or the Knack builder one in the password manager
2. Sort the this `project_evaluation` table descending by `issue_number` to find `12351`
3. This row was created after I briefly added the **Project Index** label to https://github.com/cityofaustin/atd-data-tech/issues/12351, updated the issue's title, and then ran this flow. The row should show the title as `🤖 This is a Airflow v2 service bot test being updated 🤖`.

## Ship list
- [x] Track down duplicates in other vaults and note new location in API vault


frankhereford:
@mddilley I have not gotten a chance to review this yet, but I am super excited to. Thanks Mike!!

We talked about this real quick on slack, but I want to open the idea up to the rest of the group. I wrote:

> what do you think about starting some sort of organizational hierarchy in the dags/ folder? I was thinking I’d put the cris_import.py script i’m working on in dags/vision-zero/cris-import/ so that it would be in that folder and i can put its libs and other supporting files in subfolders or alongside it

I think we were immediately aligned on the idea, and you mentioned that it was something we had talked about doing with prefect too at some point. We agreed that it was hard to find files and their supporting libraries under the airflow 1 system we had in place, too.

Thanks!!! Can't wait to review this!

dags/atd_service_bot_intake.py
```diff
@@ -0,0 +1,78 @@
...
+from datetime import datetime, timedelta
+
+
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+
+from onepasswordconnectsdk.client import Client, new_client
+import onepasswordconnectsdk
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
```
johnclary, line 10: i am really excited about this! i would prefer this convention:
- use the var name `ENV` instead of `ENVIRONMENT`
- use `prod`, `staging`, `dev` as env names. where `dev` always means local and `staging` usually means a remote non-prod environment.

these don't have to be hard and fast rules for usage but the lower-case shortnames are less prone to typos

dags/atd_service_bot_intake.py
```diff
@@ -0,0 +1,78 @@
...
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+}
+
+docker_image = "atddocker/atd-service-bot:production"
+
+REQUIRED_SECRETS = {
+    "KNACK_APP_ID": {
+        "opitem": "Service Bot",
```
johnclary: in this case, the service bot is the only repo that interacts with this Knack app, but i kind of want to see these secrets broken apart into separate 1pass items. one entry would be prod/dev for the DTS Portal Knack app and user login, and a separate entry would hold the github bot token. not so important in this case, but when we get to other ETLs that interact with, e.g., the Data Tracker, the API key is shared by many ETLs so we definitely don't want a 1:1 relationship between repo/ETL and secrets.

in other words, to avoid duplication and make rotating easy, we should avoid ETL-specific entries in the 1pass vault.

also, once we get this ETL deployed we should remove these keys from any other 1Pass vault that has them.

dags/atd_service_bot_intake.py
```diff
@@ -0,0 +1,78 @@
...
+client: Client = new_client(ONEPASSWORD_CONNECT_HOST, ONEPASSWORD_CONNECT_TOKEN)
+env_vars = onepasswordconnectsdk.load_dict(client, REQUIRED_SECRETS)
+
+
+with DAG(
+    dag_id=f"atd_service_bot_issue_intake_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    schedule_interval="* * * * *",
+    dagrun_timeout=timedelta(minutes=5),
+    tags=[DEPLOYMENT_ENVIRONMENT, "knack", "atd-service-bot", "github"],
```
johnclary: these tags are great. maybe we should adopt the tag convention `repo:atd-service-bot` to be even more explicit about source code?

dags/atd_service_bot_intake.py
```diff
@@ -0,0 +1,78 @@
...
+    catchup=False,
+) as dag:
+    t1 = DockerOperator(
```
johnclary, line 58: TIL there is an [arg to this operator](https://airflow.apache.org/docs/apache-airflow-providers-docker/stable/_api/airflow/providers/docker/operators/docker/index.html) called `force_pull`!

> Pull the docker image on every run. Default is False.

And it occurs to me that while the webhook will keep our DAG/airflow repo current, we have not discussed the issue of keeping Docker images current. The v1 instance uses a cron job—and I think we can fully replace that with this arg?

dags/atd_service_bot_index_issues_to_dts_portal.py
```diff
@@ -0,0 +1,70 @@
+import os
+from datetime import datetime, timedelta
+
+
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+
+from onepasswordconnectsdk.client import Client, new_client
+import onepasswordconnectsdk
```
johnclary: i see that `onepasswordconnectsdk` and `docker` are unpinned in the `requirements.txt` — should we nail those down to a major version?

dags/atd_service_bot_issues_to_socrata.py
```diff
@@ -0,0 +1,90 @@
...
+client: Client = new_client(ONEPASSWORD_CONNECT_HOST, ONEPASSWORD_CONNECT_TOKEN)
+env_vars = onepasswordconnectsdk.load_dict(client, REQUIRED_SECRETS)
+
+with DAG(
+    dag_id=f"atd_service_bot_github_to_socrata_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    schedule_interval="21 5 * * *",
+    dagrun_timeout=timedelta(minutes=60),
+    tags=[DEPLOYMENT_ENVIRONMENT, "socrata", "atd-service-bot", "github"],
+    catchup=False,
```
johnclary, line 64: 👍 this is the arg that prevents missed DAG runs from queuing up. afaik we **always** want this behavior and it's easy to forget!

dags/atd_service_bot_index_issues_to_dts_portal.py
johnclary: i also see that the [weather](https://github.com/cityofaustin/atd-airflow/blob/airflow-v2/dags/weather.py#LL3C8-L3C16) DAG depends on `pendulum`, but that's not in the requirements file, maybe because pendulum is bundled as an Airflow dependency? if that's the case, i think we want to add pendulum to our requirements file with a comment that the weather DAG depends on it. if we leave it unpinned, iirc PIP will defer to the version specified by airflow's requirements which i guess is the best option.

dags/atd_service_bot_intake.py
```diff
@@ -0,0 +1,78 @@
...
+    },
+    "KNACK_DTS_PORTAL_SERVICE_BOT_PASSWORD": {
+        "opitem": "Service Bot",
+        "opfield": "shared.dtsPortalLoginPassword",
+        "opvault": VAULT_ID,
+    },
+}
+
+client: Client = new_client(ONEPASSWORD_CONNECT_HOST, ONEPASSWORD_CONNECT_TOKEN)
+env_vars = onepasswordconnectsdk.load_dict(client, REQUIRED_SECRETS)
```
Charlie-Henry: I think you might have mentioned this at dev sync but is there a way to just pull in an entire group of secrets as a dict similar to what we have in Prefect (and I think some other airflow 1 dags?) without having to define each secret individually?

dags/atd_service_bot_intake.py
```diff
@@ -0,0 +1,78 @@
...
+        task_id="dts_sr_to_github",
+        image=docker_image,
+        api_version="auto",
+        auto_remove=True,
+        command="./atd-service-bot/intake.py",
+        environment=env_vars,
+        tty=True,
+    )
+
+    t1
```
Charlie-Henry, line 69: I'm probably missing something but why does this DAG not have this but the others do?
```
if __name__ == "__main__":
    dag.cli()
```

dags/atd_service_bot_intake.py
```diff
@@ -0,0 +1,78 @@
...
+}
+
+client: Client = new_client(ONEPASSWORD_CONNECT_HOST, ONEPASSWORD_CONNECT_TOKEN)
+env_vars = onepasswordconnectsdk.load_dict(client, REQUIRED_SECRETS)
+
+
+with DAG(
+    dag_id=f"atd_service_bot_issue_intake_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    schedule_interval="* * * * *",
```
Charlie-Henry: We did reduce this down to every 3 minutes frequency when we brought it over to Prefect, idk if we need it to run every minute again ⚡

dags/atd_service_bot_index_issues_to_dts_portal.py
```diff
@@ -0,0 +1,70 @@
...
+
+client: Client = new_client(ONEPASSWORD_CONNECT_HOST, ONEPASSWORD_CONNECT_TOKEN)
+env_vars = onepasswordconnectsdk.load_dict(client, REQUIRED_SECRETS)
+
+with DAG(
+    dag_id=f"atd_service_bot_issues_to_dts_portal_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    schedule_interval="13 7 * * *",
+    dagrun_timeout=timedelta(minutes=60),
+    tags=[DEPLOYMENT_ENVIRONMENT, "atd-service-bot", "github"],
```
mddilley: thought about this some more, and I'm not sure how useful having the environment in the tags is since it will be pretty clear if this is running in the local or production stack

dags/atd_service_bot_index_issues_to_dts_portal.py
```diff
@@ -0,0 +1,70 @@
...
+    },
+    "GITHUB_ACCESS_TOKEN": {
+        "opitem": "Service Bot",
+        "opfield": "shared.githubAccessToken",
+        "opvault": VAULT_ID,
+    },
+}
+
+client: Client = new_client(ONEPASSWORD_CONNECT_HOST, ONEPASSWORD_CONNECT_TOKEN)
+env_vars = onepasswordconnectsdk.load_dict(client, REQUIRED_SECRETS)
```
mddilley: Could be cool to add a 1PW util to import for this since we will be following this pattern in most DAGs. Something we could pass the `REQUIRED_SECRETS` to and get back `env_vars` (without having the same imports over and over)

mddilley:
Pushed a quick update because I liked Frank's suggestion about having folders to organize DAGs from the start. I named the folder after the repo where the scripts live like we did for Prefect 2. Open to anything other ideas, of course!

johnclary:
👋 i have a longer review in progress—but i'm here to speak out in favor of a mono-directory for DAGs, because nested folder structure inevitably leads to arbitrary organizing that is not fun to sort through if you're not a subject matter expert. if we focus on good tagging conventions, the Airflow UI should be great for sorting through DAGs. i also want to think of the VZ scripts as the only exception to the rule that files and supporting libraries should be stored outside this repo and brought in via docker containers.

mddilley:
@johnclary - thanks for dropping that comment. 🙏 I'm going to roll back the folder update to leave some space for discussion. I like your point about tagging. 🏷️

Charlie-Henry (commented):
Looks good to me, still poking around and learning lots so I'll just leave my comments for now. Thx @mddilley !

dags/atd_service_bot_intake.py
frankhereford: Charlie, that's a good question. Here's [a list of methods](https://github.com/1Password/connect-sdk-python/blob/main/src/onepasswordconnectsdk/client.py) available in the python library that we're using to access secrets in 1password. I have not tried them yet, but `get_item_by_title()` or `get_item_by_id()` may be able to pull in an entire entry.

I do see what you're saying about having to build out the object to define what secrets you want, and that is a new thing to do, but it does replace the work of building out the secret to store in airflow/prefect with all the items you would need. Kind of a wash, and even if there isn't a built-in method to get all the entries from a 1PW item, I bet we could build one.

Thanks for the good question! I am looking forward to seeing what we can come up with.

Charlie-Henry: `get_item_by_title()` worked like a charm @frankhereford!

https://github.com/cityofaustin/atd-airflow/pull/103

frankhereford: Nice!! I'm glad that did the trick - what a nice shortcut! I'm so glad you asked about this!!

frankhereford:
@johnclary wrote:

> i'm here to speak out in favor of a mono-directory for DAGs, because nested folder structure inevitably leads to arbitrary organizing that is not fun to sort through if you're not a subject matter expert. if we focus on good tagging conventions

Thanks John. I can absolutely see the value in not hiding files behind a purely-DTS known structure or hierarchy. I've advocated in the past for our public repos to be as agnostic to our specific implementations as possible, and I think your assertion aligns really well to that. The Airflow UI is the entry point into the list of DAGs, not the directory listing in `/dags`.

I also think your point about using quality tagging on our DAGs is really smart, and I'm 100% behind that. I think it's going to come out accessible and have a high quality user experience for us, and tags are huge part of making that happen.

dags/atd_service_bot_index_issues_to_dts_portal.py
johnclary: 100% - would be awesome to just call `get_env_vars(REQUIRED_SECRETS)` and get the results. and in that case we could  probably pull `VAULT_ID` out of the required secrets. 

also, do we want to be using the airflow `Variable` type for these env vars? i need to read through the rest of the orchestration code to understand how they are mounted into the env.

johnclary (commented):
this is awesome, @mddilley! thank you! i have not had a chance to run the local setup before i skip town, but wanted to share some feedback b/c deploying these DAGs is effectively going live in prod w/ this new stack.

i want to make sure we don't duplicate secrets in our new Airflow vault vs our existing 1pass vaults. how would y'all feel about this: if we add a secret to the new vault, we should track it down in any other vaults and replace the contents of the old 1pass item with a note that says something like "this secret now lives in the API vault under the name XYZ". i also left a comment down below about thinking through if/how to bundle related secrets.

i also see there there are some details in the readme that need to be filled in: like documenting where to find the env vars for deployment, how the webhook works, etc. i'm not saying that's on you Mike or that it belongs in this PR, but it would be could to talk through what loose ends we want to tie up before we go live. e.g. should we be reviewing in this PR the code for the rest of the stack, e.g. the webhook flask app?

thanks again Mike and Frank for putting this together!!

p.s. wrt to the DAG folder structure—i'm open to that discussion! in our brief run with prefect i definitely came to prefer our old Airflow way of the monodirectory.

dags/atd_service_bot_intake.py
mddilley: Thanks for flagging this. 🙏 This was just a bad copy/paste job on my part. The DAGs work without that last bit, and [this Airflow example of a legacy v1 DAG](https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/example_dags/tutorial_dag.html) does not include it so I'll take it out for the other two.

dags/atd_service_bot_intake.py
mddilley: I like this suggestion a lot so I'm going to push this change 🙏

dags/atd_service_bot_index_issues_to_dts_portal.py
mddilley: I think this is wise since it has bit us before elsewhere. @frankhereford - I would update here, but it might be good to make the change into the `airflow-v2` branch since this PR will probably take longer to merge since we have a lot of good discussion to consider. What do you think?

dags/atd_service_bot_issues_to_socrata.py
mddilley: +1 - I'm glad that I saw this in the legacy DAGs because it can cause some major unexpected behavior if left out!

dags/atd_service_bot_intake.py
mddilley: This is a great point about the downside of an entry per ETL, and something that Frank and I have touched on here and there. Finding good ways to prevent inadvertent duplication. I could see us getting into a jam where a password updates, and we only update one item for one ETL because we didn't know about another one elsewhere in this vault. 

I'm taking notes on all these good points that you are bringing up, and I'm going to float the idea of a near future Airflow 2 meeting today in dev sync. That way we can come away from with decisions on things and get some of the ETLs running. ⚙️

dags/atd_service_bot_index_issues_to_dts_portal.py
mddilley: I removed this from the tags 🧹

dags/atd_service_bot_index_issues_to_dts_portal.py
frankhereford: Yes, I am totally on board with pinning the requirements to a major version. I had intended to do that at some point, and just had let it slip though the cracks. 

I'll be happy to open a micro PR that does that pinning, and I'll get that done and up on the boards here momentarily.

Also yes - pendulum is something installed by the Airflow team in the base image on which we build our lightly customized image. It's always going to be available and we don't need to worry about requiring it.

dags/atd_service_bot_index_issues_to_dts_portal.py
mddilley: Agreed, and we could make it support both this `load_dict` method and the `get_item_by_title` method that Charlie used [here](https://github.com/cityofaustin/atd-airflow/pull/103)

dags/atd_service_bot_index_issues_to_dts_portal.py
frankhereford: PR https://github.com/cityofaustin/atd-airflow/pull/104, Pin Python Library Versions

dags/atd_service_bot_index_issues_to_dts_portal.py
```diff
@@ -0,0 +1,66 @@
...
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+
+from onepasswordconnectsdk.client import Client, new_client
+import onepasswordconnectsdk
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+ONEPASSWORD_CONNECT_HOST = os.getenv("OP_CONNECT")
+ONEPASSWORD_CONNECT_TOKEN = os.getenv("OP_API_TOKEN")
+VAULT_ID = os.getenv("OP_VAULT_ID")
```
frankhereford: There's nothing I would change here, but I did want to touch on these environment variable related assignments. These are not coming out of the airflow secret store, they are coming out of the airflow deployment itself. They come up from the `.env` file that is in the root of the checkout.

I like @johnclary's idea of having a helper function that we can use to abstract the retrieval of secrets down to a single call. That function would need access to these secrets, but that's totally fine. It would reduce the amount of boiler plate code that we see in every DAG, and I think that's never a bad thing. I'm 👍 on the idea.

dags/atd_service_bot_index_issues_to_dts_portal.py
```diff
@@ -0,0 +1,66 @@
...
+
+default_args = {
+    "owner": "airflow",
+    "description": "Create/update 'Index' issues in the DTS portal from Github.",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 12, 1),
```
frankhereford, line 16: I learned recently that the way Mike is doing this is 100% correct, and Airflow recommends that you do /not/ try to set the start date dynamically, as the start date is one of the things that it uses to identify DAGs.

Mike is establishing a great pattern here.

dags/atd_service_bot_index_issues_to_dts_portal.py
```diff
@@ -0,0 +1,66 @@
...
+        "opitem": "Service Bot",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.knackApiKey",
+        "opvault": VAULT_ID,
+    },
+    "GITHUB_ACCESS_TOKEN": {
+        "opitem": "Service Bot",
+        "opfield": "shared.githubAccessToken",
+        "opvault": VAULT_ID,
+    },
+}
```
frankhereford: I'm a huge fan of getting development vs. production secrets here with this f-string method. 👍

dags/atd_service_bot_index_issues_to_dts_portal.py
```diff
@@ -0,0 +1,66 @@
...
+
+with DAG(
+    dag_id=f"atd_service_bot_issues_to_dts_portal_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    schedule_interval="13 7 * * *",
+    dagrun_timeout=timedelta(minutes=60),
+    tags=["repo:atd-service-bot", "knack", "github"],
+    catchup=False,
+) as dag:
+    t1 = DockerOperator(
```
frankhereford, line 50: You mentioned that you had not refactored this to use the `@dag()` and `@task()` operators. I think the way you've done this is fantastic - there is no real upside to using the decorators as opposed to doing it this way. You might argue that using them can lead to shorter DAGs, but when you're porting an airflow-1 DAG to airflow-2, I think going with the "classic" syntax is great. Airflow 2 supports both fully.

dags/atd_service_bot_intake.py
frankhereford: Thanks @johnclary, that's a really good point.

Right now, the `DEPLOYMENT_ENVIRONMENT` gets its name from being a description of the intent of the particular airflow environment it's running in. Right now we have local ones, `development`, and a production one, `production`. 

I think this is different from different operating modes of a single DAG, which line up really well to the enumeration you stated: `production`, `staging`, & `development`. I think being able to make a DAG kick off in a certain mode is really important, and I don't think our stack is addressing it yet. I think it would be neat to have a DAG check its own tags looking for one that says `operating_mode: staging` or `operating_mode: production` and pick the secrets it needs or to modify its behavior based on that. The Airflow API is available to DAGs during execution, so I am hopeful that this is possible, but I've not tried anything like that yet. 

I'm not sure what is the best, but I'm very much looking forward to more discussion on how we should do this, or if I'm even framing the consideration correctly in my head.

dags/atd_service_bot_issues_to_socrata.py
```diff
@@ -0,0 +1,86 @@
...
+    },
+}
+
+client: Client = new_client(ONEPASSWORD_CONNECT_HOST, ONEPASSWORD_CONNECT_TOKEN)
+env_vars = onepasswordconnectsdk.load_dict(client, REQUIRED_SECRETS)
+
+with DAG(
+    dag_id=f"atd_service_bot_github_to_socrata_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    schedule_interval="21 5 * * *",
```
frankhereford, line 61: Not really important, but I'm not sure I understand the reason why we pick 21 minutes after the 5th hour for this one. It seems oddly specific. If there is a reason, having a comment in here to explain that would be helpful, or otherwise, maybe `0 5 * * *` would be better. 

If the worry is that we'll have a ton of DAGs fire up at once, then we should seek to address that via the [parallelism](https://github.com/cityofaustin/atd-airflow/blob/airflow-v2/airflow.cfg#L26-L30) configuration option or one of the related configurables.

frankhereford (approved):
Thank you Mike! You've set up a bunch of great patterns and sparked a ton of meaningful discussion. This is fantastic.

dags/atd_service_bot_index_issues_to_dts_portal.py
chiaberry: I never knew why we had to include start dates in our dags on airflow1, TIL its helps they identify DAGs

dags/atd_service_bot_issues_to_socrata.py
chiaberry: fwiw, one of the things John told me about writing dags on airflow was to try to not schedule them to run when other dags were running, to pick different times. so if I had to guess, I'd say someone picked this arbitrary time to not step on other dag toes.

dags/atd_service_bot_intake.py
frankhereford: I was thinking about this more, and we can totally have a second airflow stack running that has the environment set as "staging" and doing that might allow us to combine the concepts of the airflow deployment environment and the operating mode of DAGs. 🤔

dags/atd_service_bot_index_issues_to_dts_portal.py
mddilley: I created [this issue](https://github.com/cityofaustin/atd-data-tech/issues/12432) so the 1PW util can PR'd separately. cc @chiaberry  

I think that we should do this while we still have just a few DAGs so we can then use it going forward.

dags/atd_service_bot_index_issues_to_dts_portal.py
mddilley: Agreed - I created https://github.com/cityofaustin/atd-data-tech/issues/12432 to capture this while we still have just a few DAGs to update with a util.

mddilley:
I made another attempt at organizing these secrets and split out the Socrata, Zenhub, Github, and Knack details into separate entries in our vault. It still uses the environment when needed for the Socrata 4x4s and Knack apps (dev or prod).

dags/atd_service_bot_issues_to_socrata.py
```diff
@@ -0,0 +1,87 @@
...
+        "opvault": VAULT_ID,
+    },
+    "ZENHUB_ACCESS_TOKEN": {
+        "opitem": "Zenhub Access Token",
+        "opfield": ".password",
+        "opvault": VAULT_ID,
+    },
+    "SOCRATA_API_KEY_ID": {
+        "opitem": "Socrata Key ID, Secret, and Token",
+        "opfield": "add more.apiKeyId",
```
mddilley: I found out that 1PW somewhat awkwardly defines the section as `add more` when you click the `+ Add more` option in an entry without creating a separate section. Not sure what is better - creating a section only to not have `add more` in here or just go with this. 🤷

That would like like: `socrata.apiKeyId` (with `socrata` just being there to be there) vs. `add more.apiKeyId`

frankhereford: 🤮 Yech, that is awkward, I like your description, and I think it's too kind. Thank you for figuring it out! 1PW leaves no clues that I'm seeing.

It handles the `.username` and `.password` "sectionless entries" reasonably enough, I wish they would assume `.apiKeyId` was a top level entry as well. 

Anyway, I'm cool either way.. but I'm leaning towards making a named section in 1PW. My reasoning is that I think the 1Password UI is advertising it as being a top level entry without a section, but the API requires that you address it with an express, complete resource designation, and a sure-fire way to make that relationship 1:1 and clear is to use an explicit section in the 1PW GUI instead of the implicit one.

johnclary: agreed this is yech! i'm in favor of creating a section as well.

johnclary:
wrt to the weather DAG: is the intention for this to be a kind of health check for the airflow stack? i love it as a proof of concept, but seeing the DAG get it's own container, haproxy entry, and 1pass entry...it feels like a lot of code to keep around unless we have a good reason for it.

dags/atd_service_bot_issues_to_socrata.py
mddilley: cool y'all! i pushed an update with a section and also added a blurb about this in the [GitBook docs](https://app.gitbook.com/o/-LzDQOVGhTudbKRDGpUA/s/-M4Ul-hSBiM-3KkOynqS/1password-connect-server/usage#vault-entry-sections) 🙏

frankhereford:
@johnclary wrote: 
> is the intention for this to be a kind of health check for the airflow stack?

Hi John - yes, that is the intention. The DAG ends up testing a lot of the core functionality that we look for in the airflow stack. It checks the basic E, T, and L functionality, and airflow's larger functionality of running a scheduled job. Most importantly, it uses docker itself to process an image from an intersection camera by annotating it and adding an icon on it. Finally, it includes a visible output of a 1PW secret access. Seeing the weather output with a current timestamp coupled with street camera image on that page /also/ with a current timestamp is a indication that all of those things are working as intended.   

> DAG get it's own container, haproxy entry, and 1pass entry...

It does get its own barebones apache container to serve up the static content, and haproxy is configured to route requests to that apache server based on hostname, and there is a 1PW entry - that is all totally true and valid because they are resources, but I think the compute overhead is limited to running that Apache service and airflow kicking off the flow every 5 minutes, fwiw.

> it feels like a lot of code to keep around unless we have a good reason for it

My opinion is that it's sufficiently helpful in diagnosing airflow problems, or at least it was very helpful when putting all this together. If we don't want to keep it around, I can get that because it's always good to be lean because is less is more in terms of application code, but I'd observe that it pushes the task of knowing how to check that everything is working as intended onto the operator. It's sort of a canary in the coal mine, in that way.

I'd like to keep it, but I am totally open to giving it the ax if I'm over-valuing it. Thanks for looking at it and thinking about it!!

dags/atd_service_bot_intake.py
mddilley: This was added 🙏

chiaberry (approved):

Charlie-Henry (approved):

mddilley:
thanks y'all! I've added some items for further discussion to the [ETL meeting doc](https://docs.google.com/document/d/1aoZJacu9XxmyOTKLY88l1SfjOeqiAzeJGXS96m5Snu8/edit#)

PR #102, opened 2023-05-24
Add Slack notifications to Airflow v2

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12363

This PR adds Slack failure notifications by recycling the slack operator code from Airflow v1 along with a test DAG so we can verify that it is working. It also updates the service bot DAGs to use the notifier.

Afaik we have never used the success or critical failure handlers but it seems reasonable to keep them around if we ever need that.

## Testing

You can see this works by seeing the outputs from me running locally:
https://austininnovation.slack.com/archives/C013G4RNJ01/p1684944350489279
https://austininnovation.slack.com/archives/C013G4RNJ01/p1684946595554619

If you want to test this directly can also set this up locally by adding a connection in the local stack Airflow dashboard.

I added the config along with steps to the end of the password manager entry called **Airflow - Slack Bot**.  Once that is set up, you can run the DAG called `test_slack_notifier_development`.

## Ship list
- [x] Set up connection in production Airflow v2 dashboard


.gitignore
```diff
@@ -1,4 +1,4 @@
-dags/__pycache__
+*__pycache__
```
frankhereford, line 1: Good call!

dags/test_slack_notifier.py
frankhereford, line 1: This is premium - having a test fixture for the slack integration. I dig this!

requirements.txt
```diff
@@ -1,2 +1,3 @@
 onepasswordconnectsdk
 docker
+apache-airflow-providers-slack==7.3.*
```
frankhereford, line 3: I'm going to add the `stack-update` flag to this PR, if that's OK.

frankhereford (approved):
This is wonderfully done; thank you Mike! I've read over it, but I have not tested it -- your test that showed up in slack is plenty good enough IMHO. Thanks again!

dags/test_slack_notifier.py
mddilley: Totally took the lead from your motivation to include the weather DAG. 🧪 🙏

dags/utils/slack_operator.py
```diff
@@ -0,0 +1,81 @@
...
+
+# This is the Conn Id that we set when creating the connection in the Airflow dashboard
+# in Admin > Connections.
+SLACK_CONN_ID = "slack"
+
+
+def task_fail_slack_alert_critical(context):
+    slack_webhook_token = BaseHook.get_connection(SLACK_CONN_ID).password
+    slack_msg = """
+            <!channel> :red_circle: Critical Failure
```
chiaberry, line 12: does this mean @ the whole channel? 
I was going to propose a different icon to differentiate from normal failures, but if it alerts the whole channel, thats different enough.

chiaberry (approved):
I like that you included the option for success messaging and super critical failure, in case we need to use them sometime in the future. 

I saw your test messages in Slack, that and reading the code leads me to ✔️

dags/utils/slack_operator.py
mddilley: Yes, that is exactly what it translates it to in the Slack API. This was 100% copy/paste from the existing Slack operator code, and it is good to know that we have this option if we ever needed that. 🚨

dags/test_slack_notifier.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def task_fail():
+    raise Exception("Task failed")
+
+
+with DAG(
+    dag_id=f"test_slack_notifier_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    schedule_interval="*/5 * * * *",
```
johnclary: According to [the docs](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dag-run.html#cron-presets) we can set the schedule to be `None`, so that the DAG can only be manually triggered. That probably fits the intent of this DAG, to only be triggered as a test?

dags/test_slack_notifier.py
```diff
@@ -0,0 +1,40 @@
...
+    "start_date": datetime(2015, 12, 1),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+
+def task_fail():
+    raise Exception("Task failed")
```
johnclary: Maybe this should be "Task failed successfully" or super verbose "Task failure test successfully triggered" etc?

dags/utils/slack_operator.py
johnclary: whoa! good to know!

dags/test_slack_notifier.py
mddilley: i agree with this to avoid spamming the Airflow channel 📥

dags/test_slack_notifier.py
mddilley: This could definitely get confusing for others when triggered so I'm going with your second suggestion 🙏

dags/utils/slack_operator.py
```diff
@@ -0,0 +1,81 @@
...
+        exec_date=context.get("execution_date"),
+        log_url=context.get("task_instance").log_url,
+    )
+    failed_alert = SlackWebhookOperator(
+        task_id="slack_failure",
+        http_conn_id="slack",
+        webhook_token=slack_webhook_token,
+        message=slack_msg,
+        username="airflow",
+    )
```
johnclary, line 55: it would be really nice to get the actual python exception text in the slack message. i did a bit of searching (and recalling from memory) and i don't think there's a way to pull this out of the task `context`, and it would involve some heavy lifting. all that to say this looks great and replicates Airflow v1 perfectly ✨

johnclary (approved):
looks great + works great. thank you @mddilley! 🚢

johnclary (approved):
looks great + works great. thank you @mddilley! 🚢

PR #103, opened 2023-05-24
Trail Counters Data Publishing DAG

Charlie-Henry:
Just tested out migrating a simple Prefect flow to airflow 2 using Mike and Frank's work. Got this working locally 👍.

[Companion Prefect flow](https://github.com/cityofaustin/atd-prefect/blob/main/flows/atd-trail-counter-data/ped_data_pub.py).

Not critical that we merge this as it's running fine on Prefect 2, but I'm opening this just so we have it when we want to move it over.

Feel free to test this, running it won't break anything in terms of storing trail counter data.


.gitignore
```diff
@@ -7,3 +7,4 @@ weather/*.jpg
 venv
 *.swp
 *.http
+.idea/
```
Charlie-Henry: pycharm adds these

dags/ped_data_pub.py
```diff
@@ -0,0 +1,59 @@
...
+client: Client = new_client(ONEPASSWORD_CONNECT_HOST, ONEPASSWORD_CONNECT_TOKEN)
+
+# Gathering one pass entries into a dictionary
+vault_data = client.get_item_by_title(one_pass_entry_name, VAULT_ID)
+
+# Parsing object returned from vault into dictionary
+env_vars = {}
+for entry in vault_data.fields:
+    if entry.value:
+        env_vars[entry.label] = entry.value
```
Charlie-Henry: Code for getting all the items in a 1pass entry into a dictionary. This is returning all the items in the "Ped Data Pub" entry.

Charlie-Henry: @frankhereford @mddilley just tagging y'all

frankhereford: This is fantastic! I can't wait to review this!!

dags/ped_data_pub.py
```diff
@@ -0,0 +1,59 @@
...
+    "owner": "airflow",
+    "description": "Scrapes trail counter data from the public eco-counter website and publishes it in Socrata",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 12, 1),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+}
+
+one_pass_entry_name = "Ped Data Pub"
```
frankhereford: I think it would be worth considering giving this entry a more verbose name. Maybe something like "Trail Counter Publisher (ped_data_pub.py)"

dags/ped_data_pub.py
```diff
@@ -0,0 +1,59 @@
...
+        auto_remove=True,
+        command="python counter_data.py",
+        environment=env_vars,
+        tty=True,
+    )
+
+    t1
+
+if __name__ == "__main__":
+    dag.cli()
```
frankhereford: Since we have a local stack, I think it's increasingly less likely that we'll be running these scripts outside of the context of Airflow, so I'm not sure if we want to keep this pattern going, but it certainly doesn't hurt anything either. I'm good either way.

frankhereford (approved):
I think this looks great Charlie - thanks for porting this one!

chiaberry (commented):
I am eventually going to approve this, but I know its kinda in limbo since Mike's got a PR open making 1password util functions ala the functions you have here. 

Once more things get merged and the utils are used, I'll switch this to approved

dags/ped_data_pub.py
mddilley: This came from me making copy/pastes from the Airflow v1 DAGs. I removed it from mine so I think this should go. 🙏

dags/ped_data_pub.py
```diff
@@ -0,0 +1,59 @@
...
+for entry in vault_data.fields:
+    if entry.value:
+        env_vars[entry.label] = entry.value
+
+with DAG(
+    dag_id="atd_trail_counters",
+    default_args=default_args,
+    schedule_interval="00 8 * * *",
+    dagrun_timeout=timedelta(minutes=60),
+    tags=["atd-trail-counter-data", "Socrata"],
```
mddilley: I went with the `repo:<repo-name>` tag convention so I think this could be updated to keep things in sync.

mddilley (commented):
Ran this and it worked great! Thanks for jumping in and writing one of these! I ran it locally, and we are in the green. ✅

I know it isn't super urgent to get this in, but I wanted to drop links for the utils while it was fresh in my mind. https://github.com/cityofaustin/atd-airflow/tree/airflow-v2/dags/utils

dags/ped_data_pub.py
johnclary: @Charlie-Henry we have a separate entry in the 1pass vault for socrata api access. i suggest either using that, or extracting the socrata username/pw from this entry and creating a separate entry for it.

dags/ped_data_pub.py
```diff
@@ -0,0 +1,59 @@
+import os
+from datetime import datetime, timedelta
+
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+
```
johnclary, line 8: would you add the SlackOperator now that we have it available? See https://github.com/cityofaustin/atd-airflow/blob/b10bc9d82bddb60a20f43c1b5bce3a0e44d7d7a6/dags/atd_service_bot_intake.py#L7

johnclary (changes requested):

Charlie-Henry:
I'm going to go ahead and merge this in to `airflow-v2` and the next PR I will have to add some additional stuff like using the last execution date and John's changes to the onepass task.

PR #104, opened 2023-05-25
Pin Python Library Versions

frankhereford:
Based on a comment on PR #100 by @mddilley from @johnclary, this PR adds version pinning to the libraries we're bringing in on top of what is made available to us by the Airflow base image. These version choices are made by building the image a-new, and `grep`ing out the pertinent lines of `pip freeze` and then wildcarding after the major release.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-onepasswordconnectsdk
-docker
+onepasswordconnectsdk==1.*
+docker=6.*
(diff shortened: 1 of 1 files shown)


mddilley (approved):
Spun up the local stack, and I see the versions when I checked them. 🙏 🚀

roseeichelmann (approved):
til `pip freeze` :)

chiaberry (approved):

PR #106, opened 2023-05-31
Certbot scripts

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12329

This adds two bash scripts:
- a SSL cert renew script (uses 1Password Connect Server to grab IAM secrets needed) for a specified domain
- a script that runs the first script on all domains used by the stack, builds the haproxy image, and restarts it to start using the renewed cert 

## Testing

This might be one that is best to read through the code. I thought this would be a good one to see how our current and future Airflow setup SSL is handled as this was new to me. I'm happy to expand in the readme beyond the certbot link if it is unclear what is happening here. It also may need some adjustment when putting this on atd-data03, and I'll submit another PR if that is the case.

## Ship list
- [x] ~set up renew script cronjob as root on atd-data03 with logging~ added to https://github.com/cityofaustin/atd-data-tech/issues/12453


mddilley:
@frankhereford Could you take a first pass at this whenever you get a chance? Finally circled back on it and wanted to make sure the renew script aligns with what you were thinking when we discussed it a while ago. 🙏

toolbox/certbot/README.md
```diff
@@ -0,0 +1,15 @@
+# Airflow scripts on atd-data03
+
+Heads up that these scripts assume that the Airflow project lives in `/usr/airflow/atd-airflow/`.
```
frankhereford: I like how you've packed this location into an variable in the shell scripts though, this is still very flexible. 🙌

toolbox/certbot/renew.sh
```diff
@@ -0,0 +1,47 @@
...
+certbot/dns-route53 certonly --dns-route53 -d airflow.austinmobility.io
+
+cp /etc/letsencrypt/live/airflow.austinmobility.io/privkey.pem $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/cert.key
+
+cp /etc/letsencrypt/live/airflow.austinmobility.io/fullchain.pem $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/cert.crt
+
+cat /etc/letsencrypt/live/airflow.austinmobility.io/cert.pem > $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/airflow.austinmobility.io.pem
+
+cat /etc/letsencrypt/live/airflow.austinmobility.io/privkey.pem >> $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/airflow.austinmobility.io.pem
+
```
frankhereford: The haproxy image needs to be rebuilt because the SSL certificates are [copied into the image](https://github.com/cityofaustin/atd-airflow/blob/production/haproxy/Dockerfile#L3) at build time.

In theory, it'd be fine to stick a `docker compose build` in the restart script, but I'm inclined to invoke building only when it's expressly wanted by the developer/operator. It's fast, but not instant, on this stack. I sorta feel like requiring the build operation to be expressly requested by the operator is a good thing, but I could be swayed on that real easy.

toolbox/certbot/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Request certificate
+
+The request script has already been run when Airflow was set up on atd-data03. The initial certificate expires three months from its request, and the renew script is set to run every month after to keep it valid.
+
+## Renew certificate
+
+The renew script is set up as a cron job with a frequency of 0 0 1 * *.
```
frankhereford: Midnight on the first day of every month. I think this is exactly what would be ideal, but I'm thinking maybe asking for it at like 3AM or 5AM might be better because I can imagine that certbot service gets /really/ hammered at midnight on the 1st of the month. This is fine though, probably better to go for the ideal and shift it if requests time out or something.

That raises another point - we should be intentional and capture the output of these `cron`'d scripts in a log written to disk. just something like `0 0 1 * * /path/to/renew.sh >> /var/log/airflow_cert_renewal.log` would do it.

toolbox/certbot/renew.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+echo "Let's renew the certificate for airflow"
+export ATD_AIRFLOW_HOMEDIR="/usr/airflow/atd-airflow";
+
+# Load the same environment variables as the Airflow stack
+source $ATD_AIRFLOW_HOMEDIR/.env
+
+# Pull op v2 (latest is currently outdated and does not include the op read command needed below)
+docker pull 1password/op:2
```
frankhereford: This surprises me! Super nice catch. 

image

They don't subscribe to conventions about `:latest`, but they are just conventions, not requirements. `:latest` is just another tag like any other. The image you have listed is fully matured and been in service and receiving frequent updates for ages now. Thanks for this TIL!

toolbox/certbot/renew.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# Retrieve and store the AWS Access Keys from 1Password
+AWS_ACCESS_KEY_ID=$(docker run -it --rm --name op \
+-e OP_CONNECT_HOST=$OP_CONNECT \
+-e OP_CONNECT_TOKEN=$OP_API_TOKEN \
+1password/op:2 op read op://$OP_VAULT_ID/Certbot\ IAM\ Access\ Key\ and\ Secret/accessKeyId)
```
frankhereford: This is just a perfect example of going from a 1PW secret location into an environment variable in one shell command with no system requirements beyond docker. This is an excellent pattern.

toolbox/certbot/renew.sh
```diff
@@ -0,0 +1,47 @@
...
+
+AWS_SECRET_ACCESS_KEY=$(docker run -it --rm --name op \
+-e OP_CONNECT_HOST=$OP_CONNECT \
+-e OP_CONNECT_TOKEN=$OP_API_TOKEN \
+1password/op:2 op read op://$OP_VAULT_ID/Certbot\ IAM\ Access\ Key\ and\ Secret/accessSecret)
+
+# Now, remove the old concatenated certificates, renew the certificate, and replace with the new concatenated certificates
```
frankhereford: Before I comment on any other lines here, I want to share my understanding of the SSL certificate requirements that haproxy has. For every domain name that it is going to be able to terminate SSL for, it is going to look for a file called `<full_domain_name>.pem` that we've placed in the `haproxy/ssl/` folder. That single file contains two things:

* The private key generated at the creation of the CSR to be sent up the line to certbot. The private key never leaves our hands, just the CSR.
* The certificate itself issued by certbot after it validates ownership of the domain asset.

These two files are concatenated together in that order and given the name `<full_domain_name>.pem`.

toolbox/certbot/renew.sh
```diff
@@ -0,0 +1,47 @@
...
+CERT_PATH="/usr/airflow/atd-airflow/haproxy/ssl"
+cd $CERT_PATH
+rm cert.key
+rm cert.crt
```
frankhereford: Later on, we're going to pipe (`>`) the certificate onto the file location of the PEM. The `>` operator starts with truncating the file in the location, so I don't think we need to clean-up anything here. In a following comment, I'll suggest that we don't even need to write the `cert.key` and `cert.crt` files at all into haproxy's context.

toolbox/certbot/renew.sh
```diff
@@ -0,0 +1,47 @@
...
+docker run -it --rm --name certbot \
+-e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
+-e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \ 
+-v "/etc/letsencrypt:/etc/letsencrypt" \
+-v "/var/lib/letsencrypt:/var/lib/letsencrypt" \ 
+certbot/dns-route53 certonly --dns-route53 -d airflow.austinmobility.io
+
+cp /etc/letsencrypt/live/airflow.austinmobility.io/privkey.pem $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/cert.key
+
+cp /etc/letsencrypt/live/airflow.austinmobility.io/fullchain.pem $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/cert.crt
```
frankhereford: I don't think these files are required. I feel like I owe this some hands-on testing to make sure, but to do that, I'm going to need to do it on our production airflow machine since I've turned down my `airflow.fyi` original playground. If you've found that more than a `.pem` file is required for each domain, please let me know, and I'll get all over that to make sure I'm giving good information here.

toolbox/certbot/renew.sh
```diff
@@ -0,0 +1,47 @@
...
+rm cert.crt
+
+docker pull certbot/dns-route53:latest
+
+docker run -it --rm --name certbot \
+-e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
+-e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \ 
+-v "/etc/letsencrypt:/etc/letsencrypt" \
+-v "/var/lib/letsencrypt:/var/lib/letsencrypt" \ 
+certbot/dns-route53 certonly --dns-route53 -d airflow.austinmobility.io
```
frankhereford: A little bit of me wants to say we should inspect the return value of this command, and if it's anything except `0`, we should bail on the following commands, but, the old, working cert that wasn't renewed will simply be reinstalled, so no harm no foul. I think this is good.

toolbox/certbot/renew.sh
```diff
@@ -0,0 +1,47 @@
...
+
+cp /etc/letsencrypt/live/airflow.austinmobility.io/privkey.pem $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/cert.key
+
+cp /etc/letsencrypt/live/airflow.austinmobility.io/fullchain.pem $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/cert.crt
+
+cat /etc/letsencrypt/live/airflow.austinmobility.io/cert.pem > $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/airflow.austinmobility.io.pem
```
frankhereford: `>` is overwriting the `.pem` file. 👍

toolbox/certbot/renew.sh
```diff
@@ -0,0 +1,47 @@
...
+
+cat /etc/letsencrypt/live/airflow.austinmobility.io/privkey.pem >> $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/airflow.austinmobility.io.pem
```
frankhereford: `>>` is appending to the end of the `.pem` file. 👍

toolbox/certbot/renew.sh
frankhereford, line 1: I have a fair number of comments on this file, but please don't take that to mean it's not fantastic work. It's really good, and it would renew airflow.austinmobility.io as-is.

I do think we have a requirement that it's not addressing though. The production airflow installation currently is serving four domain names:
* `airflow.austinmobility.io`
* `airflow-webhook.austinmobility.io`
* `airflow-workers.austinmobility.io`
* `airflow-weather.austinmobility.io`

I think this script could be adapted to take the domain name that is attempting to be registered as a command line argument and renamed `renew_domain_with_certbot.sh` or something along those lines. That then coupled with a 5-liner shell script that just invokes the main script four times, once for each domain needed could be written and that tiny script put in cron. It's that second short invocation script that would then also be used to rebuild the haproxy image (`docker compose build haproxy`) and then call the `restart_airflow.sh` script only after the renewals are complete. Really even, it could just restart the haproxy container after its rebuilt, which would be pretty slick.

_Caveat emptor_: `docker compose restart service-name` restarts the current running container with the **same** image that it was running from, even if a newer image has been built while it was running. `docker compose stop service-name; docker compose start service-name` is not the same as `docker compose restart service-name` when trying to update a container to be running from a updated image.

toolbox/certbot/renew.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+echo "Let's renew the certificate for airflow"
+export ATD_AIRFLOW_HOMEDIR="/usr/airflow/atd-airflow";
+
+# Load the same environment variables as the Airflow stack
+source $ATD_AIRFLOW_HOMEDIR/.env
```
frankhereford: Really nice; less .env files out there to keep track of is premium!

toolbox/certbot/request.sh
frankhereford, line 1: Check my math here, but I think we can get away with removing this script entirely. The initial certificate allocation and a renewal of an existing cert may be handled differently internally inside the certbot command and its conversation with the certbot service, but the invocation of that command looks to be the same. We run it the same way with this docker command and we get the keys out of its output and into our haproxy build location the same way both times. Do you agree?

toolbox/certbot/restart-airflow.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -o errexit
```
frankhereford: I really like this ability here to have a shell (read: script) bail if any of the commands that are executed fails. 

But in this case, only the docker commands have any meaningful risk of failing. If the `cd` or the `echo` commands don't succeed, we have much bigger problems than our airflow install. 

The docker commands - if the `down` fails, then I feel like it's still prudent to continue on with the `up` operation. 

This is super edge, niche case stuff though. I wouldn't give this more than a second's thought and hit the resolve button with no regrets for this comment if you want :)

frankhereford (commented):
Mike, I am sorry for a 15 comment review, but I wanted to give you as much detail and feedback as I could since you had asked me to give it a first pass. 

I think this is awesome stuff, and it's far-and-away already the nicest tooling we've had around certbot and I know we'll be using it here and reusing it in many projects into the future.

Please let me know about anything that I was unclear on, or if you'd like to discuss any of this. Thank you for this great work!

toolbox/certbot/renew.sh
mddilley: yeah! I wouldn't have caught this without the trouble running the `read` command. `latest` kept pulling v1 which I also found super surprising and took me a while to realize what was going on!

toolbox/certbot/renew.sh
mddilley: cool, i initially removed the writing of these files, but then went back on it to get closer to the existing script. I'll go with the change that you are suggesting to simplify this process. 🙏

toolbox/certbot/renew.sh
mddilley: That is my suspicion as well. I started to think that they might be needed as a intermediary placeholder, but I also think we can do away with them altogether. I'll do some testing.

toolbox/certbot/request.sh
mddilley: I'm on the fence. I included it just as reference to what was already run when for the initial request. Having the examples that you shared with me was really helpful when I doing this part.

toolbox/certbot/renew.sh
mddilley: This is a great idea, and thanks for pointing out that we have more to worry about than just the airflow domain. I'm going to refactor this so that we can take care of all of them at once. 🙏

mddilley: Also, with these updates, I'm not even sure if we need a "restart the whole stack" script. I think that I like being explicit that we are restarting haproxy to use the renewed certs.

toolbox/certbot/request.sh
frankhereford: I dig that. There's no harm in keeping them around, that's for sure. Maybe a comment in a readme that the renewal and the initial issuance commands are formed the same, and that it's only internal to the certbot system that they are handled differently?

toolbox/certbot/README.md
mddilley: Cool, I changed the readme to just say monthly in case we need to adjust it some and also noted the logging that we can put into place when setting this up on the server.

toolbox/certbot/request.sh
mddilley: Oh, I gotcha. I misread your first comment, and I'm just now realizing that the certbot command is the same for both. 🙃 I'll remove the request script to avoid any confusion and add a note to the readme. 🙏

frankhereford (approved):
I think this is ready for larger review anytime you're game. This looks ship shape to me!

toolbox/certbot/renew.sh
mddilley: Just updating that this worked out when removing the intermediate files. 🚀

toolbox/certbot/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+For SSL termination management of the resources in this Airflow stack, we use [certbot](https://certbot.eff.org/).
+
+## Request certificate
+
+The request script has already been run when Airflow was set up on atd-data03. The initial certificate expires three months from its request. The request command is identical to the certbot command used in the renew script.
+
+## Renew certificate
+
+The renew script is set up as a cron job with a monthly frequency with output logged to `/var/log/airflow_cert_renewal.log`. The `renew.sh` script renews all of the domains used by the stack using the `renew_domain_with_certbot.sh` script that takes one argument like:
```
frankhereford: One thing that just occurred to me that you might want add here is that the renewal script must be run as root. Certbot's output in `/etc` is owned by root and set so that only root can read it because it contains private keys and other sensitive information. 

You could maybe even put something like this at the top of the script to remind users if they forget:

```
# Check if the script is running as root
if [ "$EUID" -ne 0 ]
  then echo "Please run as root"
  exit
fi
```

chiaberry (approved):
Read the code, thanks for the comments in the scripts, makes it easier to follow for sure. 

Just have one clarifying question, but this feels good to go.

toolbox/certbot/README.md
chiaberry: The script renews monthly even though the certs last for 3 months? 

Also thank you for spelling out where to find the logs!

frankhereford: I think I may have originally suggest the monthly timing on the renewal process, and to be honest, after you asked, I was questioning it also that it might be too frequent too. 

I looked for some advice on stack overflow, and I kept seeing suggestions that have the renewal cronjob set to even more frequent intervals, and then I found this in their [official documentation](https://eff-certbot.readthedocs.io/en/stable/using.html#automated-renewals) (edited for clarity):

```
0 0,12 * * * root certbot renew -q
```

I think because certbot will intelligently inspect the certificate in place and return without doing anything if it doesn't need to be renewed, they suggest a 12-hour interval for the renewal cron entry. That said, I think this gives them room to reduce the amount of time their certificates are issued for without breaking people's installs, and as long as they are giving out 3 month certs, monthly still seems alright to me.

toolbox/certbot/renew.sh
```diff
@@ -0,0 +1,14 @@
...
+
+echo "Renewing the certificates for the Airflow stack"
+/usr/airflow/atd-airflow/toolbox/certbot/renew_domain_with_certbot.sh airflow.austinmobility.io
+/usr/airflow/atd-airflow/toolbox/certbot/renew_domain_with_certbot.sh airflow-webhook.austinmobility.io
+/usr/airflow/atd-airflow/toolbox/certbot/renew_domain_with_certbot.sh airflow-workers.austinmobility.io
+/usr/airflow/atd-airflow/toolbox/certbot/renew_domain_with_certbot.sh airflow-weather.austinmobility.io
+
+# Build the HAProxy image with the renewed certificates
+cd /usr/airflow/atd-airflow
+docker compose build haproxy
```
johnclary, line 11: rather than build these into the image, would it be better/safer to mount them into the container on start?

johnclary: i'm kind of thrown off that the image build name in the docker compose name seems to shadow the base `haproxy` image name. i'm not saying this is actually what's happening or it's a bad pattern in any way, but it has me thinking about if we want to use github actions + dockerhub for building these images so that we can better keep track of which build is running on prod vs local. i realize is outside the scope of this issue and i'm happy to make an issue!

toolbox/certbot/renew.sh
johnclary: and a good reminder why it's never a good idea to pin to `latest` in a production use case: you never know what you're gonna get!

toolbox/certbot/renew_domain_with_certbot.sh
```diff
@@ -0,0 +1,42 @@
...
+-e OP_CONNECT_HOST=$OP_CONNECT \
+-e OP_CONNECT_TOKEN=$OP_API_TOKEN \
+1password/op:2 op read op://$OP_VAULT_ID/Certbot\ IAM\ Access\ Key\ and\ Secret/accessSecret)
+
+# Now, remove the old concatenated certificates, renew the certificate, and replace with the new concatenated certificates
+CERT_PATH="/usr/airflow/atd-airflow/haproxy/ssl"
+cd $CERT_PATH
+rm $DOMAIN.pem
+
+docker pull certbot/dns-route53:latest
```
johnclary: do we want to pin the certbot version here?

johnclary (changes requested):
Lots of TILs in here for me: thank you Mike and Frank!

in addition to my docker related questions, i have to ask: should we consider packaging this into a DAG? that way we'd get all the benefits of being able to monitor it on Airflow + slack failure notifications.


...and as i type this i guess that last part of running `docker restart` is kind of the dealbreaker?

frankhereford:
@johnclary wrote:

> should we consider packaging this into a DAG?

I would totally be open to doing this, except, I'm not sure if we can reliably restart the airflow stack from inside a DAG. We might be able to rig up something that looks for a file that the DAG leaves out there or something like that, but otherwise, I think this might be easier and simpler to leave outside the airflow context.

johnclary:
thanks @frankhereford—agreed. fwiw i would prefer to see the certs mounted into haproxy through a volume rather than built-in. but that wouldn't solve this issue without telling the haproxy api to load the new cert ([article](https://www.haproxy.com/blog/dynamic-ssl-certificate-storage-in-haproxy)).

TLDR: i'm in favor of keeping this as a shell script

toolbox/certbot/renew.sh
frankhereford: 👍 I am totally OK with changing the name of the `haproxy` service inside the `docker-compose.yml` file to differentiate it from the base image that it originates from. 

👍 If we wanted to publish the haproxy image, then I think your idea about getting the keys into it at run time and not built into the image is a must have. It's for the best to do it that way really - it makes it impossible to leak them via pushing the image.

To your question about building this image with GitHub actions, the event that the build is responding to is actually the certbot renewal event, and not a PR getting merged into a certain branch. That's not to say we can't add some `docker push` commands into the script to keep a copy of the image on dockerhub if we want to go the publishing route.

To the question of if we want to push these images up to docker hub, I think that would be OK. If the intent is to run the same images in production and in local development, we'll probably have some trouble because we'll need probably want to build multi-architecture images, and more importantly, there are a handful of changes that differ between the soon-to-be `main` branch (currently `airflow-v2`) and `production`. 

Here are some details on the differences between `production` and the main branch. `production` has [a commit](https://github.com/cityofaustin/atd-airflow/commit/a855cf95012ad84230e29a1c97bea9f9146a2ce1) ahead of the main branch, with the changes shown here: https://github.com/cityofaustin/atd-airflow/compare/airflow-v2...production. It will also accumulate merge commits ahead of the main branch as we merge the main branch into it, but that's correct as the record of merging into `production`. 

This commit ahead of the main branch contains the changes to the main branch that configure it to run in production being the port configuration and the SSL configuration. If we were to merge this back into main, we'd loose the feature of always being able to checkout the main branch and spin it up locally.

johnclary: Frank and I had a chat about some ways forward. I will open a separate issue that recaps that conversation 🙌

johnclary: I created [#12552](https://github.com/cityofaustin/atd-data-tech/issues/12552) for unifying our git/docker setup for local dev.

That issue does not mention bind-mounting the SSL certs via docker compose: We probably want to include that change in this PR?

toolbox/certbot/renew_domain_with_certbot.sh
mddilley: Yep, that is a great idea considering that I did that for the 1PW CLI 🙏

toolbox/certbot/renew.sh
mddilley: Thank y'all for discussing this here as I hadn't really thought about alternatives to having some extra config that makes prod. I like the idea of splitting out into prod and local Docker compose files. 🚀

I just pushed an update to bind mount the ssl directory into the haproxy container. ~Are we looking to update the way this folder is copied into the container in the production haproxy `Dockerfile` on atd-data03 now or wait until we work on the issue John linked?~ _Edit: I looked at that issue again, and I noticed that it specifically links to that line in the haproxy Dockerfile._

toolbox/certbot/README.md
mddilley: I ended up adding this to the readme. Thanks for the reminder! 🙏

johnclary (approved):
🚢 

thanks Mike!!

toolbox/certbot/README.md
johnclary: we discussed today that we would like to copy EFF and run this 12-hourly

mddilley:
@mddilley update the readme to say the cron schedule is every 12 hours

PR #107, opened 2023-06-01
Make 1Password util for Airflow 2 DAGs to grab secrets

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12432

This PR adds a 1Password utility to reduce lib imports and config in DAGs. I added a bit of info on the utils so far. Open to any feedback on the format of that info. I know the 1PW discussion is on-going. 🙏

## Testing

1. [Spin up the local stack
](https://github.com/cityofaustin/atd-airflow/blob/airflow-v2/README.md#local-setup)
2. This is working if the service bot DAGs are visible in the Airflow dashboard and there are no DAG import errors.

PR #108, opened 2023-06-07
DAG to invoke CRIS import script

frankhereford:
## Associated PRs

This PR has a sibling PR in the VZ repository which contains the image that is invoked in this DAG. You can find it here: https://github.com/cityofaustin/atd-vz-data/pull/1235.

## Testing

We have a few options here. This can be a review the code and rely on my testing, or I would be more than happy to give a demonstration of this working on my local VZ & airflow stacks. I can do that individually or maybe even as a dev-sync share if that would be helpful. 

Of course, individual testing is always welcome! Here is the general outline of how to do it. There are shared resources in here like the SFTP endpoint, so you may want to mention that you're doing this testing in the dev channel or something so two people don't try to do this at the same time.

* Modify your local VZ stack to expose the `graphql-engine` on port 8084, because 8080 will be used by the airflow web ui.
* Spin up the local VZ DB and populate it with data from the read replica.
* Spin up the local `graphql-engine` instance, now listening on 8084 of your machine.
* Bring up the local airflow stack on this branch. 
  * We have some other DAGs in flux in terms of what they are expecting in 1PW and what is available, so I had to remove them from my local checkout so I didn't get import errors in the airflow UI.
* Get on the SFTP endpoint and use `sudo` to become root so you can jump into `txdot`'s home directory. See that it is empty. If it's not, you're probably at work really early (or up after midnight working late) and looking at a file that will be imported that day at 7AM. This file is important, and it's probably easiest if you leave it in place until the ETL consumes it at 7AM.
* Sign into Chia's CRIS account, and request an export be uploaded. From memory, the parameters you need are:
  * CSV export
  * Process date (not crash date)
  * Tri-county area (Wilco, Travis & Hays)
  * Extract password out of 1PW
  * A process date interval of your choosing, but it must be entirely in the past and not include today. The longer the interval you pick, the longer your testing is going to take BTW. A day or two is good.
* Watch that `/home/txdot` folder with something like `while sleep 5; do ls -l; done` and wait for the file to be delivered - normally 15-30 minutes.
* And finally, you can trigger the DAG. 
  * It'll pull the import image of docker hub, and the image is built with both X86 and ARM chip support, so this should work natively on modern macs and linux machines. 
  * If you are on a linux machine though, you'll need to make some changes in the 1PW entry or docker-compose file that remove the reliance on the use of the magic DNS entry provided by Docker Desktop: `host.docker.internal`.
* You should see the DAG complete, and you should have updated data in your database and you also should see the file removed from the SFTP endpoint.

## ✅ ~Merge Warning~

This branch has had both the [md-12432-1pw-util](https://github.com/cityofaustin/atd-airflow/pull/107) and the [md-12363-slack-notifier](https://github.com/cityofaustin/atd-airflow/pull/102) branch merged into it.  It should not go in until both of those PRs have been merged.

- [x] Slack Notifier PR
- [x] 1Password Utility PR


dags/vz_cris_import.py
frankhereford, line 1: This is the only file that needs review in this PR. Everything else came in from merging the slack and 1pw utility branches into this so I could use them in this script.

.gitignore
frankhereford, line 1: Every file except for the `vz_cris_import.py` DAG comes into this branch via merging in some other in-development branches that were required. Please feel free to skip everything else in your review.

dags/vz_cris_import.py
```diff
@@ -0,0 +1,39 @@
...
+    start_date=datetime(2023, 6, 1),
+    catchup=False,
+    tags=["vision-zero", "cris"],
+    on_failure_callback=task_fail_slack_alert,
+)
+def cris_import():
+    @task()
+    def invoke_cris_import_via_docker():
+        client = docker.from_env()
+        docker_image = "atddocker/vz-cris-import:latest"
```
chiaberry, line 23: will there be a need for `vz-cris-import:production` and `vz-cris-import:staging`, or will we only ever need to use the latest tag?

dags/vz_cris_import.py
```diff
@@ -0,0 +1,39 @@
+import os
+import docker
+from datetime import datetime
+
+from airflow.decorators import dag, task
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.environ.get("ENVIRONMENT", 'development') # operating environment
```
chiaberry, line 8: Should this be updated to be prod?

chiaberry: or "production" i mean, I dont have a preference on the shorter version or spelled out

dags/vz_cris_import.py
frankhereford: We don't really have a good setup to run a staging server of this because we only have one SFTP endpoint. We could set one up, but this ETL has never really had a concept of it that I've been aware of.

dags/vz_cris_import.py
frankhereford: Here 'development' is the default value that is assigned into the `DEPLOYMENT_ENVIRONMENT` variable when the containers environment doesn't provide anything. It will be set to production when the `ENVIRONMENT` variable contains that like on `atd-data03`. Our local stacks have it set to `development`, or if it's not defined, this line acts as a back-up.

dags/vz_cris_import.py
frankhereford: Thank you for that question Chia. I had a comment on here earlier that I removed, because I think I had missed the point of your question. Sorry about that!

I think we can get by without a different build for this in production and staging. I think that the same script and image works for both situations and is entirely based on if the secrets coming out of 1PW are under the `development` or `production` sections.

dags/vz_cris_import.py
chiaberry: derp, yeah this is the default and not what its being set to. Totally forgot how this command worked

dags/vz_cris_import.py
```diff
@@ -0,0 +1,39 @@
...
+
+@dag(
+    dag_id="vz-cris-import",
+    description="A process which will import the VZ CRIS data into the database on a daily basis using data on the SFTP endpoint",
+    schedule="0 7 * * *",
+    start_date=datetime(2023, 6, 1),
+    catchup=False,
+    tags=["vision-zero", "cris"],
```
chiaberry: could you also add a tag `repo:atd-vz-data`?

chiaberry (approved):
This is my first time seeing a dag written with the decorators, nice to see the different style. 

Left a note about tags, I _think_ we agreed to adopt a `repo:<repo-name>` tagging convention, but if im misremembering, ignore the suggestion.

dags/vz_cris_import.py
frankhereford: Great call, thank you for asking. It's in there!

mddilley:
@mddilley update the readme to say the cron schedule is every 12 hours

dags/atd_service_bot_index_issues_to_dts_portal.py
```diff
@@ -34,6 +34,7 @@
     "GITHUB_ACCESS_TOKEN": {
         "opitem": "Github Access Token Service Bot",
         "opfield": ".password",
+
```
mddilley: This change snuck in here in here 👀

mddilley (approved):
Looks great! I did not test this out locally, but I do think that it would be good to capture your test steps in Gitbook or somewhere. Even just a copy/paste for now would be great. I have some rough notes of my own, but yours are much better! 🚢

dags/atd_service_bot_index_issues_to_dts_portal.py
frankhereford: Wow, nice catch! I looked, and I brought that in when I merged `airflow-v2` into this branch at some point, and then it was corrected in `airflow-v2` since. Great catch; this would have undone someone's work!

patrickm02L (approved):
Adding my concurrence based on Dev reviews.

PR #109, opened 2023-06-09
Refine Airflow 2 readme

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12454

Tagged reviewers to get this on all our radars. Feel free to offer suggestions or take on a TODO.

TODOs
- [x] Point to the 1PW entry that contains the needed secrets for the env file:
- [x] Add some more details about local dev
- [x] How to rebuild/restart when updates are made to the stack (not needed when only adding/updating a DAG only) @frankhereford 
- [x] Cover webhook purpose and setup @frankhereford 
- [x] Cover why we need HAProxy and how it works @frankhereford 
- [x] Developing a DAG - where to add a file, tags, etc.? @mddilley
- [x] Add PR template
- [x] Update to reflect the removal of webhook and weather DAG

## Testing
n/a


pull_request_template.md
```diff
@@ -0,0 +1,13 @@
+## Associated issues
+
```
chiaberry, line 2: What do you think about adding a section to link the the repo that contains the scripts the dag runs? I can't think of a good name but something like

## Associated Repo / Companion Repo / Code Repo

frankhereford:
@mddilley and everyone:

I tagged myself on 3 of the unchecked items in the list Mike has provided. I think I'm a good fit to write about those, but please let me know if you would prefer to work on any of them. My goal is to provide one section per day.

Mike, is it OK if I commit my changes onto your branch? I can make a new feature branch too if you'd like too, either way is cool with me.

mddilley:
@frankhereford - thanks for picking those up 🙏, and let me know if I can help out with any of them. Also, it is totally cool with me to commit to this branch so we'll be able to review and merge this all together. Thanks!

pull_request_template.md
mddilley: I think that this would be a great reminder to point reviewers to the code that a DAG runs! 🙏

README.md
```diff
@@ -90,13 +146,26 @@ The 1Password utility is a light wrapper of the [1Password Connect Python SDK](h
...
-* Create remote worker image example
-  * Use `docker compose` new `profile` support
-* Add slack integration
-* 🤔 Extend webhook to rotate key in 1Password after every use
-  * a true rolling token, 1 use per value
-
-## Example DAGs
-* You can turn on [this field](https://github.com/frankhereford/airflow/blob/main/docker-compose.yaml#L65) to get about 50 example DAGs of various complexity to borrow from
+
+- Make it disable all DAGs on start locally so it fails to safe
```
chiaberry, line 148: what does "so it fails to safe" mean?

frankhereford: This is a comment from my notes-to-self thoughts made during the initial development that has made it here into this README. It can certainly be improved in how it shares this idea:

I had imagined a mechanism that detects a start-up of the airflow stack with the environment context set to `development`. When this happens, this mechanism would run though all DAGs which are installed on the local stack and turn them off. The idea being that this eliminates the risk of a DAG which updates an outside data source or target from running before the developer realizes it and intervenes. Certainly, we'd never knowingly setup a DAG that does this, but I could imagine something being mid-development that might do something on execution that a developer didn't intend. 

I'd prioritize this enhancement at "ultra-low" and also be totally fine striking it from the ideas list.

chiaberry: ah that makes sense. I just didnt understand what "fails to safe" means, but I can see how preventing dags from running unexpectedly would be helpful.

.DS_Store
frankhereford, line 1: # 🧹🧹
 I snuck in a few housekeeping actions onto this PR. In retrospect, seeing that I found 4 of them, I should have made another PR. This was my mistake, sorry for the noise!

.github/.DS_Store
frankhereford, line 1: 🧹🧹

.github/workflows/sanity-check.yml
frankhereford, line 1: I struck this GitHub action. It was a very "frank-thing" that hung on from when I was developing this initially. It was a workflow that would prevent merging of PRs with certain labels, and I think it's a problem we're not going to bump into with our workflow.

README.md
```diff
@@ -1,85 +1,141 @@
...
-_AIRFLOW_WWW_USER_PASSWORD=<pick your initial admin pw here>
-AIRFLOW_PROJ_DIR=<absolute path of airflow checkout>
-OP_API_TOKEN=<Get from 1Password here: 'name TBD'>
-OP_CONNECT=<URL of the 1Password Connect install>
-OP_VAULT_ID=<OP Vault ID>
+_AIRFLOW_WWW_USER_PASSWORD=<Pick your initial admin password here>
+AIRFLOW_PROJ_DIR=<The absolute path of your Airflow repository checkout>
+OP_API_TOKEN=<Get from 1Password entry named "Connect Server: Production Access Token: API Accessible Secrets">
+OP_CONNECT=<Get from 1Password entry named "Endpoint for 1Password Connect Server API">
+OP_VAULT_ID=<Get from 1Password entry named "Vault ID of API Accessible Secrets vault">
```
roseeichelmann, line 25: this is really helpful documentation on where to get the passwords!

README.md
```diff
@@ -1,85 +1,141 @@
...
-  * This will reset the history of your dag runs and switch states.
+Now,
+
+- Airflow is available at http://localhost:8080
+  - You can log in to the dashboard using the username and password set in your `.env` file
+- The webhook flask app is available at http://localhost:8081
+- The workers' status page available at http://localhost:8082 
+- The test weather DAG output is available at http://localhost:8083
+
+### Developing a DAG
```
roseeichelmann: i feel like this could be called "Executing a DAG"? "developing" makes me think of the creation of the DAG in the first place. nitpicky and i dont care too much either way so this doesnt have to change :p

pull_request_template.md
mddilley: I went ahead and pushed this before it gets lost in the mix. Thanks @chiaberry!

README.md
```diff
@@ -1,85 +1,141 @@
...
-* This is not a part of building or operating the stack locally.
+# DTS Airflow stack
 
-```
-docker buildx build \
---push \
---platform linux/arm/v7,linux/arm64/v8,linux/amd64 \
---tag frankinaustin/signal-annotate:latest .
-```
+This stack is used to run DTS ETL processes and the production instance is deployed on `atd-data03`. Local development is available, and instructions are below.
```
johnclary: i'm happy to commit this change myself in a bit, but before i forget: 
```
deployed on `atd-data03` at `/usr/airflow/atd-airflow`
```

README.md
mddilley: Not nitpicky at all! this section was intended to say where to put files, utils, etc for them to be usable, but I think that adding some details about how to run the DAGs and how that relates to the dashboard would be good too. 🙏 I'll roll that into this section and update the title.

README.md
mddilley: thanks John! good call because I found myself poking around when looking for the existing Prefect and Airflow checkouts on-prem. 🎯 Just added this with another edit.

mddilley: thanks John! good call because I found myself poking around when looking for the existing Prefect and Airflow checkouts on-prem. 🎯 Just added this with another edit.

frankhereford:
I pushed up some commits that updated the readme to remove info about the webhook and weather DAG. Thanks!!

frankhereford (approved):
🚢 🚢  Thank you!!! 🚢 🚢

mddilley:
@frankhereford my last commit is my attempt to cover the auth differences for the Flower service set up in the haproxy config - feel free to reword!

README.md
```diff
@@ -90,13 +127,24 @@ The 1Password utility is a light wrapper of the [1Password Connect Python SDK](h
...
+
+- ⛔ Stop all containers and execute this to reset your local database.
+  - Do not run in production unless you feel really great about your backups.
+  - This will reset the history of your dag runs and switch states.
+
+```
+docker compose down --volumes --remove-orphans
+```
+
 ## Ideas
```
mddilley, line 146: Just caught this after scanning through. Should we capture these in issues and remove this part?

johnclary (approved):
thank you Frank and Mike! this looks awesome!  

I have a few minor additions I'd like to make but will suggest those through a separate PR.


🚢🚢🚢🚢🚢🚢🚢🚢

README.md
frankhereford: I think this is actually valuable to keep around for local stack users. We'd obviously never do it in production, but it can be nice to give yourself a tabula rasa locally. I'm game either way though.

PR #111, opened 2023-06-15
Adds production compose file and reworks some settings

johnclary:
### Associated Issues
- https://github.com/cityofaustin/atd-data-tech/issues/12552
- https://github.com/cityofaustin/atd-data-tech/issues/12557

Here's my crack at updating our stack to have a separate dev + prod configuration. I've also moved some settings into the `airflow.cfg`. I'd very much like your feedback on the settings I tweaked: I mostly wanted to point out that some of the defaults seem iffy but I'm not confident in my choices.

### Test

You can test local stack like normal (`docker compose up`), but you should add a fernet key to your env:
```
# .env
# this key should not be used in production
_AIRFLOW__CORE__FERNET_KEY=PTkIRwL-c46jgnaohlkkXfVikC-roKa95ipXfqST7JM=
```

If you're tweaking any of the config settings, don't forget to `docker container prune` between stopping and starting.

If you'd like to test the production config overrides that command is:

```
$ docker compose -f docker-compose.yaml -f docker-compose-production.yaml up
```

And you can observe the prod haproxy port mappings via `docker ps`. I guess I'm just assuming this will work?

## Todo
- add new fernet key env var to README
- create and install a production fernet key (and save it in 1pass)
- figure out celery flower basic auth (comment below)


docker-compose.yaml
```diff
@@ -18,8 +18,6 @@
 
 # Basic Airflow cluster configuration for CeleryExecutor with Redis and PostgreSQL.
 #
-# WARNING: This configuration is for local development. Do not use it in a production deployment.
-#
```
johnclary, line 22: i've been looking through this article and i think we're checking the boxes:
https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/production-deployment.html

docker-compose.yaml
```diff
@@ -146,15 +143,15 @@ services:
...
+
   postgres:
     image: postgres:13
     environment:
       POSTGRES_USER: airflow
       POSTGRES_PASSWORD: airflow
       POSTGRES_DB: airflow
     volumes:
-      - postgres-db-volume:/var/lib/postgresql/data
+      - ${AIRFLOW_PROJ_DIR:-.}/db:/var/lib/postgresql/data
```
johnclary: this is an important change! the previous entry for `postgres-db-volume` was empty, and as far as i can tell the current deployment on `atd-data03` does not have a persisted postgres database.

docker-compose.yaml
```diff
@@ -44,26 +42,24 @@
...
     AIRFLOW__DATABASE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres/airflow
     # For backward compatibility, with Airflow <2.3
     AIRFLOW__CORE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres/airflow
     AIRFLOW__CELERY__RESULT_BACKEND: db+postgresql://airflow:airflow@postgres/airflow
     AIRFLOW__CELERY__BROKER_URL: redis://:@redis:6379/0
-    AIRFLOW__CORE__FERNET_KEY: ''
-    AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION: 'true'
-    AIRFLOW__CORE__LOAD_EXAMPLES: 'false'
-    AIRFLOW__API__AUTH_BACKENDS: 'airflow.api.auth.backend.basic_auth,airflow.api.auth.backend.session'
+    AIRFLOW__CORE__FERNET_KEY: ""
```
johnclary, line 59: [the docs](https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/security/secrets/fernet.html) suggest that airflow generates a key on init, but i'm not quite sure if/how this would work in our dockerized deployment. i _think_ there's a risk that (1) our secrets just aren't being encrypted or (2) the fernet key in the airflow container could fall out of sync with encrypted data in the postgres db.

TLDR: we should just set our own fernet key (one for dev and one for prod).

dags/.airflowignore
```diff
@@ -0,0 +1 @@
+utils
```
johnclary, line 1: before i added this, the scheduler was logging an error about our slack operator not being a viable DAG. i guess we need to make sure to ignore anything in the `dags` directory that isn't a DAG.

docker-compose.yaml
```diff
@@ -44,26 +42,24 @@
...
     AIRFLOW__DATABASE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres/airflow
     # For backward compatibility, with Airflow <2.3
     AIRFLOW__CORE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres/airflow
     AIRFLOW__CELERY__RESULT_BACKEND: db+postgresql://airflow:airflow@postgres/airflow
     AIRFLOW__CELERY__BROKER_URL: redis://:@redis:6379/0
-    AIRFLOW__CORE__FERNET_KEY: ''
-    AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION: 'true'
-    AIRFLOW__CORE__LOAD_EXAMPLES: 'false'
-    AIRFLOW__API__AUTH_BACKENDS: 'airflow.api.auth.backend.basic_auth,airflow.api.auth.backend.session'
+    AIRFLOW__CORE__FERNET_KEY: ${_AIRFLOW__CORE__FERNET_KEY}
```
johnclary, line 12: we definitely need to persist our fernet key through an env var. if we don't do this, airflow generates one when it starts up, and if the docker container were stopped we would be unable to decrypt airflow variables. 

and hopefully we're not really using variables thanks to the 1pass integration!

airflow.cfg
```diff
@@ -828,7 +828,7 @@ worker_concurrency = 16
 # If autoscale option is available, worker_concurrency will be ignored.
 # http://docs.celeryproject.org/en/latest/reference/celery.bin.worker.html#cmdoption-celery-worker-autoscale
 # Example: worker_autoscale = 16,12
-# worker_autoscale =
+worker_autoscale = 4,1
```
johnclary, line 831: i do not have an informed opinion on number of concurrent workers, but 16 seemed like a lot. my inclination was to start small and see how things performed.

regardless of how many we want to roll with, this autoscale setting seems better so that we don't have idle workers?

airflow.cfg
```diff
@@ -935,7 +935,7 @@ worker_precheck = False
 # See:
 # http://docs.celeryproject.org/en/master/userguide/configuration.html#std:setting-broker_transport_options
 # Example: visibility_timeout = 21600
-# visibility_timeout =
+visibility_timeout =  3600
```
johnclary, line 938: > The visibility timeout defines the number of seconds to wait for the worker
> to acknowledge the task before the message is redelivered to another worker.
> Make sure to increase the visibility timeout to match the time of the longest
> ETA you're planning to use

6 hours seemed very long, but i don't understand the ramifications of this 🤷

airflow.cfg
```diff
@@ -972,7 +972,7 @@ scheduler_idle_sleep_time = 1
 # Number of seconds after which a DAG file is parsed. The DAG file is parsed every
 # ``min_file_process_interval`` number of seconds. Updates to DAGs are reflected after
 # this interval. Keeping this number low will increase CPU usage.
-min_file_process_interval = 30
+min_file_process_interval = 300
```
johnclary, line 975: I noticed that my log files were filling very quickly, and it was because Airflow re-parses every DAG file at this interval (regardless of if it's switched on/scheduled) and logs the result.

It seems like a lot of overhead considering our DAG code very rarely changes, but the obvious down side is that if your DAG's code changes, you have to wait. Frustratingly, there does not seem to be a way to manually trigger the DAGs to re-parse.

I'll keep looking into ways around this, but one option would be to leave this interval small here but use an override to make the interval larger in prod.

airflow.cfg
```diff
@@ -1019,7 +1019,7 @@ zombie_detection_interval = 10.0
 # will not do scheduler catchup if this is ``False``,
 # however it can be set on a per DAG basis in the
 # DAG definition (catchup)
-catchup_by_default = True
+catchup_by_default = False
```
johnclary, line 1022: I was happy to find this setting! I don't think we've ever wanted a catchup to occur, so seems like a good idea to nix that default.

docker-compose.yaml
```diff
@@ -72,44 +20,34 @@ x-airflow-common:
...
-    - ${AIRFLOW_PROJ_DIR:-.}/dags:/opt/airflow/dags
-    - ${AIRFLOW_PROJ_DIR:-.}/logs:/opt/airflow/logs
-    - ${AIRFLOW_PROJ_DIR:-.}/plugins:/opt/airflow/plugins
-    - ${AIRFLOW_PROJ_DIR:-.}/config:/opt/airflow/config
-    - ${AIRFLOW_PROJ_DIR:-.}/weather:/opt/airflow/weather
+    - ${AIRFLOW_PROJ_DIR}/dags:/opt/airflow/dags
+    - ${AIRFLOW_PROJ_DIR}/logs:/opt/airflow/logs
+    - ${AIRFLOW_PROJ_DIR}/plugins:/opt/airflow/plugins
+    - ${AIRFLOW_PROJ_DIR}/config:/opt/airflow/config
+    - ${AIRFLOW_PROJ_DIR}/airflow.cfg:/opt/airflow/airflow.cfg
```
johnclary, line 29: the important change here is getting the `airflow.cfg` mounted into the container

docker-compose.yaml
```diff
@@ -1,69 +1,17 @@
...
-    AIRFLOW__CORE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres/airflow
-    AIRFLOW__CELERY__RESULT_BACKEND: db+postgresql://airflow:airflow@postgres/airflow
-    AIRFLOW__CELERY__BROKER_URL: redis://:@redis:6379/0
-    AIRFLOW__CORE__FERNET_KEY: ''
-    AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION: 'true'
-    AIRFLOW__CORE__LOAD_EXAMPLES: 'false'
-    AIRFLOW__API__AUTH_BACKENDS: 'airflow.api.auth.backend.basic_auth,airflow.api.auth.backend.session'
+  environment: &airflow-common-env
+    AIRFLOW__CORE__FERNET_KEY: ${_AIRFLOW__CORE__FERNET_KEY}
+    AIRFLOW__CELERY__FLOWER_BASIC_AUTH: ${_AIRFLOW_WWW_USER_USERNAME}:${_AIRFLOW_WWW_USER_PASSWORD}
```
johnclary: this setting breaks the flower server completely :/ 

i have not been able to successfully get auth working for flower and was hoping y'all might be able to have better luck than me. The `FLOWER_AUTH_USERNAME` and `FLOWER_PASSWORD` entries below are used by haproxy: that all seemed to be working fine, except that it did not seem like the flower server was actually enforcing auth. 

It'd be nice to pair on this b/c i'm stuck / confused.

toolbox/certbot/README.md
johnclary, line 1: the certbot files are from https://github.com/cityofaustin/atd-airflow/pull/106

docker-compose.yaml
```diff
@@ -145,22 +86,24 @@ services:
...
-      - postgres-db-volume:/var/lib/postgresql/data
+      - ${AIRFLOW_PROJ_DIR}/db:/var/lib/postgresql/data
     healthcheck:
       test: ["CMD", "pg_isready", "-U", "airflow"]
       interval: 10s
       retries: 5
       start_period: 5s
     restart: always
+    ports:
+      - 5432:5432
```
johnclary: this enables you to connect to the airflow postgres, which was helpful for debugging settings. but in practice this might actually be annoying if we're trying to develop an ETL against a local instance of VZ or Moped, since they also use `5432`?

airflow.cfg
```diff
@@ -216,7 +216,7 @@ daemon_umask = 0o077
 # SqlAlchemy supports many different database engines.
 # More information here:
 # http://airflow.apache.org/docs/apache-airflow/stable/howto/set-up-database.html#database-uri
-sql_alchemy_conn = sqlite:////opt/airflow/airflow.db
+sql_alchemy_conn = postgresql+psycopg2://airflow:airflow@postgres/airflow
```
johnclary, line 219: do we need a secure/obfuscated pg password in production? my thinking was that if an unintended user can reach the postgres instance we have bigger problems.

mddilley: agreed - it takes some effort for us to reach this server so it seems low risk to me.

airflow.cfg
johnclary: also worth noting, there is a separate setting for how often airflow checks for brand new DAGs. that's called 
 `dag_dir_list_interval`  and the default setting is `300` seconds.

johnclary: sorry for all the noise here: i'm reading through [more docs](https://airflow.apache.org/docs/apache-airflow/stable/faq.html#faq) about DAG parsing and scheduler, and i want to test my assumptions some more!

airflow.cfg
mddilley: i read about this at https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#worker-autoscale, and this feels right to me. 4 max and 1 min feels like a good place to start. 👍

airflow.cfg
```diff
@@ -848,7 +848,7 @@ worker_enable_remote_control = true
 
 # The Celery broker URL. Celery supports RabbitMQ, Redis and experimentally
 # a sqlalchemy database. Refer to the Celery documentation for more information.
-broker_url = redis://redis:6379/0
+broker_url = redis://:@redis:6379/0
```
mddilley, line 851: cool, i see this is removed from the docker compose and moved here 👍

airflow.cfg
mddilley: this seems fair to me! I'm not aware of any tasks that would need Airflow to wait this long (assuming this is how that works). I also had trouble assessing exactly what is going but I found [this](https://cloud.google.com/composer/docs/composer-2/troubleshooting-dags#troubleshooting_poison_pill) which gives a little more context since the link in the Airflow docs is 404.

airflow.cfg
mddilley: good to know about these settings. 5 min seems reasonable in prod, and I like the idea of overriding in prod to get faster feedback locally.

docker-compose.yaml
mddilley: thanks for flagging this encryption aspect - i totally overlooked this and didn't understand how this worked. And, yes, I like the plan of leaning fully into 1PW for secrets.

docker-compose.yaml
mddilley: I'll have to take a closer look at this because I'm not sure if I can piece this together enough to pair. I'll poke around and see if I can surface anything.

docker-compose-production.yaml
mddilley, line 1: TIL about Docker compose overrides https://docs.docker.com/compose/extends/#multiple-compose-files. Didn't realize you could pass multiple files, and I like this pattern a lot.

mddilley (commented):
So much to learn here - thanks for diving into the nitty gritty here, John! The links were super helpful too. I threw some opinions here too, and I'm going to hold off on approval for now so I can take some time to look at and understand the flower auth more.

airflow.cfg
johnclary: hopefully the last update from me on this. here's what i found after running some tests:
- `min_file_process_interval`: how frequently airflow re-processes DAG files. this is roughly the time it takes for modified DAG code changes to take effect. (not accounting for serialization: i'm not attempting to understand how that factors in). however, there's one caveat:

> The DAG parser will skip the `min_file_process_interval` check if a file is recently modified.

In my experience testing this, the `min_file_process_interval` was always skipped because changing a DAGs code means...modifying it. I think this behavior is not making much sense b/c it's intended to support use cases where you have thousands of DAGs 🤷 

TLDR: i think we should set the `min_file_process_interval` to at least 5 minutes because it will have no affect when making small + infrequent changes to DAGs, because those changes get picked up immediately.

- `dag_dir_list_interval` - this is the frequency with which Airflow will process brand new DAGs added to the `/dags` folder. we have it set to 5 minutes (the default). unfortunately, there is afaik no way to manually trigger Airflow to load in new DAGs, which means that 99.99% of the time there will be no new DAGs to be processed, and .01% of the time one of us will be waiting a few minutes to see our new DAG show up in the list. 

TLDR: I think we should we should set `dag_dir_list_interval` this to a shorter interval (60 seconds?) in local dev only.

airflow.cfg
johnclary: fwiw, we have been using the plain old `SequentialExecutor` in our [v1 deployment](https://github.com/cityofaustin/atd-airflow/blob/a5853a8974c270d1249c56bd147bb86f14718e51/config/airflow.cfg#L53), which makes me optimistic that even just 4 workers might suffice.

docker-compose-production.yaml
johnclary: @frankhereford taught me that one: it's slick!

dags/.airflowignore
mddilley: Thanks for pointing this out - I'll add this to the readme PR.

docker-compose.yaml
johnclary: thanks Frank for pointing out that the basic auth is handled (in prod only) via haproxy. i have restored this to the previous (not broken) state in this [PR](https://github.com/cityofaustin/atd-airflow/pull/113/files).

airflow.cfg
Charlie-Henry: Parking data ETL tends to run longer than 3600s so I'm not sure if they mean that this timeout should always be greater than our longest ETL? (vs ETA?)

Charlie-Henry (approved):
Thanks for walking us through this yesterday 💯, not that I fully understand what's going on here or tested these changes

airflow.cfg
johnclary: extended docs [here](https://docs.celeryq.dev/en/stable/getting-started/backends-and-brokers/redis.html#id1) but i still don't understand what ETA means in this context 🤷 why are the example settings soooo many hours?

Charlie-Henry: Yeah not sure what's going on 
> This causes problems with ETA/countdown/retry tasks where the time to execute exceeds the visibility timeout; in fact if that happens it will be executed again, and again in a loop.

johnclary: ok, [this article](https://engineering.instawork.com/celery-eta-tasks-demystified-424b836e4e94?gi=3ee459f3b272) was slightly more helpful.

my read from this is that if an ETL task runs longer that `visibility_timeout`, there's a risk of it being re-executed by a concurrent worker. since we never want that to happen, we should set this to a high number (21,600s aka 6 hours if that covers the parking data use case) and also make sure to use the `execution_timeout` [setting](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/tasks.html#timeouts) to make sure our DAGs timeout before we hit the `visibility_timeout`.

this will definitely be a setting to keep an eye on once we launch. and if we are confounded by the CeleryExecutor we can always revert to the simple `SequentialExecutor`.

docker-compose.yaml
```diff
@@ -121,6 +59,8 @@ services:
       - airflow-webserver
       - webhook
     volumes:
+      - ${AIRFLOW_PROJ_DIR}/haproxy/ssl:/usr/local/etc/haproxy/ssl
+      - ${AIRFLOW_PROJ_DIR}/haproxy/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg
       - ${AIRFLOW_PROJ_DIR:-.}/haproxy/ssl:/usr/local/etc/haproxy/ssl
```
frankhereford: The SSL directory appears to be in here twice, which will cause the `docker compose` configuration parser to fail.

docker-compose.yaml
frankhereford: I think it's neat to be able to get in here with DBeaver or something, but I think your concern about how that is going to prevent VZ or Moped's DB instances to fire up due to the port conflict is actually going to be a pretty regular annoyance. Do you think we could pick another port for this if we want to leave it open, or if we're maybe happy with the `psql` client, we could leave it not exposed and use a command like:

```
docker exec -it -e PGUSER=airflow -e PGPASSWORD=airflow -e PGDATABASE=airflow atd-airflow-postgres-1 psql
```

The same command but ending in `pg_dump` will work too for dumping the DB.

docker-compose.yaml
frankhereford: Thank you John! I had gotten used to this getting in via building the container, and I never thought that we'd loose that when we made those changes a while back to make our own supplementary `Dockerfile` instead of having the whole one in there.

docker-compose-production.yaml
```diff
@@ -0,0 +1,7 @@
+services:
+  haproxy:
+    ports:
+      - 80:80
+      - 81:81
+      - 82:82
+      - 83:83
```
frankhereford: I think this may not be what you want - the production haproxy needs to listen on port `80` and `443` of the host machine to respond to incoming HTTP and HTTPS requests. The haproxy port 80 listener will only ever redirect to the same path on 443; it's the SSL enforcer. On port 443, in production, haproxy will use the hostname contained in the `Host` header to pick what backend to proxy the request to.

In development, the haproxy machine needs to listen to as many ports as we have web services. This would look a lot like what you have here, except maybe would be 8080, 8081, 8082, 8083. There in the development haproxy config, haproxy would listen to those four ports and would route to the correct backend service based on what port the request came in, because all requests have the same `Host` header value of `localhost`.

frankhereford (changes requested):
This is fantastic work John! Thank you!

Locally, this won't start up for me because of a double inclusion of a certain volume in the `docker-compose.yml` file, which I think is a one-liner removal.

The other thing that I am wondering about - I think we may need to have two haproxy configuration files, one for production and one for local. The reason being there is that haproxy needs to use port based filtering for which backend to use in local, but can use the `Host` header in production. 

I think I'm struggling a little to convey what I mean here, but if you look at this change set, you can see the differences in that file between local and production.

See `haproxy.cfg` in:

https://github.com/cityofaustin/atd-airflow/compare/airflow-v2...production

I'd love to sync on this if I can help explain if that would be helpful.

docker-compose.yaml
mddilley: Good catch! This came from my certbot branch which was merged into `airflow-v2` and then brought into here. 🙋‍♂️ I made the same update as John.

docker-compose.yaml
```diff
@@ -147,22 +87,24 @@ services:
...
+
   postgres:
     image: postgres:13
     environment:
       POSTGRES_USER: airflow
       POSTGRES_PASSWORD: airflow
       POSTGRES_DB: airflow
     volumes:
-      - postgres-db-volume:/var/lib/postgresql/data
+      - ${AIRFLOW_PROJ_DIR}/db:/var/lib/postgresql/data
```
frankhereford: I'm wondering about this change to keeping the database files around and available to postgres via a bind mount instead of a docker managed volume. @johnclary, maybe we can sync for a minute on this. I'm really curious to learn about this choice.

johnclary: thanks for pointing this out @frankhereford! i completely misunderstood the way docker volumes work. i totally agree we should use a proper docker volume. i will revert this in #113.

will also update the readme to add a note about running `$ docker volume rm atd-airflow_postgres-db-volume` locally if you want to start from a scratch setup in your dev environment

frankhereford: Awesome! Another way to go tabula rasa in a dev environment that leverages the docker compose file is `docker compose down --volumes`. 6 one way, half dozen the other - both are 🚀

docker-compose.yaml
```diff
@@ -1,154 +1,57 @@
...
-    AIRFLOW__CORE__EXECUTOR: CeleryExecutor
-    AIRFLOW__DATABASE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres/airflow
-    # For backward compatibility, with Airflow <2.3
-    AIRFLOW__CORE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres/airflow
-    AIRFLOW__CELERY__RESULT_BACKEND: db+postgresql://airflow:airflow@postgres/airflow
-    AIRFLOW__CELERY__BROKER_URL: redis://:@redis:6379/0
-    AIRFLOW__CORE__FERNET_KEY: ''
-    AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION: 'true'
-    AIRFLOW__CORE__LOAD_EXAMPLES: 'false'
-    AIRFLOW__API__AUTH_BACKENDS: 'airflow.api.auth.backend.basic_auth,airflow.api.auth.backend.session'
```
frankhereford, line 66: I think these lines need to make it back in, at least to allow for a fresh install where it builds up the DB from scratch.

frankhereford (changes requested):
I had to add these back in to get it running locally.

docker-compose.yaml
```diff
@@ -229,6 +229,7 @@ services:
     user: "0:0"
     volumes:
       - ${AIRFLOW_PROJ_DIR}:/sources
+      - ${AIRFLOW_PROJ_DIR}/airflow.cfg:/opt/airflow/airflow.cfg
```
johnclary, line 230: @mddilley @frankhereford here was the problem. even though this service uses the `airflow-common-env`, the `volumes` setting gets overwritten here. so the init script was attempting to fire with zero configuration :/

docker-compose.yaml
johnclary: that's a great point that you can always attach to the container and `psql` into it. i have removed this port binding from the compose b/c we agree it will be more annoying than it's worth.

mddilley (approved):
Spun it up locally, works great, and looks great! 🚀

frankhereford (approved):

PR #113, opened 2023-06-23
Removes webhook and weather DAG and adds Docker CI

johnclary:
This PR:
- Removes the weather DAG
- Removes the webhook stack
- Adds docker build/push CI to our custom airflow image
- Updates the docker compose to use the airflow image from dockerhub

## Steps to test
1. `docker compose up` should get you going. Our airflow image will pull from the hub on your first go.
1. Visit `localhost:8080` to access the airflow UI. Observe that all DAGs have imported without error
2. Visit `localhost:8081` to view the Flower UI. You should see one worker available.


.github/workflows/docker-image-build-push.yml
```diff
@@ -0,0 +1,31 @@
...
+        uses: docker/setup-qemu-action@v2
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      -
+        name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_TOKEN }}
```
johnclary, line 23: ~afaik these env vars do not exist yet in this repo. i don't have admin rights to set them up here~

✅ I remove the 1pass credentials from the repo secrets and added these

Charlie-Henry (approved):
Matches what we discussed yesterday

mddilley (approved):
I spun this up locally and was able to reach the Airflow and Flower web servers. Thanks for taking this on, John! I added a task to https://github.com/cityofaustin/atd-airflow/pull/109 so we can make sure the readme reflects these changes. We list ports for these services, etc.

Time for this code to move on but it was a job well done in proving this entire stack and leading to its adoption! 🚀 ☀️ 😎

.github/workflows/docker-image-build-push.yml
frankhereford: I think this is great, and it totally works perfectly. 

What would you think about putting the 1PW API credentials back in as secrets, and then adding some steps to this workflow to extract them from the 1PW store. You can see an example of accessing 1PW creds in workflows [here](https://github.com/frankhereford/airflow/blob/main/.github/workflows/production_deployment.yml).

The reason I'm suggesting this is to see if we can keep all of our programmatic secrets in 1PW unless there just isn't a way to do it at all. My idea is the more we lean into keeping them in a unified place, the less we have to remember where secrets are stored and where we might have to rotate them.

Dockerfile
```diff
@@ -5,6 +5,5 @@ RUN apt-get update
 RUN apt-get install -y aptitude magic-wormhole vim black
 
 USER ${AIRFLOW_UID}
-COPY airflow.cfg /opt/airflow/airflow.cfg
```
frankhereford, line 8: I saw where this got put in via a bind mount in the other PR. This is really fine; it's very much in the docker ideal of making a generic tool that can be configured at run time. 💯 🚀

docker-compose.yaml
```diff
@@ -7,12 +7,9 @@ x-airflow-common: &airflow-common
   # In order to add custom dependencies or upgrade provider packages you can use your extended image.
   # Comment the image line, place your Dockerfile in the directory where you placed the docker-compose.yaml
   # and uncomment the "build" line below, Then run `docker-compose build` to build the images.
-  # image: ${AIRFLOW_IMAGE_NAME:-apache/airflow:2.6.1}
-  build: .
+  image: atddocker/atd-airflow:production
```
frankhereford, line 10: This right here -- shifting away from the image-d'jour that happens to be on the machine at the time is huge. This is a really big improvement in terms of many things from disaster recovery to setting up testing environments - you name it.

Thank you! It's also a good example of the open-by-default ideal to publish the image. ❤️

docker-compose.yaml
```diff
@@ -281,7 +245,7 @@ services:
 
   flower:
     <<: *airflow-common
-    command: ["celery", "flower"]
+    command: celery flower
```
frankhereford, line 248: This would be a welcome PR back into `apache/airflow` I bet!

frankhereford (commented):
> ☔ I'm not sure where I'll be able to check the weather forecast now ...

In all seriousness, this is really great work. I'm reminded to remind myself that our public repos are **public**, and the less "DTS" or "ATD" or "TPW" specific stuff we bake into them, the better. This work exemplifies that ideal.

😢 I'm not currently able to get this up and running locally at the moment, and I'm working on figuring out why. I'm sure it's something on my end, and I'll flip this to approval as soon as I get it identified. I did want to publish these comments though since we're going to be talking about them tomorrow. 

Thanks again for all this John, it's 💯

Dockerfile
johnclary: 👍 my thinking was that it's more transparent if our base image has the default config, but i don't feel strongly 🤷

docker-compose.yaml
johnclary: this change is actually reverting to airflow's original command: i made the change while trying to get flower to work 🙈

.github/workflows/docker-image-build-push.yml
johnclary: i'm onboard! will do 🚀

johnclary:
i'm going to merge this up and put finishing touches in #111

PR #114, opened 2023-06-29
Track cron entry for certbot

mddilley:
This PR adds a file with the cron entry for the certbot script so we can track it. I think this lines up with [what we talked about](https://github.com/cityofaustin/atd-data-tech/issues/12453#issuecomment-1607478527) 🙏 but happy to make any adjustments.

Post prod merge and stack update list:
- [x] Symlink `sudo ln -s /usr/airflow/atd-airflow/toolbox/certbot/cron_entry /etc/cron.d/renew_airflow_ssl_certificates`
- [x] Set up logrotate
    - create `/etc/logrotate.d/certbot`
    - add config shown below
```
/var/log/airflow_cert_renewal.log {
    weekly
    rotate 4 (not sure how many weeks back we should keep around)
    compress
    missingok
    delaycompress (do we want to keep the most recently rotated log uncompressed?)
    create 0600 root root
}
```


toolbox/certbot/cron_entry
```diff
@@ -0,0 +1 @@
+0 */12 * * * root /usr/airflow/atd-airflow/toolbox/certbot/renew.sh >> /var/log/airflow_cert_renewal.log
```
frankhereford: Do you mind sticking a EOF line feed on this?

frankhereford: Normally, I'd think maybe we should setup a `logrotate.d` file also to roll this log, but I think the amount of data we'll accumulate in there is going to be plenty small to not worry about it. Do you agree?

frankhereford (approved):
Looks so good!! Thank you!

toolbox/certbot/cron_entry
mddilley: Oh yup, thanks for catching that! 🙏

mddilley: I think so? I don't feel familiar enough with this to make the call, but maybe can mirror what we do with the Airflow logs if that isn't too much.

johnclary: i say we rotate it! set it and forget it!

johnclary (approved):

PR #116, opened 2023-06-30
Fetches 1pass secrets in task

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/12723

This PR mostly just establishes a pattern to access 1Password secrets through [XComs](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/xcoms.html).  It is meant to avoid making many unnecessary API calls to the 1pass connect server, as discussed in [this slack thread](https://austininnovation.slack.com/archives/CHZE6BC6L/p1687530254736569)

I got carried away with README edits! I'm happy to revise/revert: feedback is very welcome.

## Associated repo
- atd-airflow

## Testing

**Steps to test:**
1. Checkout this branch and spin up Airflow as usual:
```
$ docker compose up
```
2. Trigger any of our existing service bot DAGs: they should run normally

---
#### Ship list
- [x] Code reviewed 
- [ ] Product manager approved
- [ ] Add note to 1PW secrets moved to API vault and check for duplicates

PR #118, opened 2023-07-11
Trail Counters to Production

Charlie-Henry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12453

## Associated repo
https://github.com/cityofaustin/atd-trail-counter-data

## Testing

**Steps to test:**
1. git pull this branch down to your local machine
2. `docker compose build` this version of the airflow stack
3.  `docker compose run --rm airflow-cli dags test atd_trail_counters` will run the DAG
4. View the logs and verify it's pulling data from the ecocounter website something like:
```
DEBUG:counter_data.py:0 records found for Shoal Creek Blvd at 38th St PC
DEBUG:counter_data.py:0 records found for Southern Walnut Creek Trail - Urban Trail
...
```

---
#### Ship list
- [ ] Code reviewed 


dags/ped_data_pub.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @task(
+        task_id="get_env_vars",
+        execution_timeout=timedelta(seconds=30),
+    )
+    def get_env_vars():
+        from utils.onepassword import load_dict
+
+        env_vars = load_dict(REQUIRED_SECRETS)
+        return env_vars
```
Charlie-Henry: Copying John's pattern from here:
https://github.com/cityofaustin/atd-airflow/pull/116

dags/ped_data_pub.py
```diff
@@ -0,0 +1,80 @@
...
+
+    env_vars = get_env_vars()
+
+    start = "{{ prev_start_date_success.strftime('%Y-%m-%d') if prev_start_date_success or '1970-01-01'}}"
```
johnclary: this was erroring for me until i changed `or` to `else`:

```
[2023-07-12, 13:59:53 UTC] {abstractoperator.py:597} ERROR - Exception rendering Jinja template for task 'trail_counter_data_publish', field 'command'. Template: "python counter_data.py --start {{ prev_start_date_success.strftime('%Y-%m-%d') if prev_start_date_success or '1970-01-01'}}"
Traceback (most recent call last):
....
jinja2.exceptions.UndefinedError: 'None' has no attribute 'strftime'
```

dags/ped_data_pub.py
```diff
@@ -0,0 +1,80 @@
...
+        "opfield": "socrata.apiKeyId",
+    },
+    "COUNTERS_DATASET": {
+        "opitem": "Socrata Dataset IDs",
+        "opfield": "datasets.Trail Counters",
+    },
+}
+
+with DAG(
+    dag_id="atd_trail_counters",
```
johnclary: it'd be nice to have this DAG's file name be more similar to the DAG name. e.g. `atd_trail_counters.py`

johnclary (dismissed):
@Charlie-Henry this is looking great. just a few minor changes, and I'm going to start a Slack thread wrt to Socrata entries in 1password.

When I tested the code, it did process a bunch of data. Is that ok? On that note, would you mind adding a bit more detail to this repo's [readme](https://github.com/cityofaustin/atd-trail-counter-data)? It wasn't quickly obvious from the docs what this script was going to do, if we need to worry about duplicating data, and where the public dataset can be found.

dags/ped_data_pub.py
```diff
@@ -0,0 +1,80 @@
+# test locally with: docker compose run --rm airflow-cli dags test atd_trail_counters
+
+import os
```
frankhereford: It looks like this isn't used, so I think it can go.

dags/ped_data_pub.py
```diff
@@ -0,0 +1,80 @@
...
+from datetime import datetime, timedelta
+
+from airflow.decorators import task
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+
+from utils.slack_operator import task_fail_slack_alert
+from utils.onepassword import load_dict
+
+from onepasswordconnectsdk.client import Client, new_client
```
frankhereford: Same with these, I think these are unused.

dags/ped_data_pub.py
```diff
@@ -0,0 +1,80 @@
...
+
+import os
+from datetime import datetime, timedelta
+
+from airflow.decorators import task
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+
+from utils.slack_operator import task_fail_slack_alert
+from utils.onepassword import load_dict
```
frankhereford: It looks like you're importing this in a function's namespace below, so I think this can be removed.

frankhereford (approved):
I agree with John's suggestions, esp with the bugfix and the filename. I was able to get this running, and I don't have anything valuable to add beyond a couple of imports that can get removed. This is great Charlie, and thanks for cleaning up this branch so we didn't loose your work!!

🚢 🚢 🚢

Charlie-Henry:
@johnclary 
> it did process a bunch of data. Is that ok?

Yes, likely used yesterday's date in the `--start` param and pulled in that data. Each row is protected by an ID so no duplication issues should arise. Run this script to your heart's content!

Charlie-Henry:
Since the file was renamed so lots of deltas here's what I addressed with my commit:
- Dropped unnecessary imports thx @frankhereford 
- `else` instead of `or` in the start param
-  Made a trail counters 1pass entry where the only item is the Socrata resource ID and is now pointing to that instead

[Dataset is here](https://datahub.austintexas.gov/Transportation-and-Mobility/Trail-Counters-Daily-Totals/26tt-cp67) internal because [blocked](https://github.com/cityofaustin/atd-data-tech/issues/8642)

PR #120, opened 2023-07-12
Build Airflow containers with ARM64 support

frankhereford:
## Associated issues

none

## Testing

Mac Users:

Really, if you can spin up your local setup with the `atddocker/atd-airflow:testing` then you should be good to go.

AMD64 (windows & linux folks) - it was never broken!

**Steps to test:**

Modify your docker compose file around line 10 to use the `testing` tag. Don't forget to set it back because I'll remove this tag when I merge this.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/.github/workflows/docker-image-build-push.yml
+++ b/.github/workflows/docker-image-build-push.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - production
-      - multi-arch-building
 
 jobs:
   docker:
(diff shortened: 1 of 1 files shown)


chiaberry (approved):
thank you!

mddilley (approved):
I started the stack with the `atddocker/atd-airflow:testing` image, and everything came up. Thanks for patching this, and glad that Chia surfaced it. 🚀

PR #121, opened 2023-07-12
Port atd_metrobike_trips to airflow V2

Charlie-Henry:
Porting this airflow v1 [metrobike trip publishing ETL](https://github.com/cityofaustin/atd-airflow/blob/master/dags/atd_metrobike_trips.py) to v2.

## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/9388
We don't have a epic for porting v1 to v2 for airflow, yet.

## Associated repo
[atd-metrobike](https://github.com/cityofaustin/atd-metrobike
)

## Testing

**Steps to test:**
1. `git pull` this branch down
2. `docker compose run --rm airflow-cli dags test atd_metrobike_trips`
3. 99% of the time this will end up being a non-op as the ETL will `return` if it detects that the data in Socrata is already up to date with the data stored in the dropbox folder. This is configured to run weekly on Monday mornings.
4. check for `atd_metrobike_trips_socrata ran successfully!` in the logs

---
#### Ship list
- [ ] Code reviewed 
- [ ] Turn off Airflow V1 DAG

### DIFF
--- a/dags/atd_metrobike_trips.py
+++ b/dags/atd_metrobike_trips.py
@@ -1,16 +1,20 @@
 # Test locally with: docker compose run --rm airflow-cli dags test atd_metrobike_trips
 
+import os
+
 from datetime import datetime, timedelta
 
 from airflow.decorators import task
 from airflow.models import DAG
 from airflow.operators.docker_operator import DockerOperator
 
+from utils.onepassword import get_env_vars_task
 from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
 
 default_args = {
     "owner": "airflow",
-    "description": "Fetch metrobike trip data from dropbox and publish to Socrata",
     "depend_on_past": False,
     "start_date": datetime(2020, 12, 31),
     "email_on_failure": False,
@@ -19,8 +23,6 @@ default_args = {
     "retry_delay": timedelta(minutes=5),
     "on_failure_callback": task_fail_slack_alert,
 }
-
-docker_image = "atddocker/atd-metrobike:production"
 
 # assemble env vars
 
@@ -43,33 +45,23 @@ REQUIRED_SECRETS = {
     },
 }
 
+# runs weekly in prod environment at 1:33am Monday
 with DAG(
     dag_id="atd_metrobike_trips",
+    description="Fetch metrobike trip data from dropbox and publish to Socrata",
     default_args=default_args,
-    schedule_interval="33 1 * * 1",  # runs weekly at 1:33am Monday
+    schedule_interval="33 1 * * 1" if DEPLOYMENT_ENVIRONMENT == "production" else None,
     dagrun_timeout=timedelta(minutes=60),
-    tags=["repo:atd-metrobike", "metrobike"],
+    tags=["repo:atd-metrobike", "metrobike", "socrata"],
     catchup=False,
 ) as dag:
-
-    @task(
-        task_id="get_env_vars",
-        execution_timeout=timedelta(seconds=30),
-    )
-    def get_env_vars():
-        from utils.onepassword import load_dict
-
-        return load_dict(REQUIRED_SECRETS)
-
-    env_vars = get_env_vars()
+    env_vars = get_env_vars_task(REQUIRED_SECRETS)
 
     t1 = DockerOperator(
         task_id="atd_metrobike_trips_socrata",
-        image=docker_image,
-        api_version="auto",
+        image="atddocker/atd-metrobike:production",
         auto_remove=True,
         command="python publish_trips.py",
-        network_mode="bridge",
         environment=env_vars,
         tty=True,
         force_pull=True,
--- a/dags/atd_metrobike_trips.py
+++ b/dags/atd_metrobike_trips.py
@@ -1,0 +1,78 @@
+# Test locally with: docker compose run --rm airflow-cli dags test atd_metrobike_trips
+
+from datetime import datetime, timedelta
+
+from airflow.decorators import task
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+
+from utils.slack_operator import task_fail_slack_alert
+
+default_args = {
+    "owner": "airflow",
+    "description": "Fetch metrobike trip data from dropbox and publish to Socrata",
+    "depend_on_past": False,
+    "start_date": datetime(2020, 12, 31),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=5),
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+docker_image = "atddocker/atd-metrobike:production"
+
+# assemble env vars
+
+REQUIRED_SECRETS = {
+    "SOCRATA_API_KEY_SECRET": {
+        "opitem": "Socrata Key ID, Secret, and Token",
+        "opfield": "socrata.apiKeySecret",
+    },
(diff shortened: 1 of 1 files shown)


dags/atd_metrobike_trips.py
```diff
@@ -0,0 +1,78 @@
...
+    "depend_on_past": False,
+    "start_date": datetime(2020, 12, 31),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=5),
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+docker_image = "atddocker/atd-metrobike:production"
```
frankhereford: This variable is defined here in the global scope but only used once in the DockerOperator. Maybe get rid of this variable and just put the value in the DockerOperator, for clarity?

dags/atd_metrobike_trips.py
```diff
@@ -0,0 +1,78 @@
...
+
+from utils.slack_operator import task_fail_slack_alert
+
+default_args = {
+    "owner": "airflow",
+    "description": "Fetch metrobike trip data from dropbox and publish to Socrata",
+    "depend_on_past": False,
+    "start_date": datetime(2020, 12, 31),
+    "email_on_failure": False,
+    "email_on_retry": False,
```
frankhereford, line 21: I don't think we're going to run with email as a mechanism to warn us when things go south, right? I think we're going to lean in to the slack notifier. If you agree, I think these can go.

dags/atd_metrobike_trips.py
```diff
@@ -0,0 +1,78 @@
+# Test locally with: docker compose run --rm airflow-cli dags test atd_metrobike_trips
```
frankhereford, line 1: I like these testing related comments you're adding Charlie, thank you!

dags/atd_metrobike_trips.py
```diff
@@ -0,0 +1,78 @@
...
+
+    env_vars = get_env_vars()
+
+    t1 = DockerOperator(
+        task_id="atd_metrobike_trips_socrata",
+        image=docker_image,
+        api_version="auto",
+        auto_remove=True,
+        command="python publish_trips.py",
+        network_mode="bridge",
```
frankhereford: I think the only time we're going to need `network_mode="bridge"` is when we're writing a DAG that needs to interoperate with the signals network, which is not available on `atd-data03`. This is putting the DAG's running container on the COA network instead of behind the virtual NAT/firewall of the docker service. Without a reason to do so, I think this is best left out.

dags/atd_metrobike_trips.py
```diff
@@ -0,0 +1,78 @@
...
+        from utils.onepassword import load_dict
+
+        return load_dict(REQUIRED_SECRETS)
+
+    env_vars = get_env_vars()
+
+    t1 = DockerOperator(
+        task_id="atd_metrobike_trips_socrata",
+        image=docker_image,
+        api_version="auto",
```
frankhereford: We're interfacing with docker though its socket and no longer over a HTTP. I think this is OK to drop.

dags/atd_metrobike_trips.py
frankhereford: For ARM users (macs), this image does not pull and run because it doesn't have a manifest for ARM64. Are you building this by hand or with CI? Can you build this with ARM64 support, or adjust the CI to do it? LMK if I can help - i have some recipes for this I can share with you. I can't test this locally (on my mac, at least) yet.

frankhereford (dismissed):
This looks great Charlie - thank you for slaying another one!!

I'm requesting changes only on my two comments below:
* 99%: Can you remove the `network_mode` and `api_version` DockerOperator arguments?
* ~1%: Multi-architecture build for testing on a mac~

I think we might want to have a discussion as a team about if we are going to ask each other to make images that we push to docker hub that are related to airflow have ARM64 support to run natively? I am sort of surprised that the mac won't pull the AMD64 image and execute it through the emulation layer. I feel there must be something subtle going on that causes it to do this:

```
$ docker pull atddocker/atd-metrobike:production
production: Pulling from atddocker/atd-metrobike
no matching manifest for linux/arm64/v8 in the manifest list entries
```

**Also, how can we get Charlie a modern Mac?** I know it has to be frustrating to be asked to build something for an architecture that he doesn't have handy. I'm sorry for that Charlie, and thank you!!

dags/atd_metrobike_trips.py
Charlie-Henry: This was built by the CI in [atd-metrobike](https://github.com/cityofaustin/atd-metrobike). So that's the dockerfile that needs some TLC then we should be able to get you to run this locally. Frustrating but I guess that's the tech debt we have on some of these older docker images.

dags/atd_metrobike_trips.py
frankhereford: I made the same mistaken observation on Chia's PR! 🤦

This is right on and I'm going to start including these configuration directives.

dags/atd_metrobike_trips.py
frankhereford: Totally! Thanks for the info on where this is built. I'm going to make an issue or maybe a micro PR to Mac this image ma friendly. Thanks!!

johnclary: I'm on it! https://github.com/cityofaustin/atd-metrobike/pull/6

dags/atd_metrobike_trips.py
```diff
@@ -0,0 +1,78 @@
...
+        "opfield": "production.Dropbox Token",
+    },
+}
+
+with DAG(
+    dag_id="atd_metrobike_trips",
+    default_args=default_args,
+    schedule_interval="33 1 * * 1",  # runs weekly at 1:33am Monday
+    dagrun_timeout=timedelta(minutes=60),
+    tags=["repo:atd-metrobike", "metrobike"],
```
johnclary: i've been using a `socrata` tag as well

dags/atd_metrobike_trips.py
```diff
@@ -0,0 +1,78 @@
...
+) as dag:
+
+    @task(
+        task_id="get_env_vars",
+        execution_timeout=timedelta(seconds=30),
+    )
+    def get_env_vars():
+        from utils.onepassword import load_dict
+
+        return load_dict(REQUIRED_SECRETS)
```
johnclary: this looks good, but if you merge production into this branch you can import this task:

```python
from utils.onepassword import get_env_vars_task
```

dags/atd_metrobike_trips.py
```diff
@@ -0,0 +1,78 @@
...
+
+from airflow.decorators import task
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+
+from utils.slack_operator import task_fail_slack_alert
+
+default_args = {
+    "owner": "airflow",
+    "description": "Fetch metrobike trip data from dropbox and publish to Socrata",
```
johnclary: i don't think this `description` arg does anything, because these are default args for tasks. 

if you pass this into the `DAG` constructor you will see the description in the Airflow UI:
Screenshot 2023-07-17 at 8 48 23 AM

i had this wrong in every single DAG in airflow v1 🙈

johnclary (approved):
Other than Frank's feedback and moving that description arg this is  🚢 🚢 🚢 🚢

johnclary:
@Charlie-Henry here's the airflow migration mega issue: https://github.com/cityofaustin/atd-data-tech/issues/12860

dags/atd_metrobike_trips.py
frankhereford: Worked great - thank you!

PR #122, opened 2023-07-12
Creates DAG to publish signal studies

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/7168

## Associated repo
- https://github.com/cityofaustin/atd-knack-services

## Testing

1. Start your local rig
2. Trigger the `atd_knack_services_signal_studies` DAG.
3. You should see all three tasks succeed with ~420 records processed.
4. [This](https://data.austintexas.gov/Transportation-and-Mobility/Traffic-and-Pedestrian-Signal-Evaluations-BETA-/h4cy-hpgs) is the dataset you just updated

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_knack_signal_studies.py
+++ b/dags/atd_knack_signal_studies.py
@@ -1,9 +1,9 @@
 import os
-from datetime import datetime, timedelta
 
 from airflow.decorators import task
 from airflow.models import DAG
 from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration
 
 from utils.slack_operator import task_fail_slack_alert
 
@@ -13,7 +13,7 @@ default_args = {
     "owner": "airflow",
     "description": "Publishes all signal stuides records from Knack to Socrata",
     "depends_on_past": False,
-    "start_date": datetime(2015, 12, 1),
+    "start_date": datetime(2015, 1, 1, tz="America/Chicago"),
     "email_on_failure": False,
     "email_on_retry": False,
     "retries": 0,
@@ -54,9 +54,9 @@ REQUIRED_SECRETS = {
 with DAG(
     dag_id=f"atd_knack_services_signal_studies",
     default_args=default_args,
-    schedule_interval=None,
-    dagrun_timeout=timedelta(minutes=5),
-    tags=["repo:atd-knack-services", "knack", "github", "socrata"],
+    schedule_interval="5 4 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    dagrun_timeout=duration(minutes=5),
+    tags=["repo:atd-knack-services", "knack", "socrata"],
     catchup=False,
 ) as dag:
     docker_image = "atddocker/atd-knack-services:production"
@@ -65,7 +65,7 @@ with DAG(
 
     @task(
         task_id="get_env_vars",
-        execution_timeout=timedelta(seconds=30),
+        execution_timeout=duration(seconds=30),
     )
     def get_env_vars():
         from utils.onepassword import load_dict
--- a/dags/atd_knack_signal_studies.py
+++ b/dags/atd_knack_signal_studies.py
@@ -1,0 +1,100 @@
+import os
+from datetime import datetime, timedelta
+
+from airflow.decorators import task
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+default_args = {
+    "owner": "airflow",
+    "description": "Publishes all signal stuides records from Knack to Socrata",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 12, 1),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+REQUIRED_SECRETS = {
+    "KNACK_APP_ID": {
+        "opitem": "Knack AMD Data Tracker",
+        "opfield": f"production.appId",
+    },
+    "KNACK_API_KEY": {
+        "opitem": "Knack AMD Data Tracker",
+        "opfield": f"production.apiKey",
+    },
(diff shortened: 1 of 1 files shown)


dags/atd_knack_signal_studies.py
```diff
@@ -0,0 +1,100 @@
...
+    t1 = DockerOperator(
+        task_id="atd_knack_signal_studies_to_postgrest",
+        image=docker_image,
+        api_version="auto",
+        auto_remove=True,
+        command=f"./atd-knack-services/services/records_to_postgrest.py -a {app_name} -c {container}",
+        environment=env_vars,
+        tty=True,
+        force_pull=True,
+        mount_tmp_dir=False,
```
johnclary, line 86: I was getting a permissions error until I added this arg. I'm baffled, because I can run other DAGs that use `DockerOperator` without this arg.

Would someone test this on their end?

dags/atd_knack_signal_studies.py
```diff
@@ -0,0 +1,100 @@
...
+    "PGREST_JWT": {
+        "opitem": "atd-knack-services PostgREST",
+        "opfield": "production.jwt",
+    },
+}
+
+with DAG(
+    dag_id=f"atd_knack_services_signal_studies",
+    default_args=default_args,
+    schedule_interval=None,
```
johnclary: derp—i forgot to schedule this :/ i will take a look at our other Knack DAGs to try to find an optimal time. this needs to run once daily.

johnclary: not sure if this is a great idea, but fwiw we _could_ create a Data Tracker worker pool if we wanted to enforce some concurrency limit on ETLs. https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/pools.html

dags/atd_knack_signal_studies.py
chiaberry: [2023-07-12, 14:34:32 CDT] {docker.py:333} WARNING - Using remote engine or docker-in-docker and mounting temporary volume from host is not supported. Falling back to `mount_tmp_dir=False` mode. You can set `mount_tmp_dir` parameter to False to disable mounting and remove the warning

I get this error on my branch when I test run my ocr dag...I'll try testing yours now

dags/atd_knack_signal_studies.py
```diff
@@ -0,0 +1,100 @@
...
+    dagrun_timeout=timedelta(minutes=5),
+    tags=["repo:atd-knack-services", "knack", "github", "socrata"],
```
Charlie-Henry: why `github`?

Charlie-Henry (dismissed):
Tested it and I saw the date be updated in the dataset ✅ 

There's "[BETA]" in the dataset title, we should probably drop that. Some updates to the metadata is needed too like adding a description, column names, and column descriptions. 

I don't remember how exactly this works but I think our Postgres DB adds a `updated_at` column automatically? Would be good to add that to the Socrata dataset so we can verify that data is being updated.

dags/atd_knack_signal_studies.py
johnclary: same! [the docker operator docs ](https://airflow.apache.org/docs/apache-airflow-providers-docker/stable/_modules/airflow/providers/docker/operators/docker.html) explain what's going on, but i don't understand why this error happens for some DAGs but not others.

dags/atd_knack_signal_studies.py
johnclary: no reason 🙏

johnclary:
thanks @Charlie-Henry! i will clean up the ODP dataset 👍 

> I don't remember how exactly this works but I think our Postgres DB adds a updated_at column automatically? Would be good to add that to the Socrata dataset so we can verify that data is being updated.

i really like this idea! it would be a lil bit of a knack services refactor to get the postgres `updated_at` into the socrata payloads. we'd need to come up with a column name that doesn't collide with a column in the knack data, something like `_etl_execution_date`. but that'd be a cool feature to maybe optionally switch on this column.

dags/atd_knack_signal_studies.py
```diff
@@ -0,0 +1,100 @@
...
+    "PGREST_JWT": {
+        "opitem": "atd-knack-services PostgREST",
+        "opfield": "production.jwt",
+    },
+}
+
+with DAG(
+    dag_id=f"atd_knack_services_signal_studies",
+    default_args=default_args,
+    schedule_interval="5 4 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
johnclary, line 57: i added this hack based on [this discussion](https://austininnovation.slack.com/archives/C013G4RNJ01/p1689231790907259). it ensures the DAG will not be run on a schedule locally 😅

dags/atd_knack_signal_studies.py
```diff
@@ -0,0 +1,100 @@
...
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration
+
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+default_args = {
+    "owner": "airflow",
+    "description": "Publishes all signal stuides records from Knack to Socrata",
```
chiaberry: signal studies

dags/atd_knack_signal_studies.py
mddilley: when I tested each task without `mount_tmp_dir=False`, I was able to run them successfully but I did get the same warning that Chia did above. 🤔 Maybe we've been missing this in the logs of other DAGs since it wasn't failing them.

mddilley (approved):
I tested this locally, and I see each task process 417 records. 🚀

dags/atd_knack_signal_studies.py
johnclary: oh interesting you're both only seeing a  `WARNING`. i thought this was causing my DAGs to fail but maybe i misread the stack trace. will keep an eye on it. 🙏

johnclary:
thanks for the reviews. i'll work with Renee to get the dataset description + metadata updated.

PR #123, opened 2023-07-12
Add dag for cr3 diagram extract

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/11731

This is the prefect 1 version of the ETL: https://github.com/cityofaustin/atd-prefect/blob/main/flows/vision-zero/cr3_ocr_narrative_extract_diagram/cr3_ocr_narrative_extract_diagram.py

## Associated repo

The script used to live in this airflow repo, you can see it still in the master branch. I have moved it and made it into a Docker image, copying the pattern Frank started with the other VZ ETLs

https://github.com/cityofaustin/atd-vz-data/pull/1250

## Testing

**Steps to test:**

Running locally, trigger the dag: vz_cr3_ocr_narrative_extract_development

The original dag was scheduled `schedule_interval="*/10 18-23,2-5 * * *"`, but the prefect version looks to be scheduled every 5 minutes, between 8a-11a. So I went with the prefect schedule for this dag. 

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ _AIRFLOW_WWW_USER_USERNAME=admin
 _AIRFLOW_WWW_USER_PASSWORD=<Pick your initial admin password here>
 AIRFLOW_PROJ_DIR=<The absolute path of your Airflow repository checkout>
 # this fernet key is for testing purposes only
-_AIRFLOW__CORE__FERNET_KEYY=PTkIRwL-c46jgnaohlkkXfVikC-roKa95ipXfqST7JM=
+_AIRFLOW__CORE__FERNET_KEY=PTkIRwL-c46jgnaohlkkXfVikC-roKa95ipXfqST7JM=
 OP_API_TOKEN=<Get from 1Password entry named "Connect Server: Production Access Token: API Accessible Secrets">
 OP_CONNECT=<Get from 1Password entry named "Endpoint for 1Password Connect Server API">
 OP_VAULT_ID=<Get from 1Password entry named "Vault ID of API Accessible Secrets vault">
--- a/dags/utils/onepassword.py
+++ b/dags/utils/onepassword.py
@@ -9,7 +9,7 @@ VAULT_ID = os.getenv("OP_VAULT_ID")
 
 
 def get_client():
-    """Get oneopassword connect client
+    """Get onepassword connect client
 
     Returns:
         Client
--- a/dags/vz_cr3_extract_ocr_narrative.py
+++ b/dags/vz_cr3_extract_ocr_narrative.py
@@ -1,5 +1,5 @@
 import os
-import pendulum
+from pendulum import datetime, duration
 from datetime import datetime, timedelta
 
 from airflow.decorators import task
--- a/dags/vz_cr3_extract_ocr_narrative.py
+++ b/dags/vz_cr3_extract_ocr_narrative.py
@@ -1,6 +1,5 @@
 import os
 from pendulum import datetime, duration
-from datetime import datetime, timedelta
 
 from airflow.decorators import task
 from airflow.models import DAG
@@ -47,13 +46,13 @@ with DAG(
     default_args=default_args,
     # Every 5 minutes, at 8A, 9A, and 10A
     schedule_interval="*/5 8-10 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
-    dagrun_timeout=timedelta(minutes=5),
+    dagrun_timeout=pendulum.duration(minutes=5),
     tags=["repo:atd-vz-data", "vision-zero"],
     catchup=False,
 ) as dag:
     @task(
         task_id="get_env_vars",
-        execution_timeout=timedelta(seconds=30),
+        execution_timeout=pendulum.duration(seconds=30),
     )
     def get_env_vars():
         from utils.onepassword import load_dict
--- a/dags/vz_cr3_extract_ocr_narrative.py
+++ b/dags/vz_cr3_extract_ocr_narrative.py
@@ -20,8 +20,6 @@ default_args = {
     "on_failure_callback": task_fail_slack_alert,
 }
 
-docker_image = "atddocker/atd-vz-cr3-extract:production"
-
 REQUIRED_SECRETS = {
     "HASURA_ENDPOINT": {
         "opitem": "Vision Zero CRIS Import",
@@ -62,7 +60,7 @@ with DAG(
     
     DockerOperator(
         task_id="ocr_narrative_extract",
-        image=docker_image,
+        image="atddocker/atd-vz-cr3-extract:production",
         api_version="auto",
         auto_remove=True,
         command="./cr3_extract_diagram/cr3_extract_diagram_ocr_narrative.py -v -d --update-narrative --update-timestamp --batch 100 --cr3-source atd-vision-zero-editor production/cris-cr3-files --save-diagram-s3 atd-vision-zero-website cr3_crash_diagrams/production",
--- a/dags/vz_cr3_extract_ocr_narrative.py
+++ b/dags/vz_cr3_extract_ocr_narrative.py
@@ -1,0 +1,71 @@
+import os
+from datetime import datetime, timedelta
+
+from airflow.decorators import task
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+default_args = {
+    "owner": "airflow",
+    "description": "Extracts the diagram and narrative out of CR3 pdfs",
+    "depends_on_past": False,
+    "start_date": datetime(2019, 1, 1),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+docker_image = "atddocker/atd-vz-cr3-extract:production"
+
+REQUIRED_SECRETS = {
+    "HASURA_ENDPOINT": {
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": "production.GraphQL Endpoint",
+    },
+    "HASURA_ADMIN_KEY": {
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": "production.GraphQL Endpoint key",
+    }, 
+    "AWS_ACCESS_KEY_ID": {
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": "production.AWS Access key",
+    },
+    "AWS_SECRET_ACCESS_KEY": {
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": "production.AWS Secret key",
+    }
+}
+
+with DAG(
+    dag_id=f"vz_cr3_ocr_narrative_extract_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    schedule_interval="*/5 8-10 * * *", # Every 5 minutes, at 8A, 9A, and 10A 
+    dagrun_timeout=timedelta(minutes=5),
+    tags=["repo:atd-vz-data", "vision-zero"],
+    catchup=False,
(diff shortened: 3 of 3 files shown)


dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -0,0 +1,72 @@
...
+    "description": "Extracts the diagram and narrative out of CR3 pdfs",
+    "depends_on_past": False,
+    "start_date": pendulum.datetime(2019, 1, 1, tz="America/Chicago"),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+docker_image = "atddocker/atd-vz-cr3-extract:production"
```
mddilley: noting that this came up in Charlie's metrobike PR https://github.com/cityofaustin/atd-airflow/pull/121#discussion_r1261630375 I know that this comes from existing patterns, and I think that I agree that it is nice to see the image name directly in the DockerOperator params? Also just realizing that I did this same thing in the service bot DAGs.

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -0,0 +1,72 @@
...
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": "production.AWS Secret key",
+    }
+}
+
+with DAG(
+    dag_id=f"vz_cr3_ocr_narrative_extract_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    schedule_interval="*/5 8-10 * * *", # Every 5 minutes, at 8A, 9A, and 10A 
+    dagrun_timeout=timedelta(minutes=5),
```
mddilley: Same, I saw this callout in Frank's DAG interval PR so this could align with whatever is decided there https://github.com/cityofaustin/atd-airflow/pull/124#discussion_r1262792486

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -0,0 +1,72 @@
...
+    "AWS_SECRET_ACCESS_KEY": {
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": "production.AWS Secret key",
+    }
+}
+
+with DAG(
+    dag_id=f"vz_cr3_ocr_narrative_extract_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    schedule_interval="*/5 8-10 * * *", # Every 5 minutes, at 8A, 9A, and 10A
```
mddilley: I'm not sure if I know the history of this schedule. Does anyone know? Asking for my own knowledge and not suggesting any change!

mddilley (approved):
I tested this locally, and I see all green boxes in the Airflow dashboard. I added a few notes, but this looks great! 🚢 ✨

dags/vz_cr3_extract_ocr_narrative.py
frankhereford: This is me! I did this schedule because I was trying to limit the number of executions, because it really needs to run like 3 or 4 times per day, but occasionally 10 or so times.  It does the OCRing in chunks of 100 or 200, so I used this schedule to have it run frequently in the morning, but by the time it should be out of things to process, it stops trying until the next morning. 

edit again: This reads to me as "every five minutes starting at 8AM and ending at 11AM."

If this doesn't make sense, I'll try to explain better in dev sync, but in short: "go fast and frequent right after the CRIS import and then chill for the rest of the day"

edit: Now we're on airflow and we're not paying by the execution, we can do */5 * * * * if we want, but i'd say let's keep this so we don't just generate extra overhead on atd-data03

README.md
```diff
@@ -39,7 +39,7 @@ _AIRFLOW_WWW_USER_USERNAME=admin
 _AIRFLOW_WWW_USER_PASSWORD=<Pick your initial admin password here>
 AIRFLOW_PROJ_DIR=<The absolute path of your Airflow repository checkout>
 # this fernet key is for testing purposes only
-_AIRFLOW__CORE__FERNET_KEYY=PTkIRwL-c46jgnaohlkkXfVikC-roKa95ipXfqST7JM=
+_AIRFLOW__CORE__FERNET_KEY=PTkIRwL-c46jgnaohlkkXfVikC-roKa95ipXfqST7JM=
```
johnclary, line 42: 🙈 🙏

dags/vz_cr3_extract_ocr_narrative.py
johnclary: i have no opinion on schedule, but it would be cool to adopt this pattern for prod-only scheduling if it is agreeable to y'all!
https://github.com/cityofaustin/atd-airflow/blob/112bcb08fc38adf4a3e5c72534b6f4c6d01dc7cd/dags/atd_airflow_docker_image_prune.py#L34

johnclary (approved):
have not tested this but it looks great! 🚢

dags/vz_cr3_extract_ocr_narrative.py
mddilley: Thanks for explaining that, @frankhereford! 🙏

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -0,0 +1,70 @@
...
+        "opfield": "production.AWS Secret key",
+    }
+}
+
+with DAG(
+    dag_id=f"vz_cr3_ocr_narrative_extract_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    # Every 5 minutes, at 8A, 9A, and 10A
+    schedule_interval="*/5 8-10 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    dagrun_timeout=pendulum.duration(minutes=5),
```
frankhereford: I think you need to make this read `dagrun_timeout=duration(minutes=5),` because you're importing just `datetime` and `duration` above. Same for below where you using the duration method and the one above with the `datetime` method.

dags/utils/onepassword.py
```diff
@@ -9,7 +9,7 @@
 
 
 def get_client():
-    """Get oneopassword connect client
+    """Get onepassword connect client
```
frankhereford, line 12: Thank you! You have an eagle eye!

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -0,0 +1,70 @@
...
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": "production.AWS Secret key",
+    }
+}
+
+with DAG(
+    dag_id=f"vz_cr3_ocr_narrative_extract_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    # Every 5 minutes, at 8A, 9A, and 10A
+    schedule_interval="*/5 8-10 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
frankhereford, line 46: I love this pattern. I am going to go back and get this in all the other DAGs. This is great.

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -0,0 +1,70 @@
...
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+default_args = {
+    "owner": "airflow",
+    "description": "Extracts the diagram and narrative out of CR3 pdfs",
+    "depends_on_past": False,
+    "start_date": pendulum.datetime(2019, 1, 1, tz="America/Chicago"),
+    "email_on_failure": False,
+    "email_on_retry": False,
```
frankhereford, line 18: We don't use email alerts in airflow V2, I think these can go.

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -0,0 +1,70 @@
...
+    },
+    "AWS_SECRET_ACCESS_KEY": {
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": "production.AWS Secret key",
+    }
+}
+
+with DAG(
+    dag_id=f"vz_cr3_ocr_narrative_extract_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
```
frankhereford, line 44: Not asking you to change this, but I do want to observe that I'm not quite sure why we define some of our dag parameters in this variable up top and some here. There are arguments in the "default" arguments which are custom to the dag. What do you think? do you think it's cleaner to bring them all down into the DAG decorator?

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -0,0 +1,70 @@
...
+        return load_dict(REQUIRED_SECRETS)
+
+    env_vars = get_env_vars()
+    
+    DockerOperator(
+        task_id="ocr_narrative_extract",
+        image="atddocker/atd-vz-cr3-extract:production",
+        api_version="auto",
+        auto_remove=True,
+        command="./cr3_extract_diagram/cr3_extract_diagram_ocr_narrative.py -v -d --update-narrative --update-timestamp --batch 100 --cr3-source atd-vision-zero-editor production/cris-cr3-files --save-diagram-s3 atd-vision-zero-website cr3_crash_diagrams/production",
```
frankhereford, line 66: Do you think it would be worthwhile to bring some of these S3 paths out of the DAG code and into 1PW? It would let us have a `development` and `production` version. I think this may be more work than it's worth since we really won't be doing a ton of dev on this script. Maybe if/when we do have to extend the script, we can add those sections then.

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -0,0 +1,70 @@
...
+        environment=env_vars,
+        tty=True,
+        force_pull=True,
```
frankhereford, line 69: would you mind adding the following line to squeltch this error: "{docker.py:333} WARNING - Using remote engine or docker-in-docker and mounting temporary volume from host is not supported. Falling back to `mount_tmp_dir=False` mode. You can set `mount_tmp_dir` parameter to False to disable mounting and remove the warning"

```
mount_tmp_dir=False,
```

frankhereford (changes requested):
Chia this is awesome work! Thank you so much! We're going to be off prefect 1 in time thanks to your work!!!

I have a few comments, but the main ones are making it compile with the `pendulum` `datetime` and `duration` changes and the `mount_tmp_dir` one.

Thank you so much!

dags/vz_cr3_extract_ocr_narrative.py
chiaberry: We don't use them, but do they default as True? 
https://airflow.apache.org/docs/apache-airflow/1.10.6/_api/airflow/models/index.html#airflow.models.BaseOperator

dags/vz_cr3_extract_ocr_narrative.py
chiaberry: Not sure, I copied the patterns in the slackbot dags that were approved. I'm thinking that was also copying the patterns we had in v1, but I am not sure about that.

dags/vz_cr3_extract_ocr_narrative.py
chiaberry: I like the idea of adding those sections if we ever need a staging version.

dags/vz_cr3_extract_ocr_narrative.py
chiaberry: thank you ! fixed

dags/vz_cr3_extract_ocr_narrative.py
chiaberry: added!

dags/vz_cr3_extract_ocr_narrative.py
frankhereford: Chia, please forgive me! I absolutely didn't notice that you were setting them to false which is 100% correct. I saw that they were being set, and didn't read closely enough. This is perfect, and I am going to adopt this pattern. Thank you!

frankhereford (approved):
Thank you Chia!! This is awesome. 🚢 🚢 🚢

PR #124, opened 2023-07-13
Make all DAGs be configured with Central Time cron intervals and use Central Time explicitly 

### LINKED ISSUE #12851: Unify Airflow2 DAGs to be run in central time instead of UTC
Standardize the use of central time, even though the servers clock is in UTC.

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/12851

This PR also standardizes on the use of `pendulum` in lieu of `datetime`

## Associated repo

atd-airflow

## Testing

Spin up this branch locally and look at the `cron` interval schedules. Do they make sense in local time, not UTC? Your macs system time is set to central. Interesting, @johnclary's will be set to Eastern, and it should still work there too.

Not sure how to say this, but then please trust that I have tried this on `atd-data03` and seen that the times make sense there, even though it has its system clock set to UTC.

## A note on the changed schedules. 

It's very confusing when we actually were running these due to airflow-1 + atd-data02 + UTC vs. Central.  I am proposing these new times based on my understanding of when we want to run them and on a quick discussion we had in slack. If anyone would like to change these times, please feel free to push on this branch or let me know, and i'll be happy to.

---
#### Ship list
- [ ] Code reviewed


dags/atd_service_bot_index_issues_to_dts_portal.py
```diff
@@ -1,4 +1,5 @@
 import os
+import pendulum
 from datetime import datetime, timedelta
```
johnclary: @frankhereford i was just noodling on this in my own dag. if we're going to lean into `pendulum`, i was thinking we might as well get away from the `datetime` imports all together.

what do you think about this? swapping in `duration` for `timedelta`.

```python
from pendulum import datetime, duration


default_args = {
    "owner": "airflow",
    "description": "Publishes all signal stuides records from Knack to Socrata",
    "depends_on_past": False,
    "start_date": datetime(2015, 1, 1, tz="America/Chicago"),
    "email_on_failure": False,
    "email_on_retry": False,
    "retries": 0,
    "on_failure_callback": task_fail_slack_alert,
}



with DAG(
    dag_id=f"atd_service_bot_issues_to_dts_portal_{DEPLOYMENT_ENVIRONMENT}",
    default_args=default_args,
    schedule_interval="13 7 * * *",
    dagrun_timeout=duration(minutes=60),
    tags=["repo:atd-service-bot", "knack", "github"],
    catchup=False,
) as dag:

```

frankhereford: I love the idea of leaning into pendulum exclusively. I'll update these to do so! Thanks for the great idea.

johnclary (approved):
@frankhereford this is looking right to me! The DAG `atd_service_bot_index_issues_to_dts_portal.py` is  scheduled for 6am when I switch my timezone to EDT:

Screenshot 2023-07-13 at 12 56 26 PM

thanks for sorting this out! it's so much better than staring at `UTC` 
🙌 🚢

dags/atd_service_bot_issues_to_socrata.py
```diff
@@ -59,14 +59,14 @@
 with DAG(
     dag_id=f"atd_service_bot_github_to_socrata_{DEPLOYMENT_ENVIRONMENT}",
     default_args=default_args,
-    schedule_interval="21 5 * * *",
-    dagrun_timeout=timedelta(minutes=60),
+    schedule_interval="0 22 * * *",
```
mddilley, line 62: I think there was some confusion on this schedule with us talking about this one and Xavier talking about the VZ Socrata DAG. I wasn't very specific in [that thread](https://austininnovation.slack.com/archives/CHZE6BC6L/p1689263694947379) in supporting that this one happen at 5am. I think that 10pm works well for this too, though. The risk is having public users seeing partial data in the DTS website and Vision Zero Viewer as the datasets are cleared and rows are replaced in batches.

dags/atd_service_bot_index_issues_to_dts_portal.py
```diff
@@ -1,5 +1,5 @@
 import os
-from datetime import datetime, timedelta
+import pendulum
```
johnclary, line 2: @frankhereford this is super minor, but since we're laying down a pattern that will be copied from here on out: i think we will save tiny bits of RAM on the DAG preprocessor if you target the pendulum module imports:

```python
from pendulum import datetime, duration
```

frankhereford: Great idea John, I'll do that. Let's make our airflow setup as lean and efficient as we can!

dags/atd_service_bot_index_issues_to_dts_portal.py
```diff
@@ -13,7 +13,7 @@
     "owner": "airflow",
     "description": "Create/update 'Index' issues in the DTS portal from Github.",
     "depends_on_past": False,
-    "start_date": datetime(2015, 12, 1),
+    "start_date": pendulum.datetime(2015, 12, 1, tz="America/Chicago"),
```
mddilley, line 16: I keep wondering if a time will come when a DAG wrapper that would set config that we would want to apply to every single DAG ever would make sense, but I think it would be overkill tbh.

mddilley (approved):
Looks great, Frank! Thanks for sorting this out. It makes looking at the schedules way better.⌛ 👀 ✨

chiaberry (approved):

dags/atd_service_bot_issues_to_socrata.py
mddilley: Gosh, while re-reading this, it still sounds like I'm lumping these together into the same process. Two separate ETLs and two separate datasets which might need similar consideration in their schedules. Sorry!

PR #125, opened 2023-07-13
Create DAG to prune docker images

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/12853

As of this morning we've already got some dangling images on the v2 server. It will be good to get this up and running!

```
[claryj@atd-data03 ~]$ docker images --filter=dangling=true
REPOSITORY                   TAG       IMAGE ID       CREATED         SIZE
atddocker/vz-cris-import     <none>    cd3445efba11   3 days ago      819MB
atddocker/vz-cris-import     <none>    c05019a083d4   5 weeks ago     832MB
atddocker/atd-service-bot    <none>    b4684e75fa1b   7 months ago    157MB
atddocker/atd-knack-banner   <none>    4e45abbfab72   12 months ago   441MB
```

## Associated repo
- https://github.com/cityofaustin/atd-airflow

## Testing

You can trigger this DAG on your local airflow setup. It will prune any dangling images on your system.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/toolbox/docker/cron_entry
+++ b/toolbox/docker/cron_entry
@@ -1,0 +1,1 @@
+1 4 * * * root /usr/airflow/atd-airflow/toolbox/docker/image_prune.sh >> /var/log/airflow_image_prune.log
--- a/toolbox/docker/cron_entry
+++ b/toolbox/docker/cron_entry
@@ -1,1 +1,0 @@
-1 4 * * * root /usr/airflow/atd-airflow/toolbox/docker/image_prune.sh >> /var/log/airflow_image_prune.log
--- a/dags/atd_airflow_docker_image_prune.py
+++ b/dags/atd_airflow_docker_image_prune.py
@@ -1,0 +1,43 @@
+"""
+Prunes docker images by running `docker image prune`.
+
+Dangling docker images are common with our dockerized ETLs, because the top-most image
+layers contain ETL code. When that code changes, the previous layer is discarded,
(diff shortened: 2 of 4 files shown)


toolbox/docker/cron_entry
frankhereford, line 1: What would you think about putting this in a DAG? Because we pass down the docker socket file from `atd-data03`, if we run the shell command `docker image prune` as a [BashOperator](https://airflow.apache.org/docs/apache-airflow/stable/howto/operator/bash.html) in a DAG, even though it's inside the airflow-worker container, it'll actually be operating on the host's docker service.

No problem either way, but it might be one less cron entry we have to think about.

johnclary: oooh fun! great call. i'm all for that.

johnclary: thanks @frankhereford for that suggestion. this is much better!

mddilley: dang y'all - this is super cool! ✨

mddilley (approved):
Ran this locally and saw the command logged in Airflow. 🚀 Thanks for adding the note at the top of the DAG too. 📖

chiaberry (approved):
✂️

frankhereford (approved):
That was a really great call John on what parts of the docker system would accumulate cruft. Big TIL for me, thank you!

🚀🚢📦

johnclary:
thanks y'all!

PR #126, opened 2023-07-13
Migrates Knack services DAGs from Airflow v1 to v2

johnclary:
## Associated issues
-  https://github.com/cityofaustin/atd-data-tech/issues/12864

## Associated repo
- atd-knack-services

## Testing

**Steps to test:**
Start your local Airflow stack and trigger the DAG: `atd_knack_signals`.

---
#### Ship list
- [ ] Code reviewed 


dags/utils/onepassword.py
```diff
@@ -38,3 +40,12 @@ def get_item_by_title(entry_name):
...
     client = get_client()
     return client.get_item_by_title(entry_name, VAULT_ID)
+
+
+@task(
+    task_id="get_env_vars",
+    execution_timeout=duration(seconds=30),
+)
+def get_env_vars_task(required_secrets):
+    """Airlfow task to fetch required secrets via load_dict"""
```
chiaberry: this is to wrap the function in an airflow task to cut down on boilerplate, right? just wanting to make sure i understand correctly

also theres a typo in airflow

chiaberry (approved):

chiaberry:
(I didnt run it, but it looks correct!)

dags/utils/onepassword.py
johnclary: yes and thank you! i have tried to improve the doctring.

mddilley:
@johnclary I might be missing something, but I'm getting an error about the `-d` arg not coming through when I tested locally.
`[2023-07-14, 09:40:05 CDT] {docker.py:391} INFO - records_to_postgrest.py: error: argument --date/-d: expected one argument`

johnclary:
bah thanks @mddilley will fix 🙏

johnclary:
ok—i went ahead and wrapped that date filter logic in a task since it got a little bulky. if you don't mind testing again when you get a chance @mddilley 🙏

mddilley (approved):
Tested locally and all the tasks succeeded. ✅ Love seeing the utils get leveled up! 🚀

dags/atd_knack_signals.py
```diff
@@ -0,0 +1,112 @@
...
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration, now
+
+from utils.onepassword import get_env_vars_task
+from utils.slack_operator import task_fail_slack_alert
+from utils.knack import get_date_filter_arg
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
```
Charlie-Henry, line 13: Add a `description`, here's the one from the airflow v1 DAG

```
"description": "Load signals (view_197) records from Knack to Postgrest to AGOL and Socata",  # noqa:E501
```

dags/atd_knack_signals.py
```diff
@@ -0,0 +1,112 @@
...
+}
+
+
+with DAG(
+    dag_id=f"atd_knack_signals",
+    description="Load signals (view_197) records from Knack to Postgrest to AGOL and Socata",
+    default_args=DEFAULT_ARGS,
+    schedule_interval="1 */2 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    dagrun_timeout=duration(minutes=5),
+    tags=["repo:atd-knack-services", "knack", "socrata", "agol"],
```
Charlie-Henry: add  `data-tracker`?

Charlie-Henry (approved):
I looked at our previous [implementation of this in V1](https://github.com/cityofaustin/atd-airflow/blob/a5853a8974c270d1249c56bd147bb86f14718e51/dags/atd_knack_signals.py#L52) and we supplied a of date '1970-01-01' instead of `None` when we wanted to do our monthly refresh of the data. 

It does look like in `records_to_postgrest.py` that it'll [delete everything](https://github.com/cityofaustin/atd-knack-services/blob/8adedf3e4a8bb7b47a464d6048328e84a52041c4/services/records_to_postgrest.py#L93C5-L98C10) and replace it if we didn't supply a date param, and a similar thing happens in `records_to_agol`. We probably want this but just wanted to flag it for you that this is slightly different.

dags/atd_knack_signals.py
johnclary: I moved the description down to the `DAG` params-  that seemed like the right place? 🤷

Charlie-Henry: oops yes that's much better

johnclary: thanks @Charlie-Henry for flagging the change in date filter behavior. as far as i know, the proper truncate/replace behavior (with no date filter) is going to be preferred. iirc, the reason i initially wrote these the  '1970-01-01'  way is because you can't pass `None` to a CLI arg (`--date`) 🤷 

that said, when we deploy to prod, i think we'll want to monkey patch in today's date for `prev_start_date_success` - otherwise the first time these DAGs run they will all run in replace mode, and there are some real big datasets out there.

PR #127, opened 2023-07-14
Fix the cr3 crash diagram dag

chiaberry:
I updated one of the places Frank caught in my PR, but not both places. When I went to turn it on on atd-03, i got an error message

Screenshot 2023-07-14 at 10 08 59 AM

---
#### Ship list
- [ ] Code reviewed 


johnclary (approved):
🚢

mddilley (approved):
🚢 🙌

PR #132, opened 2023-07-18
Make webserver.base_url configurable via environment file with suitable default for local dev

frankhereford:
This PR makes the webserver.base_url configuration value for airflow settable in the environment file. If it's not set, a suitable value for local development will be used.

Please see this [conversation](https://austininnovation.slack.com/archives/C013G4RNJ01/p1689690216401759?thread_ts=1689689445.887899&cid=C013G4RNJ01). Thanks to @johnclary and @chiaberry for the idea!

The approach above was inspected, and then a different, final solution was arrived at on advice from @johnclary in slack PMs. Thank you!

---
#### Ship list
- [ ] Code reviewed

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,6 @@ The stack is composed of:
 
 ## Table of Contents
 
-
 - [Getting Started](#getting-started)
   - [Developing a new DAG](#developing-a-new-dag)
   - [Tags](#tags)
--- a/README.md
+++ b/README.md
@@ -14,17 +14,21 @@ The stack is composed of:
 
 ## Table of Contents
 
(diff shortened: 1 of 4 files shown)


frankhereford:
I'm withdrawing request for review briefly. John suggested another approach which is less hacky, and I'd like to pursue that. Thanks!

johnclary (approved):
thanks @frankhereford for working through this. i can't wait to have our log URLs fixed! 
🚢 🚢 🚢 🚢

mddilley (commented):
Cool! I was wondering how this was going to work. 🚀

PR #133, opened 2023-07-18
Update CR3 download script

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12849

This has the changes that I showed in dev sync, and I'm open to all thoughts on this. I'm seeing a few ways to go here and we can always make multiple ways possible. Here is what some of those would look like:
- Having the DAG run routinely to watch a 1PW cookie entry for changes
   - Go through the steps to get the cookie (we have to think about how long a session would live)
   - Update the cookie entry in 1Password
   - wait for the DAG to see the update, run, and see a success alert to the Airflow slack channel
- Manually run from the Airflow dashboard with a parameter
   - Go through the steps to get the cookie
   - Get on the VPN and login to the Airflow v2 dashboard
   - Run the DAG manually and feed the cookie in a parameter
- Same as the last one but update the cookie 1Password entry and then manually run it instead of using an Airflow parameter

## Associated repo
https://github.com/cityofaustin/atd-vz-data/pull/1254

## Testing

**Steps to test:**
I pushed a [`development` tag to DockerHub](https://hub.docker.com/layers/atddocker/vz-cr3-download/development/images/sha256-b8cc0ec5f337c372fa5ed46ae91d6a827900c11d48371735b12e24fb9d4b9d2f?context=repo) that has the code changes in the VZ PR linked above.

Spin up the local stack and trigger the `vz-cr3-download` DAG. It will skip the `vz_cr3_download` task. Update the `CRIS CR3 Download Cookie` 1PW entry and trigger it again. You should see the modified date shown in 1PW update. It should run the `vz_cr3_download` task and log that there are no pdfs to download (unless you run it in the morning before the designated CRIS downloader runs the current process).

If you want to go the extra mile, you can: 
- find an older crash record without serious injuries or fatalities (we don't want to disrupt VZ team efforts)
- carefully, [SSH tunnel and connect to the production database](https://app.gitbook.com/o/-LzDQOVGhTudbKRDGpUA/s/-MIQvl_rKnZ_-wHRdp4J/dev-guides/how-tos/how-to-connect-to-the-rds-instance#2.-open-the-tunnel) and update the one crash record's `cr3_stored_flag` column from `Y` to `N` (i tested this with crash id `16302089`)
- run the script and it will log one pdf download from CRIS and upload to S3

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/vz_cr3_download.py
+++ b/dags/vz_cr3_download.py
@@ -1,0 +1,56 @@
+import os
+
+from airflow.decorators import task
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration
+
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+default_args = {
+    "owner": "airflow",
+    "description": "Downloads CR3 pdfs from CRIS and uploads to S3",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 1, 1, tz="America/Chicago"),
+    "retries": 0,
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+REQUIRED_SECRETS = {}
+
+with DAG(
+    dag_id=f"vz-cr3-download",
+    default_args=default_args,
+    schedule_interval=None,
+    dagrun_timeout=duration(minutes=5),
+    tags=["repo:atd-vz-data", "cris", "s3", "cr3"],
+    catchup=False,
+) as dag:
+    docker_image = "atddocker/"
+
+    @task(
+        task_id="get_env_vars",
+        execution_timeout=duration(seconds=30),
+    )
+    def get_env_vars():
+        from utils.onepassword import load_dict
+
+        return load_dict(REQUIRED_SECRETS)
+
+    env_vars = get_env_vars()
+
+    t1 = DockerOperator(
+        task_id="vz_cr3_download",
+        image=docker_image,
+        api_version="auto",
+        auto_remove=True,
+        command="",
+        environment=env_vars,
+        tty=True,
+        force_pull=True,
+        mount_tmp_dir=False,
+    )
+
+    t1
--- a/dags/vz_cr3_download.py
+++ b/dags/vz_cr3_download.py
@@ -8,6 +8,7 @@ from pendulum import datetime, duration, now, parse
 
 from utils.onepassword import get_env_vars_task, get_item_last_update_date
 from utils.slack_operator import task_fail_slack_alert, task_success_slack_alert
+from utils.time import get_previous_run_date
 
 DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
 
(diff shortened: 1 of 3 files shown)


dags/vz_cr3_download.py
```diff
@@ -0,0 +1,102 @@
...
+from pendulum import datetime, duration, now, parse
+
+from utils.onepassword import get_env_vars_task, get_item_last_update_date
+from utils.slack_operator import task_fail_slack_alert, task_success_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+default_args = {
+    "owner": "airflow",
+    "description": "Downloads CR3 pdfs from CRIS and uploads to S3",
```
johnclary: i've had a hard time finding the docs on this, but i'm pretty sure you want to pass this property to the `DAG()` constructor instead. we have a bad pattern in our v1 DAGs.

dags/vz_cr3_download.py
```diff
@@ -0,0 +1,102 @@
...
+        force_pull=True,
+        mount_tmp_dir=False,
+        on_success_callback=task_success_slack_alert,
+    )
+
+    branch_false = DummyOperator(task_id="skip_vz_cr3_download")
+
+    @task.branch(task_id="choose_branch")
+    def choose_branch(updated_at):
+        minutes_ago_utc = now() - duration(minutes=5)
```
johnclary: i'm not sure if this is actually want we want, but one suggestion would be to use the `prev_start_date_success` instead of "five minutes ago". so the DAG runs anytime the cookie is newer than the last run 🤷 

[here's an example](https://github.com/cityofaustin/atd-airflow/blob/a724d183451041be11a603666a4977e4563dc4dd/dags/utils/knack.py#L23) of pulling that out of the task context

johnclary (approved):
@mddilley this is such a clever approach, and i think it will be really nice to have the download script running alongside our other DAGs in Airflow 🚀

dags/vz_cr3_download.py
```diff
@@ -0,0 +1,102 @@
...
+        updated_at_utc = updated_at
+
+        if updated_at_utc > minutes_ago_utc:
+            print("Downloading CR3 pdfs")
+            return ["vz_cr3_download"]
+        else:
+            print("Cookie entry not updated - skipping CR3 pdf downloads")
+            return ["skip_vz_cr3_download"]
+
+    choose_branch = choose_branch(updated_at=updated_at_dict["updated_at"])
```
johnclary: by the way, this is awesome! 🔥

dags/vz_cr3_download.py
mddilley: yess - thank you! i kept focusing on synchronizing the DAG schedule and this duration. Making this change would take care of that.

mddilley: I added this and a task util to return last run date. I ran into an issue getting the last date because the `@task.branch` decorator doesn't support the `**context` arg. Just decided to taskify it.

frankhereford (approved):
I've reviewed the code and tested it in my local stack, and it worked perfectly. I don't see anything in this that I would change at all. 💯 

This is a verrry clever use of metadata provided by 1Password, and it does wonders to make this process a more centralized and reproducible experience. This is next level use of the 1PW / Airflow integration capabilities!

Mike - thank you for putting this together and for the creativity that you brought to this problem!!!

dags/vz_cr3_download.py
```diff
@@ -0,0 +1,104 @@
...
+        task_id="vz_cr3_download",
+        image=f"atddocker/vz-cr3-download:{DEPLOYMENT_ENVIRONMENT}",
+        api_version="auto",
+        auto_remove=True,
+        command="python cr3_download.py",
+        environment=env_vars,
+        tty=True,
+        force_pull=True,
+        mount_tmp_dir=False,
+        on_success_callback=task_success_slack_alert,
```
chiaberry, line 85: I'm glad this is here!

PR #134, opened 2023-07-19
change certbot script permissions

johnclary:
These changes were on atd-data03 so i pushed them up here. Interestingly, it looks like frank's SSH key is taking precedence in the ssh config on that machine - that has to be my bad 😬


frankhereford (approved):
`chmod 755` looks good to me, thank you!

johnclary:
thanks @frankhereford FYI my updated note in the PR description 👀

mddilley (commented):
Thanks for pushing this and catching that there was a diff on atd-data03 after I updated those permissions. 🙏

PR #135, opened 2023-07-19
Migrate VZ Socrata ETL to Airflow v2

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12495

## Associated repo
https://github.com/cityofaustin/atd-vz-data/pull/1258

## Testing

**Steps to test:**

This screenshot shows the script working using the local VZ stack + local Airflow stack, because it is kind of a lot to ask y'all to patch this and the Airflow stack to make this work.

Screenshot 2023-07-19 at 7 50 24 PM

**If you decide to try to test this locally, please coordinate with @tillyw since it affects the micromobility VZV work that she is working on.** I tested this using my local VZ stack only after carrying out the steps in https://github.com/cityofaustin/atd-vz-data/pull/1252 to make sure this doesn't block that ongoing work.

1. Update the [VZ stack `docker-compose.yaml` on line 16](https://github.com/cityofaustin/atd-vz-data/blob/825362d288087e78a6ec713e422cb33d657062ed/docker-compose.yml#L16) with the following change:
```
- 8084:8080
```
2. This is to avoid the VZ graphql engine and Airflow webserver from colliding ports.
3. Start the local VZ stack and follow the steps in https://github.com/cityofaustin/atd-vz-data/pull/1252 to make sure the micromobility data makes its way into this test dataset.
4. Patch this DAG so that the value of `HASURA_ENDPOINT` is `http://host.docker.internal:8084/v1/graphql`.
5. Run the DAG

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/dags/vz_socrata_export.py
+++ b/dags/vz_socrata_export.py
@@ -20,12 +20,12 @@ default_args = {
 
 REQUIRED_SECRETS = {
     "HASURA_ENDPOINT": {
-        "opitem": "Vision Zero CRIS Import",
+        "opitem": "Vision Zero graphql-engine Endpoints",
         "opfield": f"{DEPLOYMENT_ENVIRONMENT}.GraphQL Endpoint",
     },
     "HASURA_ADMIN_KEY": {
-        "opitem": "Vision Zero CRIS Import",
-        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.GraphQL Endpoint key",
+        "opitem": "Vision Zero graphql-engine Endpoints",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.Admin Key",
     },
     "SOCRATA_API_KEY_ID": {
         "opitem": "Socrata Key ID, Secret, and Token",
@@ -52,7 +52,7 @@ REQUIRED_SECRETS = {
 with DAG(
     dag_id=f"vz_socrata_export_{DEPLOYMENT_ENVIRONMENT}",
     default_args=default_args,
-    schedule_interval="0 9 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    schedule_interval="0 4 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
     dagrun_timeout=duration(minutes=20),
     tags=["repo:atd-vz-data", "socrata"],
     catchup=False,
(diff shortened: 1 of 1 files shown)


dags/vz_socrata_export.py
```diff
@@ -0,0 +1,72 @@
...
+from airflow.operators.docker_operator import DockerOperator
+
+from utils.slack_operator import task_fail_slack_alert
+from utils.onepassword import get_env_vars_task
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+default_args = {
+    "owner": "airflow",
+    "description": "Downloads CR3 pdfs from CRIS and uploads to S3",
```
johnclary: This can move to the DAG constructor

dags/vz_socrata_export.py
```diff
@@ -0,0 +1,72 @@
...
+    "SOCRATA_DATASET_PERSONS": {
+        "opitem": "Vision Zero Socrata Export",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.SOCRATA_DATASET_PERSONS",
+    },
+}
+
+with DAG(
+    dag_id=f"vz_socrata_export_{DEPLOYMENT_ENVIRONMENT}",
+    default_args=default_args,
+    schedule_interval="0 9 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
johnclary: it looks like this is running at 9am UTC on airflow v1, so i think this needs to be adjusted to CDT.

johnclary (approved):
i'll take your word for it! thanks for getting this ported! plz check double-check that schedule 🚢 🚢 🚢 🚢 🚢

frankhereford (approved):
Thank you! As noted in the other PR review, I've only gone over this code, and I think it looks great. 🚢 🚢 🚢

dags/vz_socrata_export.py
mddilley: Thanks for catching this again.... too much copy/paste on my part. 🙏

dags/vz_socrata_export.py
mddilley: 🙏 Will update this

chiaberry (approved):
thanks for including the screenshot of the output, i am not going to run it locally

just update the dag's description and its 🚢

mddilley:
Thanks @chiaberry - copy/paste strikes again! 📝

PR #139, opened 2023-07-24
atd-signal-comms: create start date task

johnclary:
I overlooked that this script is fussy about timestamp formatting: it requires exactly `yyyy-mm-dd`. This takes care of that.

### DIFF
--- a/dags/atd_signal_comms.py
+++ b/dags/atd_signal_comms.py
@@ -185,7 +185,8 @@ with DAG(
     )
 
     (
-        cameras_s3
+        start_date
+        >> cameras_s3
         >> detectors_s3
         >> dms_s3
         >> battery_backup_s3
--- a/dags/atd_signal_comms.py
+++ b/dags/atd_signal_comms.py
@@ -2,11 +2,11 @@ import os
 
 from airflow.models import DAG
 from airflow.operators.docker_operator import DockerOperator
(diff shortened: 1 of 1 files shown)


chiaberry (approved):
✔️

dags/atd_signal_comms.py
```diff
@@ -61,6 +61,17 @@
     },
 }
 
+@task(
+    task_id="get_start_date",
+)
+def get_start_date(**context):
```
mddilley, line 67: Lots of development happening in parallel so I wanted to share that I started a [time helpers file](https://github.com/cityofaustin/atd-airflow/blob/4ca029cf9702600f51cf48c9457473233a742479/dags/utils/time.py) in another open PR. Yours is slightly different though and is making me think that I should have used `now()` there too.

mddilley (approved):
⏳🚀

johnclary:
thanks y'all!

johnclary:
it ran on prod 😅 
Screenshot 2023-07-24 at 2 18 03 PM

dags/atd_signal_comms.py
johnclary: ah, thanks for pointing this out. i'm going to think about this some more since i'm annoyed the knack one i made is not usable in other domains.

PR #140, opened 2023-07-25
ETL DAG ports from Airflow V1 for VZ location & CR3 maintenance

frankhereford:
## Associated issues

This PR puts a couple more check marks on: https://github.com/cityofaustin/atd-data-tech/issues/12860

Sibling PR: https://github.com/cityofaustin/atd-vz-data/pull/1257

## Associated repo

`atd-vz-data`

## Testing

* Configure your local VZ `graphql-engine` service to listen on port `8084` if needed
* Spin up local VZDB & local `graphql-engine`
* Spin up local airflow
* Trigger `vision_zero_process_missing_pdfs`
* Trigger `vision_zero_reassociate_missing_locations`

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_visionzero_cr3_process_missing_pdfs.py
+++ b/dags/atd_visionzero_cr3_process_missing_pdfs.py
@@ -53,7 +53,7 @@ with DAG(
     tags=["repo:atd_vz_data", "vision-zero", "cr3", "pdf"],
     catchup=False,
 ) as dag:
-    docker_image = "atddocker/vz-pdf-maintenance:latest"
+    docker_image = "atddocker/vz-pdf-maintenance:production"
 
     env_vars = get_env_vars_task(REQUIRED_SECRETS)
 
--- a/dags/atd_visionzero_cr3_process_missing_pdfs.py
+++ b/dags/atd_visionzero_cr3_process_missing_pdfs.py
@@ -53,7 +53,7 @@ with DAG(
     tags=["repo:atd_vz_data", "vision-zero", "cr3", "pdf"],
     catchup=False,
 ) as dag:
-    docker_image = "atddocker/process_missing_cr3_pdfs:production"
+    docker_image = "atddocker/vz-process-missing-cr3-pdfs:production"
 
     env_vars = get_env_vars_task(REQUIRED_SECRETS)
 
--- a/dags/atd_visionzero_cr3_process_missing_pdfs.py
+++ b/dags/atd_visionzero_cr3_process_missing_pdfs.py
@@ -5,7 +5,6 @@ from airflow.operators.docker_operator import DockerOperator
 from pendulum import datetime, duration
 
 from utils.onepassword import get_env_vars_task
-from utils.knack import get_date_filter_arg
 from utils.slack_operator import task_fail_slack_alert
 
 DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
--- a/dags/atd_visionzero_cr3_process_missing_pdfs.py
+++ b/dags/atd_visionzero_cr3_process_missing_pdfs.py
@@ -43,7 +43,7 @@ REQUIRED_SECRETS = {
 }
 
 with DAG(
-    dag_id="vision_zero_process_missing_pdfs",
+    dag_id="vz_process_missing_pdfs",
     description="Execute housekeeping routine manage missing or malformed CR3 PDFs",
     default_args=DEFAULT_ARGS,
     schedule_interval="*/30 * * * *"
--- a/dags/atd_visionzero_cr3_process_missing_pdfs.py
+++ b/dags/atd_visionzero_cr3_process_missing_pdfs.py
@@ -1,0 +1,82 @@
+import os
+
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration
+
+from utils.onepassword import get_env_vars_task
+from utils.knack import get_date_filter_arg
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 1, 1, tz="America/Chicago"),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "retry_delay": duration(minutes=5),
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+# secrets atd_visionzero_hasura_sql_production
+
+REQUIRED_SECRETS = {
+    "HASURA_ENDPOINT": {
+        "opitem": "Vision Zero graphql-engine Endpoints",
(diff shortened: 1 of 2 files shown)


dags/atd_visionzero_cr3_process_missing_pdfs.py
```diff
@@ -0,0 +1,84 @@
...
+    "HASURA_ENDPOINT": {
+        "opitem": "Vision Zero graphql-engine Endpoints",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.Query Endpoint",
+    },
+    "HASURA_ADMIN_KEY": {
+        "opitem": "Vision Zero graphql-engine Endpoints",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.Admin Key",
+    },
+    "AWS_BUCKET_NAME": {
+        "opitem": "CRIS CR3 Download",
```
johnclary: Super minor—and I don't know what other ETLs depend on this item name—but it would be nice to have "Vision Zero" in the names of these "CRIS" entries. Ie rename this to `Vision Zero CRIS CR3 Download` .

Screenshot 2023-07-25 at 1 11 50 PM

dags/atd_visionzero_cr3_process_missing_pdfs.py
```diff
@@ -0,0 +1,84 @@
...
+    schedule_interval="*/30 * * * *"
+    if DEPLOYMENT_ENVIRONMENT == "production"
+    else None,
+    dagrun_timeout=duration(minutes=15),
+    tags=["repo:atd_vz_data", "vision-zero", "cr3", "pdf"],
+    catchup=False,
+) as dag:
+    docker_image = "atddocker/vz-pdf-maintenance:latest"
+
+    date_filter_arg = get_date_filter_arg(should_replace_monthly=True)
```
johnclary: i think this is copypasta cruft since you're not using it 😁

dags/atd_visionzero_cr3_process_missing_pdfs.py
```diff
@@ -0,0 +1,84 @@
...
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.AWS_CRIS_CR3_BUCKET_NAME",
+    },
+    "AWS_BUCKET_ENVIRONMENT": {
+        "opitem": "CRIS CR3 Download",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.AWS_CRIS_CR3_BUCKET_PATH_ENVIRONMENT",
+    },
+}
+
+with DAG(
+    dag_id="vision_zero_process_missing_pdfs",
```
johnclary: fwiw our other Vision Zero DAGs follow the naming convention `vz_blah_blah` instead of `vision_zero_blah_blah`

johnclary (dismissed):
Niiice! Awesome to see these ready to move over from Airflow v1.

Just a few minor things here 🙌

dags/atd_visionzero_cr3_process_missing_pdfs.py
mddilley: this has me written all over it 🙋 I just checked production with my latest DAGs merged, and this only affects the CR3 download DAG that isn't currently working. Feel free to update this in 1PW and in this DAG! I can adjust the other script 👍

johnclary: thanks Mike!

dags/atd_visionzero_cr3_process_missing_pdfs.py
```diff
@@ -0,0 +1,84 @@
...
+    description="Execute housekeeping routine manage missing or malformed CR3 PDFs",
+    default_args=DEFAULT_ARGS,
+    schedule_interval="*/30 * * * *"
+    if DEPLOYMENT_ENVIRONMENT == "production"
+    else None,
+    dagrun_timeout=duration(minutes=15),
+    tags=["repo:atd_vz_data", "vision-zero", "cr3", "pdf"],
+    catchup=False,
+) as dag:
+    docker_image = "atddocker/vz-pdf-maintenance:latest"
```
chiaberry: Should this be tagged "production" instead of latest?
Also, I don't see this docker image name in the vz-data repo, but I do see one named `atddocker/process_missing_cr3_pdfs`

dags/atd_visionzero_reassociate_missing_locations.py
```diff
@@ -0,0 +1,116 @@
+import os
+
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration
+
+from utils.onepassword import get_env_vars_task
+from utils.knack import get_date_filter_arg
```
chiaberry: 👀

dags/atd_visionzero_cr3_process_missing_pdfs.py
frankhereford: Thank you both. This is a great idea, I totally agree. I've pushed a change reflects the change of the 1PW entry "CRIS CR3 Download" into "Vision Zero CRIS CR3 Download." Thanks again!

dags/atd_visionzero_cr3_process_missing_pdfs.py
frankhereford: it sure is cruft - thank you for catching that John!

dags/atd_visionzero_cr3_process_missing_pdfs.py
frankhereford: Thank you for this observation. Great catch - I've unified the naming scheme.

dags/atd_visionzero_cr3_process_missing_pdfs.py
frankhereford: re `:production`, it absolutely should. Thank you for catching that! 

Thank you also for catching that I had image name mismatch between this repo and the VZ one. I've fixed that.

dags/atd_visionzero_reassociate_missing_locations.py
```diff
@@ -0,0 +1,116 @@
...
+    # This process will find the locations for Non-CR3 crashes that do not have one but
+    # fall into a location and they are not mainlanes.
+    update_noncr3_locations = DockerOperator(
+        task_id="update_noncr3_locations",
+        image=docker_image,
+        auto_remove=True,
+        command="python scripts/atd_vzd_update_noncr3_locations.py",
+        environment=env_vars,
+        tty=True,
+        force_pull=True,
```
chiaberry: Charlie just opened a PR that set force_pull=False after the initial force pull, what do you think about doing that in this PR too?

frankhereford: o0o That is smart. I'm going to do that right now.

mddilley: ~hey y'all - is this an update that we should make across the board? i think that, in earlier DAGs, we misunderstood what this does (pull if new vs. pull every run) and sprinkled it all over a bunch of ones that use DockerOperator.~ Scratch this - I just realized what is going on here. Multiple DockerOperators using the same image in the same DAG. 👍

mddilley (approved):
Tested these locally, and all green here. ✅ 🚀 Thanks for adding the comments about what each script does in here too. 🙌

PR #143, opened 2023-07-26
Use `data_interval_start` instead of `execution_date` for Slack webhook messages

mddilley:
## Associated issues
n/a

Fixes `AirflowContextDeprecationWarning: Accessing 'execution_date' from the template is deprecated and will be removed in a future version. Please use 'data_interval_start' or 'logical_date' instead.` and formats the execution time to be in CST and be more readable. If anyone prefers the timestamp format, let me know and I'm happy to change it back.

See: 
- https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/taskflow.html#context
- https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/timezone.html#templates

## Associated repo
n/a

## Testing

**Steps to test:**
1. You can test this locally by commenting out line 42 and then printing `slack_msg` after line 55 and then triggering the Slack webhook test DAG
2. Or, you can set up the webhook locally if you really want to 🚨

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/utils/slack_operator.py
+++ b/dags/utils/slack_operator.py
@@ -1,9 +1,18 @@
 from airflow.hooks.base_hook import BaseHook
 from airflow.contrib.operators.slack_webhook_operator import SlackWebhookOperator
+from pendulum import timezone
 
 # This is the Conn Id that we set when creating the connection in the Airflow dashboard
 # in Admin > Connections.
 SLACK_CONN_ID = "slack"
+
+
+def get_central_time_exec_data(context):
+    local_tz = timezone("America/Chicago")
+    execution_date_timestamp = context.get("data_interval_start")
+    central_execution_date = local_tz.convert(execution_date_timestamp).format(
+        "MM/DD/YYYY hh:mm:ss A"
+    )
 
 
 def task_fail_slack_alert_critical(context):
@@ -43,7 +52,7 @@ def task_fail_slack_alert(context):
         task=context.get("task_instance").task_id,
         dag=context.get("task_instance").dag_id,
         ti=context.get("task_instance"),
-        exec_date=context.get("execution_date"),
+        exec_date=context.get("ts"),
         log_url=context.get("task_instance").log_url,
     )
     failed_alert = SlackWebhookOperator(
--- a/dags/utils/slack_operator.py
+++ b/dags/utils/slack_operator.py
@@ -1,6 +1,5 @@
 from airflow.hooks.base_hook import BaseHook
 from airflow.contrib.operators.slack_webhook_operator import SlackWebhookOperator
-from pendulum import timezone
 
 # This is the Conn Id that we set when creating the connection in the Airflow dashboard
 # in Admin > Connections.
@@ -8,6 +7,8 @@ SLACK_CONN_ID = "slack"
 
 
 def get_central_time_exec_data(context):
+    from pendulum import timezone
+
     local_tz = timezone("America/Chicago")
     execution_date_timestamp = context.get("data_interval_start")
     return local_tz.convert(execution_date_timestamp).format("MM/DD/YYYY hh:mm:ss A")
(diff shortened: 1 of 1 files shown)


dags/utils/slack_operator.py
```diff
@@ -1,11 +1,18 @@
 from airflow.hooks.base_hook import BaseHook
 from airflow.contrib.operators.slack_webhook_operator import SlackWebhookOperator
+from pendulum import timezone
 
 # This is the Conn Id that we set when creating the connection in the Airflow dashboard
 # in Admin > Connections.
 SLACK_CONN_ID = "slack"
 
 
+def get_central_time_exec_data(context):
```
Charlie-Henry, line 9: is `context` an object that airflow passes around to every function? I keep seeing this and don't fully understand

Charlie-Henry (approved):
ty!

frankhereford (approved):
Thank you Mike. I've gone over the code but not run anything. Looks good to me!

dags/utils/slack_operator.py
johnclary: i guess the `context` is passed as the only argument when a hook executes.

it's a little different for tasks, where the entire context gets passed as keyword arguments after your custom arguments, so you can do [this](https://github.com/cityofaustin/atd-airflow/blob/4c2d4b6d067e1e77053480214c7856e13b1979d5/dags/utils/knack.py#L8). it's annoyingly hard to find this in the documentation :/

dags/utils/slack_operator.py
```diff
@@ -1,11 +1,18 @@
 from airflow.hooks.base_hook import BaseHook
 from airflow.contrib.operators.slack_webhook_operator import SlackWebhookOperator
+from pendulum import timezone
```
johnclary: i don't know how much this matters! and i keep forgetting to do this ([for example](https://github.com/cityofaustin/atd-airflow/blob/4c2d4b6d067e1e77053480214c7856e13b1979d5/dags/utils/knack.py#L2)). but i think the optimal approach would be to import `timezone` within the `get_central_time_exec_data` function. otherwise it is imported every time each DAG is parsed, which happens every 5 minutes.

🤷

dags/utils/slack_operator.py
```diff
@@ -1,11 +1,18 @@
...
 
 # This is the Conn Id that we set when creating the connection in the Airflow dashboard
 # in Admin > Connections.
 SLACK_CONN_ID = "slack"
 
 
+def get_central_time_exec_data(context):
+    local_tz = timezone("America/Chicago")
+    execution_date_timestamp = context.get("data_interval_start")
+    return local_tz.convert(execution_date_timestamp).format("MM/DD/YYYY hh:mm:ss A")
```
johnclary, line 14: love this fancy timezone handling!

dags/utils/slack_operator.py
mddilley: oh yeah, good call - here are some docs on it https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/taskflow.html#context. Added this link and their blurb about time zones in the PR description too 🙏

Charlie-Henry: ah that makes sense ty John 🧙, is the only reason you had to `**kwarg` it there was because of the `should_replace_monthly` arg or because it's a task?

johnclary: because it's a task. so this would also be valid:

```python
@task
def get_date_filter_arg(**context):
   return context.get("some_context_var")
```

but here context would be `None` unless you provided a value when you called the task.

```python
@task
def get_date_filter_arg(context):
   return context
```

you should also theoretically be able to specify the specific context variable of interest as an argument, like this:

```python
@task
def get_date_filter_arg(prev_execution_date_success):
   return prev_execution_date_success
```

but i think i got an airflow error when i tried that.

johnclary: thanks @mddilley!

mddilley: you are right that the docs are not specific enough about using context, John! thanks for adding the examples, and, to me, the docs read like your last example should work too.

dags/utils/slack_operator.py
mddilley: Just updated and tested this. Thanks for catching this because I keep forgetting about this too. 🙏

PR #149, opened 2023-07-28
Migrate Knack <> Banner DAG from Prefect to Airflow

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/13084

## Associated repo
- [atd-knack-banner](https://github.com/cityofaustin/atd-knack-banner)

## Testing
- Once [this PR](https://github.com/cityofaustin/atd-knack-banner/pull/13) is merged, you can test this locally by getting on VPN and triggering the DAG.

I think this will still fail because of a data quality issue discussed [here](https://austininnovation.slack.com/archives/C026ZRP1TH8/p1691425344344869). But the DAG should run and report an error related to a duplicate email address in Knack.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_knack_banner.py
+++ b/dags/atd_knack_banner.py
@@ -46,7 +46,7 @@ with DAG(
     dag_id=f"atd_knack_banner",
     description="Update knack HR app based on records in Banner and CTM",
     default_args=DEFAULT_ARGS,
-    schedule_interval="45 8 * * *" if DEPLOYMENT_ENVIRONMENT == "prod" else None,
+    schedule_interval="45 7 * * *" if DEPLOYMENT_ENVIRONMENT == "prod" else None,
     dagrun_timeout=duration(minutes=30),
     tags=["repo:atd-knack-banner", "knack"],
     catchup=False,
--- a/dags/atd_knack_banner.py
+++ b/dags/atd_knack_banner.py
@@ -64,7 +64,6 @@ with DAG(
         tty=True,
         force_pull=True,
         mount_tmp_dir=False,
-        network_mode="host",
     )
 
     update_employees
--- a/dags/atd_knack_banner.py
+++ b/dags/atd_knack_banner.py
@@ -1,0 +1,70 @@
+import os
+
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+from airflow.decorators import task
+from pendulum import datetime, duration
+
+from utils.onepassword import get_env_vars_task
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 1, 1, tz="America/Chicago"),
+    "email_on_failure": False,
+    "email_on_retry": False,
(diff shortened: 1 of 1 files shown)


dags/atd_knack_banner.py
```diff
@@ -0,0 +1,70 @@
...
+        "opfield": f"production.apiKey",
+    },
+}
+
+
+with DAG(
+    dag_id=f"atd_knack_banner",
+    description="Update knack HR app based on records in Banner and CTM",
+    default_args=DEFAULT_ARGS,
+    schedule_interval="45 8 * * *" if DEPLOYMENT_ENVIRONMENT == "prod" else None,
```
johnclary: this was `45 13 * * *` on Prefect, which had it running at 745 or 845 depending on daylight savings.

chiaberry (approved):

dags/atd_knack_banner.py
```diff
@@ -0,0 +1,70 @@
...
+        "opfield": f"production.apiKey",
+    },
+}
+
+
+with DAG(
+    dag_id=f"atd_knack_banner",
+    description="Update knack HR app based on records in Banner and CTM",
+    default_args=DEFAULT_ARGS,
+    schedule_interval="45 7 * * *" if DEPLOYMENT_ENVIRONMENT == "prod" else None,
```
Charlie-Henry: In some places we're using `DEPLOYMENT_ENVIRONMENT == "production"`? Which is this variable set to?

https://github.com/cityofaustin/atd-airflow/blob/7fc4f7e7ce29f5cc6b6ce5ed1cf66bdac9aa94a2/dags/atd_kits_sig_stat_pub.py#L60

dags/atd_knack_banner.py
```diff
@@ -0,0 +1,70 @@
...
+    update_employees = DockerOperator(
+        task_id="update_employees",
+        image=docker_image,
+        auto_remove=True,
+        command=f"./atd-knack-banner/update_employees.py",
+        environment=env_vars,
+        tty=True,
+        force_pull=True,
+        mount_tmp_dir=False,
+        network_mode="bridge",
```
Charlie-Henry: is `bridge` needed when accessing things on the city network or is it just for the traffic signal network?

Charlie-Henry (approved):
Might want to take a look at `prod` vs `production` when you deploy it but 🚢

dags/atd_knack_banner.py
johnclary: for a long time i thought `bridge` here was equivalent to the `host` network mode in the docker CLI. but now i'm not sure. i think this might be unnecessary: i will test it.

dags/atd_knack_banner.py
johnclary: thanks for catching this! i copied this from `atd-signal-comms` which uses the value `prod` b/c it matches a script's CLI arg.

johnclary: https://github.com/cityofaustin/atd-airflow/blob/7fc4f7e7ce29f5cc6b6ce5ed1cf66bdac9aa94a2/dags/atd_signal_comms.py#L11

dags/atd_knack_banner.py
johnclary: this works without any `network_mode` setting. i removed it 🎉

johnclary:
it is deployed ✅

PR #152, opened 2023-08-04
Remove a VZ DAG from the DAG pool (+ Discussion)

frankhereford:
# Removing a ETL we have turned off in Airflow V2

### How do we want to handle DAGs which we are no longer going to schedule. Do we?

* Just turn it off, in case we want to turn it back on?
* Change the DAG's description in its decorator to explain why it's off?
* Remove the DAG file from the repo?
  * And also remove the tool from the associated repo, if possible?
* Other options?

I'm leaning forward removing the DAG from the repo, but this PR is intended to be a place to discuss that choice. I'm super happy if we don't merge this, particularly if we come up with a protocol that we accept as a group.

Thanks!

## Associated issues

_none_

## Associated repo

atd-vz-data

## Testing

none; this is the removal of a DAG

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 


johnclary:
If this is a maintenance script that can be run ad-hoc from the VZ repo, then i agree we should remove it from Airflow and make sure the script is well-documented elsewhere. 🙏 🚢

mddilley:
I agree. We don't intend to run it anymore under the assumption that the CR3 downloader is preventing the source of the non-pdf files this DAG is intended to clean up. Removing it here and keeping it as a standalone tool in the VZ repo feels right to me. 👍 🙏

mddilley (approved):

chiaberry (approved):
if a dag is never going to be run again, i think its fine to remove from the repo. If its being turned off but might be turned back on the in the future, leaving it in the repo doesn't feel like a bad idea. Thanks for cleaning things up!

johnclary:
@frankhereford i think you're all clear to merge 🚢

PR #155, opened 2023-08-07
Hourly Schedule for disseminating DMS Data

Charlie-Henry:
DMS message data gets sent to knack on an hourly basis at 21 past the hour. This PR would sync up this other ETL that takes that knack data to Socrata and AGOL. These datasets would now be updated hourly, with a 3 minute delay between the two ETLs. Currently it just shares updates daily, even though the underlying data in Knack is updated hourly.

## Associated repo

## Testing

**Steps to test:**
Really shouldn't need to test this, since I'm just changing the schedule. 

---
#### Ship list
- [ ] Code reviewed 


chiaberry (approved):
more!

johnclary (approved):
very good! ⛵

mddilley (approved):
🚀

PR #159, opened 2023-08-22
Port atd-finance-data to Airflow

Charlie-Henry:
Moving this [prefect flow](https://github.com/cityofaustin/atd-prefect/blob/main/flows/atd-finance-data/finance_data_to_s3.py) to airflow 2. I created a separate DAG for each type of data rather than making one giant DAG. Schedule is offset by 5 minutes per dataset.

## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/12376

## Associated repo
https://github.com/cityofaustin/atd-finance-data

## Testing

**Steps to test:**
1. Checkout this branch
2. Feel free to test any of these with the command at the top of the file or using the airflow UI

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_finance_data_fdus.py
+++ b/dags/atd_finance_data_fdus.py
@@ -1,0 +1,162 @@
+# test locally with: docker compose run --rm airflow-cli dags test atd_finance_data_task_orders
+
+import os
+
+from airflow.decorators import task
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration
+
+from utils.onepassword import get_env_vars_task
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 1, 1, tz="America/Chicago"),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "execution_timeout": duration(minutes=60),  # some queries are very slow
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+# Data tracker Knack app secrets
+DATA_TRACKER_SECRETS = {
+    "KNACK_APP_ID": {
+        "opitem": "Knack AMD Data Tracker",
+        "opfield": "production.appId",
+    },
+    "KNACK_API_KEY": {
+        "opitem": "Knack AMD Data Tracker",
+        "opfield": "production.apiKey",
+    },
+}
+
+# Finance purchasing Knack app secrets
+FINANCE_PURCHASING_SECRETS = {
+    "KNACK_APP_ID": {
+        "opitem": "Knack Finance and Purchasing",
+        "opfield": "production.appId",
+    },
+    "KNACK_API_KEY": {
+        "opitem": "Knack Finance and Purchasing",
+        "opfield": "production.apiKey",
+    },
+}
+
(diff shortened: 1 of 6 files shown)


dags/atd_finance_data_subprojects.py
```diff
@@ -0,0 +1,154 @@
...
+    },
+}
+
+# Combine env vars to create one for each Knack app
+DATA_TRACKER_SECRETS.update(OTHER_SECRETS)
+FINANCE_PURCHASING_SECRETS.update(OTHER_SECRETS)
+
+with DAG(
+    dag_id="atd_finance_data_subprojects",
+    description="Gets Finance data from a database, places it in an S3 bucket, then moves it along to Knack and socrata.",
```
chiaberry, line 123: I am tempted to say the descriptions should not be the same for all the dags, but when I went to the airflow UI to see where we see the descriptions, they are always with the dag_id, so maybe that gives enough context.

chiaberry:
Is the schedule offset by 5 minutes enough, seeing that the timeout is 60 min and some of the queries take a long time?

dags/atd_finance_data_fdus.py
```diff
@@ -0,0 +1,154 @@
...
+    },
+    "SO_WEB": {
+        "opitem": "Socrata Key ID, Secret, and Token",
+        "opfield": "socrata.endpoint",
+    },
+}
+
+# Combine env vars to create one for each Knack app
+DATA_TRACKER_SECRETS.update(OTHER_SECRETS)
+FINANCE_PURCHASING_SECRETS.update(OTHER_SECRETS)
```
johnclary, line 119: nice!

johnclary (approved):
+1 Same question as Chia.

these DAGs look fantastic. thank you porting these over!!!!! 🙌 🚢

Charlie-Henry:
> Is the schedule offset by 5 minutes enough, seeing that the timeout is 60 min and some of the queries take a long time?

I was mainly just offsetting it for the first task which queries the database, so we don't bomb it all at once. Also, fortunately the full ETL only took 5 minutes to run on Prefect after my changes in atd-finance-data yesterday!

dags/atd_finance_data_master_agreements.py
```diff
@@ -0,0 +1,143 @@
...
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 1, 1, tz="America/Chicago"),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "execution_timeout": duration(minutes=60),  # some queries are very slow
```
mddilley, line 22: 🙏 - i like these comments that explain the longer timeouts

dags/atd_finance_data_subprojects.py
mddilley: i had the same thought, but I agree that the DAG names describe what data is being handled

dags/atd_finance_data_fdus.py
```diff
@@ -0,0 +1,154 @@
+# test locally with: docker compose run --rm airflow-cli dags test atd_finance_data_fdus
```
mddilley, line 1: side note that I couldn't test locally because `atddocker/atd-finance-data:production` is not multi-architecture. Not saying that this should be because it will run on atd-data03 as-is.

mddilley (approved):
I read over the code, and this looks great! Thanks for breaking these out. It made it easier for me to understand that these jobs can each stand on their own, and I think that it will be helpful when looking at run errors. 🚢 🚀

PR #165, opened 2023-09-01
Set default docker hub connection

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/13654

This PR adds a default Docker Hub connection to our Airflow stack. The connection can be referenced by `DockerOperator` tasks to mitigate rate limiting or to use a private Docker image.

## Associated repo
- atd-airflow

## Testing

It's hard to verify the connection is in place, because, per the [Airflow docs](https://airflow.apache.org/docs/apache-airflow/stable/howto/connection.html):

> Connections defined in environment variables will not show up in the Airflow UI or using airflow connections list.

So the trick to testing this is:
1. Use the Docker Hub credentials in 1password to login to hub.docker.com. 
2. From the top-right menu, go to **Account Settings** then navigate to **Security** via the left vertical nav.
3. Create a read-only access token for yourself called `<first-name>-<last-name>-local-dev`
4. Add the `DOCKER_HUB_USERNAME` and `DOCKER_HUB_TOKEN` values to your `env` file. Note that you should set the `DOCKER_HUB_USERNAME` to our Docker username, not the name of your access token.
5. Remove the knack services image from your system (or skip this step if you don't have a copy)

```
$ docker image rmi atddocker/atd-knack-services:production
```

6. Start Airflow and trigger the `atd_knack_dms ` DAG. You can observe in the task logs that the `atd-knack-services` DAG has been pulled.
7. From the access tokens page in the Docker Hub, refresh the page and locate your access token in the list of tokens. It should be near the top of the token list and have have a "Last used" date that matches the timestamp when you triggered the DAG.

Screenshot 2023-09-05 at 12 02 23 PM

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ _AIRFLOW__WEBSERVER__BASE_URL=http://localhost:8080
 OP_API_TOKEN=<Get from 1Password entry named "Connect Server: Production Access Token: API Accessible Secrets">
 OP_CONNECT=<Get from 1Password entry named "Endpoint for 1Password Connect Server API">
 OP_VAULT_ID=<Get from 1Password entry named "Vault ID of API Accessible Secrets vault">
+DOCKER_HUB_USERNAME=<Get from 1Password entry named "Docker Hub">
+DOCKER_HUB_TOKEN=<A docker hub access token assigned to specifically to you>
 ```
 
 3. Start the Docker the stack (optionlly use the `-d` flag to run containers in the background):
--- a/dags/atd_knack_dms.py
+++ b/dags/atd_knack_dms.py
@@ -79,6 +79,7 @@ with DAG(
 
     t1 = DockerOperator(
         task_id="atd_knack_dms_to_postgrest",
+        docker_conn_id="docker_default",
         image=docker_image,
         auto_remove=True,
         command=f"./atd-knack-services/services/records_to_postgrest.py -a {app_name} -c {container} {date_filter_arg}",
@@ -91,6 +92,7 @@ with DAG(
     t2 = DockerOperator(
         task_id="atd_knack_dms_to_socrata",
         image=docker_image,
+        docker_conn_id="docker_default",
         auto_remove=True,
         command=f"./atd-knack-services/services/records_to_socrata.py -a {app_name} -c {container} {date_filter_arg}",
         environment=env_vars,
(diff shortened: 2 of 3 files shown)


Charlie-Henry (approved):
worked! I guess we need to edit all of the DAGs?

```
[2023-09-05, 17:06:47 UTC] {docker.py:160} INFO - Login into Docker Registry: https://index.docker.io/v1/:80
```

chiaberry:
same question as Charlie, once this is approved, we should update all the dags to use this?

johnclary:
Yes, we will need to update all instances of `DockerOperator` to use the connection ID. I'm happy to open a separate PR for that.

FWIW, [the docs](https://airflow.apache.org/docs/apache-airflow-providers-docker/stable/connections/docker.html) seem to suggest that we can omit the conn ID argument:

> Some hooks and operators related to Docker use docker_default by default.

After repeated testing i did not find this to be the case: the connection was not used unless i explicitly added the `docker_conn_id` arg. But I've spent some time staring at the [operator source](https://github.com/apache/airflow/blob/5e6e107f733df9d0130fb0e89d975b35f08fb911/airflow/providers/docker/operators/docker.py#L319) and [hook source](https://github.com/apache/airflow/blob/5e6e107f733df9d0130fb0e89d975b35f08fb911/airflow/providers/docker/hooks/docker.py#L62) and it really seems like it should pickup the connection by default 🤷

chiaberry (approved):

mddilley (approved):
Looks great! I had a problem with my local Fernet key that this one surfaced, but I finally got it up and running. I followed the steps and saw the Docker connection working. 🚢
```
[2023-09-05, 13:45:55 CDT] {base.py:73} INFO - Using connection ID 'docker_default' for task execution.
[2023-09-05, 13:45:55 CDT] {docker.py:160} INFO - Login into Docker Registry: https://index.docker.io/v1/:80
```

johnclary:
thank you!

frankhereford:
I'm late to the party to look at this, but it's a huge 👍 🚀. I wrote:

> I am realllly grateful for this fix John - I’m sorry I didn’t get to review it in time.
If i’m reading what this PR does, it addresses what has been a real on-going problem for a long time, where we were contributing (or possibly outright causing) the COA network as a whole to get throttled by docker hub for anonymous API usage levels.

> This looks like not only a great fix for the airflow stack, but also a QOL improvement for who knows what folks on the network

PR #171, opened 2023-09-29
Create DAG to send SR activities from Knack to 311 CSR

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/11175

## Associated repo
- `atd-knack-311`

## Testing

**Steps to test:**
This branch is deployed on our production server, you can observe the run history of this DAG.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/certs/knack/README.md
+++ b/certs/knack/README.md
@@ -1,0 +1,4 @@
+# Knack self-signed SSL certificate and key stored here
+This directory is available to the docker-compose stack as a named volume. See the volumne definition in docker-compose.yaml for more details.
+
+The certificate stored here as `esb.cert` and `esb.pem` is used as a self-signed certificate for communication with the Enterprise Service Bus (ESB) for the 311 CSR integration.
--- a/dags/atd_knack_esb_311.py
+++ b/dags/atd_knack_esb_311.py
@@ -58,7 +58,7 @@ with DAG(
     dag_id=f"atd_knack_esb_311",
     description="Publishes 311 SR activities from Knack to 311 CSR via the CTM ESB",
     default_args=DEFAULT_ARGS,
-    schedule_interval="*/7 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    schedule_interval="1-59/5 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
     tags=["repo:atd-knack-311", "311", "knack", "esb"],
     catchup=False,
 ) as dag:
--- a/dags/atd_knack_esb_311.py
+++ b/dags/atd_knack_esb_311.py
@@ -59,7 +59,7 @@ with DAG(
     description="Publishes 311 SR activities from Knack to 311 CSR via the CTM ESB",
     default_args=DEFAULT_ARGS,
     schedule_interval="*/7 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
-    tags=["repo:atd-kits", "socrata", "kits"],
+    tags=["repo:atd-knack-311", "311", "knack", "esb"],
     catchup=False,
 ) as dag:
     env_vars_data_tracker = get_env_vars_task(REQUIRED_SECRETS_DATA_TRACKER)
--- a/dags/atd_knack_esb_311.py
+++ b/dags/atd_knack_esb_311.py
@@ -1,0 +1,91 @@
+import os
+
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration, now
+
+from utils.onepassword import get_env_vars_task
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+DOCKER_IMAGE ="atddocker/atd-knack-311:production"
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 1, 1, tz="America/Chicago"),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "execution_timeout": duration(minutes=5),
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+REQUIRED_SECRETS_DATA_TRACKER = {
+    "KNACK_APP_ID": {
+        "opitem": "Knack AMD Data Tracker",
+        "opfield": f"production.appId",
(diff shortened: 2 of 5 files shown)


certs/knack/README.md
johnclary, line 1: Todo: make sure these certs are in the 1password store

docker-compose.yaml
```diff
@@ -258,6 +258,13 @@ services:
...
 volumes:
+  # postgres db data persisted here
   postgres-db-volume:
+  # volume for Knack SSL certs which will be mounted to DockerOperator tasks for 311 integration
+  knack-certs:
+    driver: local
+    driver_opts:
+      type: 'none'
+      o: 'bind'
+      device: ${AIRFLOW_PROJ_DIR}/certs/knack
```
johnclary, line 271: we are loading our local copy of the knack ssl certificate directory into the airflow stack as a named volume, which can then be referenced in the `DockerOperator` constructor. 

i tried a few alternatives, such as bind-mounting this directory in the `volumes` section of the `airflow-common-env`, but this is the only way i could get this to work.

dags/atd_knack_esb_311.py
```diff
@@ -0,0 +1,102 @@
...
+        "opfield": f"production.endpoint",
+    }
+}
+
+
+with DAG(
+    dag_id=f"atd_knack_esb_311",
+    description="Publishes 311 SR activities from Knack to 311 CSR via the CTM ESB",
+    default_args=DEFAULT_ARGS,
+    schedule_interval="1-59/5 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
johnclary, line 61: it runs every 5th minute starting at minute one—just to offset the timing from other ETLs on that timing. i dunno that is helpful or not, but that's the same schedule it uses on the legacy server.

docker-compose.yaml
johnclary: another option is to create a new env var called `KNACK_CERT_DIR` so that each we could have a more standard path to them, e.g.  `/etc/ssl/certs`.

docker-compose.yaml
```diff
@@ -29,6 +29,7 @@ x-airflow-common: &airflow-common
     - ${AIRFLOW_PROJ_DIR}/plugins:/opt/airflow/plugins
     - ${AIRFLOW_PROJ_DIR}/config:/opt/airflow/config
     - ${AIRFLOW_PROJ_DIR}/airflow.cfg:/opt/airflow/airflow.cfg
+    - knack-certs:/opt/airflow/certs/knack
```
johnclary, line 32: the containers in the docker-compose stack do not need/use this volume directly. however, i have learned that docker compose will not create the volume unless at least one services uses it (this is not documented 😕 ). in our case, our airflow scheduler is going to spin up other containers (the 311 ETL) which **do** need the volume.

the work around to this is to create an external volume outside of any docker compose file, but then that external volume becomes a dependency of our stack 🤷‍♂️

dags/atd_knack_esb_311.py
```diff
@@ -0,0 +1,102 @@
...
+    tags=["repo:atd-knack-311", "311", "knack", "esb"],
+    catchup=False,
+) as dag:
+    env_vars_data_tracker = get_env_vars_task(REQUIRED_SECRETS_DATA_TRACKER)
+    env_vars_signs_markings = get_env_vars_task(REQUIRED_SECRETS_SIGNS_MARKIGNS)
+
+    # the self-signed certificate and key must be stored within the airflow project directory
+    # according to the path specified in the docker compose volumne definition
+    cert_mount =  Mount(
+        source="atd-airflow_knack-certs",
```
johnclary, line 71: this is the named volume from out docker compose file.

Charlie-Henry (approved):
V cool! No opinions on where/how to store certs but if this works let's do it! Curious how you tested this to branch in production airflow? I liked doing that with Prefect

dags/atd_knack_esb_311.py
```diff
@@ -0,0 +1,102 @@
...
+        "opitem": "Knack AMD Data Tracker",
+        "opfield": f"production.apiKey",
+    },
+    "ESB_ENDPOINT": {
+        "opitem": "CTM Enterprise Service Bus - ESB - 311 Interface",
+        "opfield": f"production.endpoint",
+    }
+}
+
+REQUIRED_SECRETS_SIGNS_MARKIGNS = {
```
johnclary: typo

johnclary:
@Charlie-Henry i pulled/checked out this branch on the production server and then restarted the docker compose stack. so yeah it's far from ideal and is currently blocking anyone else from deploying new Airflow code to prod 🙃

docker-compose.yaml
frankhereford: Good solution here John. I agree with you that that is a pretty unfortunate hole in the documentation for `docker compose`. I think you're 100% picking the best thing to include this in the airflow stack because it has the job of providing whatever is needed for ETLs to do their business. 👍

docker-compose.yaml
frankhereford: I'm a fan of the way you've done it. I think as much as we can, packing everything airflow related into something inside the repo is the way to go. If we wanted to more tightly integrate with the OS and its environment, we would probably be best served just running the airflow services as non-containerized processes on the machine, and that's just a big departure from everything we've done so far.

frankhereford (approved):
Given your recent work with this on `atd-data03`, I have not spun this up locally. I've gone over the code and your solutions, and it all looks ship shape to me. Let me know if you want me to do any additional testing, and I'd be more than happy to. 🚢

docker-compose.yaml
mddilley: +1 - adding a certs folder makes a lot of sense in case we see this need come up again. Seeing the mount param in the `DockerOperator` makes it clear to trace what is going to me.

mddilley (approved):
Checked the history in the dashboard, and it looks like this DAG is cruising along. 🚘 Thanks for figuring out the config here - it looks like it was a tricky one! 🚀

certs/knack/README.md
```diff
@@ -0,0 +1,4 @@
+# Knack self-signed SSL certificate and key stored here
+This directory is available to the docker-compose stack as a named volume. See the volumne definition in docker-compose.yaml for more details.
```
chiaberry, line 2: another typo in "volume"

docker-compose.yaml
johnclary: thanks for the feedback, y'all!!!

PR #184, opened 2023-11-01
Backup airflow DB to S3 via dag

### LINKED ISSUE #12347: Create ETL for backing up Airflow DB on atd-data03
Now that Airflow v2 is running on atd-data03, we should create an ETL process that connects to the Postgres DB that is running in the Airflow Docker stack, pg_dump and zip it, and upload to S3. This would help us recover metadata in the instance of a failure.

Todo:
- Inspect Airflow DB to understand if/what needs to be backed up
- Determine next steps

frankhereford:
## Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/12347.

## Associated repo

Airflow

## Testing

This PR requires a new package to be installed in the airflow images, so to test this locally, modify your docker-compose.yaml file to build from `.` and build some local images for testing. This will happen via CI when merged into production. I think, also, because the DAG is entirely self contained and there's no docker reliance or anything like that -- the eyeball test may be appropriate here. 

**Steps to test:**

Normal routine - spin up local airflow and give it a whirl. 

## S3 Lifecycle Rules

I've added this lifecycle rule to the `atd-airflow` backup bucket. After 180 days, all files in the bucket are auto-deleted. Backups are only going to last 180 days. 

Also, if a backup is overwritten, we'll keep all versions until the 180 days, but after a month, we'll only keep up to 5. This is a much less important part of the rule.

image

---

#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/airflow-database-backup.py
+++ b/dags/airflow-database-backup.py
@@ -1,0 +1,63 @@
+import os
+from pendulum import datetime, duration
+
+from airflow.decorators import task
+from airflow.models import DAG
+from airflow.operators.bash_operator import BashOperator
+
+from datetime import datetime as vanilla_datetime
+
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+default_args = {
+    "owner": "airflow",
+    "description": "Backup, compress and store airflow pg_dump in S3",
+    "depends_on_past": False,
+    "start_date": datetime(2019, 1, 1, tz="America/Chicago"),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "execution_timeout": duration(minutes=30),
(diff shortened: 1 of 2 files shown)


dags/airflow-database-backup.py
```diff
@@ -0,0 +1,63 @@
+import os
+from pendulum import datetime, duration
+
+from airflow.decorators import task
+from airflow.models import DAG
+from airflow.operators.bash_operator import BashOperator
+
+from datetime import datetime as vanilla_datetime
```
mddilley, line 8: oh yep, I forgot that we import `datetime` from pendulum elsewhere 👍

mddilley (approved):
Looks great, thanks!. 🙏 I tested by building locally, and I saw my backup make it into the Airflow bucket which I will now delete. ✅ 🙌
Screenshot 2023-11-13 at 3 42 02 PM

chiaberry (approved):

PR #188, opened 2023-12-13
Add review time query task

Charlie-Henry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14143#issuecomment-1815273461

## Associated repo
https://github.com/cityofaustin/atd-executive-dashboard

## Testing

**Steps to test:**
Feel free to trigger this DAG and check if we get all ✅ 's

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_executive_dashboard_row_weekly_summary.py
+++ b/dags/atd_executive_dashboard_row_weekly_summary.py
@@ -166,4 +166,17 @@ with DAG(
         retry_delay=duration(seconds=60),
     )
 
-    t1 >> t2 >> t3 >> t4 >> t5
+    t6 = DockerOperator(
+        task_id="amanda_review_time",
+        image=docker_image,
+        auto_remove=True,
+        command=f"python AMANDA/amanda_to_s3.py --query review_time",
+        environment=env_vars,
+        tty=True,
+        force_pull=False,
+        mount_tmp_dir=False,
+        retries=3,
+        retry_delay=duration(seconds=60),
+    )
+
+    t1 >> t2 >> t3 >> t4 >> t5 >> t6


dags/atd_executive_dashboard_row_weekly_summary.py
```diff
@@ -166,4 +166,17 @@
         retry_delay=duration(seconds=60),
     )
 
-    t1 >> t2 >> t3 >> t4 >> t5
+    t6 = DockerOperator(
+        task_id="amanda_review_time",
+        image=docker_image,
+        auto_remove=True,
+        command=f"python AMANDA/amanda_to_s3.py --query review_time",
```
mddilley, line 173: Just out of curiosity, will this create a new csv called review_time.csv in the `atd-exec-dashboard`. I peeked the bucket and wanted to check my understanding of the output of the script. 🙏

mddilley (approved):
Looks great! I did a quick comparison to the other tasks, and all looks well to capture the new query output. 🚀

dags/atd_executive_dashboard_row_weekly_summary.py
Charlie-Henry: yes that should happen 🤞!

dags/atd_executive_dashboard_row_weekly_summary.py
```diff
@@ -166,4 +166,17 @@
...
+        environment=env_vars,
+        tty=True,
+        force_pull=False,
+        mount_tmp_dir=False,
+        retries=3,
+        retry_delay=duration(seconds=60),
+    )
+
+    t1 >> t2 >> t3 >> t4 >> t5 >> t6
```
johnclary, line 182: this is not anything you need to address, but i was just thinking about what the preferred flow is for situations where you want every task to run, and one task failing should not block another task. e.g. in this case, maybe we still want to run the Smartsheet ETL if the AMANDA query fails.
 
TIL you can pass a [trigger rule](https://airflow.apache.org/docs/apache-airflow/1.10.9/concepts.html#trigger-rules) to each task that sets the condition under which it will run, so I guess setting `all_done` would ensure T2 runs regardless of what happens in T1.

johnclary (approved):
OK! 🚢 🚢 🚢 🚢

PR #191, opened 2023-12-20
Automate population of CR3 metadata in VZDB

frankhereford:
## Associated issues

This work is a continuation of the work here: https://github.com/cityofaustin/atd-data-tech/issues/14899. 

## Associated repo

The Vision Zero Stack at `atd-vz-data`

## Testing

1) Spin up the VZDB and populate it with a replication of production's data.
2) Spin up the airflow stack and trigger the DAG `vz-populate-cr3-metadata`
3) Before, after, and during, try this command to quantify the number of un-metadata bearing crashes there:

```sql
select count(*)
from atd_txdot_crashes 
where cr3_file_metadata is null
```

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/vz_populate_cr3_metadata.py
+++ b/dags/vz_populate_cr3_metadata.py
@@ -1,5 +1,5 @@
 import os
-from pendulum import datetime
+from pendulum import datetime, duration
 from airflow.decorators import dag, task
 from airflow.operators.docker_operator import DockerOperator
 from utils.slack_operator import task_fail_slack_alert
@@ -34,6 +34,14 @@ REQUIRED_SECRETS = {
         "opitem": "Vision Zero graphql-engine Endpoints",
         "opfield": f"{DEPLOYMENT_ENVIRONMENT}.Admin Key",
     },
+    "AWS_ACCESS_KEY_ID": {
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.AWS Access key",
+    },
+    "AWS_SECRET_ACCESS_KEY": {
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.AWS Secret key",
+    },
 }
 
 
@@ -55,7 +63,11 @@ def populate_cr3_metadata():
     def get_env_vars():
         from utils.onepassword import load_dict
 
-        return load_dict(REQUIRED_SECRETS)
+        answers = load_dict(REQUIRED_SECRETS)
+        print(answers)
+        return answers
+
+        # return load_dict(REQUIRED_SECRETS)
 
     env_vars = get_env_vars()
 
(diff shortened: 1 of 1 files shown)


dags/vz_populate_cr3_metadata.py
```diff
@@ -0,0 +1,81 @@
...
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.AWS Secret key",
+    },
+}
+
+
+@dag(
+    dag_id="vz-populate-cr3-metadata",
+    description="A DAG which populates CR3 metadata in the VZDB",
+    schedule_interval="*/5 8-10 * * *"
```
mddilley, line 51: It took me a second to realize that this schedule works together with the script processing 100 records per run. I can never remember why this is scheduled this way, and I have asked before so maybe a comment about it would be good (if we know why).

dags/vz_populate_cr3_metadata.py
```diff
@@ -0,0 +1,81 @@
...
+    "AWS_SECRET_ACCESS_KEY": {
+        "opitem": "Vision Zero CRIS Import",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.AWS Secret key",
+    },
+}
+
+
+@dag(
+    dag_id="vz-populate-cr3-metadata",
+    description="A DAG which populates CR3 metadata in the VZDB",
```
mddilley, line 50: I wanted to link to [my comment from the corresponding PR](https://github.com/cityofaustin/atd-vz-data/pull/1328#discussion_r1432895861) because I'm still wondering how we can avoid a situation in the future where we forget how important this metadata is to the diagram and narrative extraction. I'm 100% talking about a future enhancement. This works great to get us back up and running. 🙌 🙏

mddilley (approved):
Looks great, works great, and gets us back in action - thanks for writing this DAG! 🚀

dags/vz_populate_cr3_metadata.py
chiaberry: I dont remember why it is scheduled this way either. maybe we dont have to do it this way anymore? (future refinement)

chiaberry (approved):
I agree with Mike's comments, would be nice to understand why we do it in chunks of 100 and have the ETL schedule them

frankhereford:
@chiaberry @mddilley Thanks y'all for your comments and reviews!

The scheduling is chosen here isn't the clearest, but FWIW, here's the reasoning as I understand it. The script itself that I'm bringing over has a default value of 100 records per execution in it, and it didn't look like that was being modified, so I chose not to increase the complexity of the change, and just focus on bringing it over as-is. Since we're going to be processing something like 100-500 crashes per day, we can expect it to take, say, 5 executions to get caught up. Given that relatively low number, I went with the 5-minutely execution, but did limit that to 3 hours in the morning, after we import crashes at 7AM so it doesn't keep pinging the DB with that query all day or firing up a container on the airflow machine when we could be confident that there wasn't work to do. I figured 36 executions per day, when we reasonably can expect 5 to be used, was good coverage. 

As to the original 100 records at a time choice -- I think that's a decision that was made a looong time ago, and it was a decision made with the intent of not loading up the airflow machine with a ton of work all at once when we could spread it out over time. I think it would be marginally better to do them all at once and run the script just once daily, but I also don't think it matters all that much either, so I'm with y'all - completely good with kicking the can down the road for a future enhancement when we want to pick that back up.

PR #192, opened 2023-12-21
Backing up active ROW permits log

Charlie-Henry:
Remembered this [dataset](https://data.austintexas.gov/Transportation-and-Mobility/ATD-Right-of-Way-Active-Permits/hyc6-zz9w/about_data) is not being backed up which just has running log of snapshots in time of active ROW permits retrieved from AMANDA.
## Testing

**Steps to test:**
Trigger this DAG and check for `INFO - backup_active_permits ran successfully!` 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_executive_dashboard_row_active_permits_logging.py
+++ b/dags/atd_executive_dashboard_row_active_permits_logging.py
@@ -140,6 +140,7 @@ with DAG(
     t1 = DockerOperator(
         task_id="active_permits_logging",
         image=docker_image,
+        docker_conn_id="docker_default",
         auto_remove=True,
         command=f"python active_permits_logging.py",
         environment=env_vars,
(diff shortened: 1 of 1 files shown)


dags/atd_executive_dashboard_row_active_permits_logging.py
```diff
@@ -114,4 +150,16 @@
         retry_delay=duration(seconds=60),
     )
 
-    t1
+    t2 = DockerOperator(
+        task_id="backup_active_permits",
+        image=docker_image_2,
+        auto_remove=True,
```
chiaberry, line 158: We have an open issue to do this, but I've been adding the default docker hub connection to my new dags

https://github.com/cityofaustin/atd-airflow/pull/165/files#diff-6c336b555e2faf578eda218ff84472fe28d1c7787414c68a8168e0193a13eac5R95

chiaberry (approved):
👍🏼

PR #196, opened 2024-01-03
Adjust timing for VZ OCR ETL

frankhereford:
## Associated issues

This PR is tangentially related to https://github.com/cityofaustin/atd-data-tech/issues/14899. It concerns tuning the timing around the ETL so that it is able to complete full batches of OCR and diagram extraction per execution. 

## Associated repo

`atd-vz-data`

## Testing

* Spin up local airflow and see that the script compiles. You won't see the schedule in local airflow, but please take a close look at it in the diff to see if it correctly schedules the DAG every 20 minutes, all day. 
* Look closely that the DAG is set to not timeout until 20 minutes into a task's execution.

### Associated data

image

This is data03 trucking along overnight under the timing regime set forth in this PR as it back-processes a large number of CR3s.

---

#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/vz_cr3_extract_ocr_narrative.py
+++ b/dags/vz_cr3_extract_ocr_narrative.py
@@ -17,7 +17,7 @@ default_args = {
     "email_on_failure": False,
     "email_on_retry": False,
     "retries": 0,
-    "execution_timeout": duration(minutes=5),
+    "execution_timeout": duration(minutes=20),
     "on_failure_callback": task_fail_slack_alert,
 }
 
--- a/dags/vz_cr3_extract_ocr_narrative.py
+++ b/dags/vz_cr3_extract_ocr_narrative.py
@@ -40,16 +40,15 @@ REQUIRED_SECRETS = {
     },
 }
 
+# fmt: off
 with DAG(
     dag_id=f"vz_cr3_ocr_narrative_extract_{DEPLOYMENT_ENVIRONMENT}",
     default_args=default_args,
-    # Every 5 minutes, at 8A, 9A, and 10A
-    schedule_interval="*/20 * * * *"
-    if DEPLOYMENT_ENVIRONMENT == "production"
-    else None,
+    schedule_interval="*/20 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
     tags=["repo:atd-vz-data", "vision-zero"],
     catchup=False,
 ) as dag:
+# fmt: on
 
     @task(
         task_id="get_env_vars",
--- a/dags/vz_cr3_extract_ocr_narrative.py
+++ b/dags/vz_cr3_extract_ocr_narrative.py
@@ -29,7 +29,7 @@ REQUIRED_SECRETS = {
     "HASURA_ADMIN_KEY": {
         "opitem": "Vision Zero CRIS Import",
         "opfield": "production.GraphQL Endpoint key",
-    }, 
+    },
     "AWS_ACCESS_KEY_ID": {
         "opitem": "Vision Zero CRIS Import",
         "opfield": "production.AWS Access key",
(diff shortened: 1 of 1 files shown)


dags/vz_cr3_extract_ocr_narrative.py
frankhereford, line 1: Most of this is just a linting with black -- I'll point out the two actual changes.

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -17,7 +17,7 @@
     "email_on_failure": False,
     "email_on_retry": False,
     "retries": 0,
-    "execution_timeout": duration(minutes=5),
+    "execution_timeout": duration(minutes=20),
```
frankhereford, line 20: This is configuring the OCR task to timeout at 20 minutes. I've tinkered with the timing and under its current configuration on this machine, it processes a batch in 15 minute ± 2 minutes.

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -29,35 +29,39 @@
...
 }
 
 with DAG(
     dag_id=f"vz_cr3_ocr_narrative_extract_{DEPLOYMENT_ENVIRONMENT}",
     default_args=default_args,
     # Every 5 minutes, at 8A, 9A, and 10A
-    schedule_interval="*/5 8-10 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    schedule_interval="*/20 * * * *"
+    if DEPLOYMENT_ENVIRONMENT == "production"
+    else None,
```
frankhereford, line 49: Adjust the schedule from 5-minutely for 3 hours per day (36 executions), to 20-minutely all day (72 executions).

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -29,35 +29,38 @@
...
+    },
 }
 
+# fmt: off
 with DAG(
     dag_id=f"vz_cr3_ocr_narrative_extract_{DEPLOYMENT_ENVIRONMENT}",
     default_args=default_args,
-    # Every 5 minutes, at 8A, 9A, and 10A
-    schedule_interval="*/5 8-10 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    schedule_interval="*/20 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
frankhereford, line 47: Adjust the schedule from 5-minutely for 3 hours per day (36 executions), to 20-minutely all day (72 executions).

Charlie-Henry (approved):

dags/vz_cr3_extract_ocr_narrative.py
```diff
@@ -29,35 +29,38 @@
...
     tags=["repo:atd-vz-data", "vision-zero"],
     catchup=False,
 ) as dag:
+# fmt: on
```
chiaberry, line 51: what does this mean?

mddilley (approved):
Checked the cron expression, and this all looks great to me. Thanks for updating! 🙏

dags/vz_cr3_extract_ocr_narrative.py
chiaberry: and then it will go back to 3 hours a day after the backfill is complete? or it will always be all day?

dags/vz_cr3_extract_ocr_narrative.py
frankhereford: It's a formatting directive to `black` when to turn off and on. `black` wants to break the schedule line up over 3 lines, and I find it much easier to understand with each line being one key/value pair. I really do like black's linting 99.999% of the time, but i wanted to override it here.

chiaberry: ah! TIL, thank you!

dags/vz_cr3_extract_ocr_narrative.py
frankhereford: The intent is to leave it as 24h going forward. Since it runs less frequently, there is proportionally less impact and under a busy day or a catch up day, this gives it what feels like a reasonable amount of execution time if needed.

chiaberry (approved):
🚢

PR #197, opened 2024-01-05
Task for new AMANDA query

Charlie-Henry:
## Associated issues
Adding new task for helping to track review time for ROW 
https://github.com/cityofaustin/atd-data-tech/issues/14143

## Associated repo
https://github.com/cityofaustin/atd-executive-dashboard/blob/main/AMANDA/amanda_to_s3.py

## Testing
**Steps to test:**
Trigger this DAG and check it completes ✅, if you're running this on a mac you will probably get the same error I'm getting `Cannot locate a 64-bit Oracle Client library` 😑 so I'm unsure how to test this without yolo'ing.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_executive_dashboard_row_weekly_summary.py
+++ b/dags/atd_executive_dashboard_row_weekly_summary.py
@@ -1,3 +1,5 @@
+# test locally with: docker compose run --rm airflow-cli dags test atd_executive_dashboard_row_weekly_summary
+
 import os
 
 from airflow.models import DAG
@@ -17,7 +19,7 @@ DEFAULT_ARGS = {
     "email_on_failure": False,
     "email_on_retry": False,
     "retries": 0,
-    "execution_timeout": duration(minutes=5),
+    "execution_timeout": duration(minutes=30),
     "on_failure_callback": task_fail_slack_alert,
 }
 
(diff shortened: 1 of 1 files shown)


dags/atd_executive_dashboard_row_weekly_summary.py
```diff
@@ -17,7 +19,7 @@
     "email_on_failure": False,
     "email_on_retry": False,
     "retries": 0,
-    "execution_timeout": duration(minutes=5),
+    "execution_timeout": duration(minutes=30),
```
Charlie-Henry, line 22: unlikely to take this long but want to be safe

mddilley (approved):
yolo! 🚀 The new task looks great. 🚢

dags/atd_executive_dashboard_row_weekly_summary.py
```diff
@@ -179,4 +181,17 @@
...
+        command=f"python AMANDA/amanda_to_s3.py --query ex_permits_issued",
+        environment=env_vars,
+        tty=True,
+        force_pull=False,
+        mount_tmp_dir=False,
+        retries=3,
+        retry_delay=duration(seconds=60),
+    )
+
+    t1 >> t2 >> t3 >> t4 >> t5 >> t6 >> t7
```
johnclary, line 197: @Charlie-Henry i can't remember if we've talked about this, or if it is even desired in this case, but it is possible to use [trigger rules](https://airflow.apache.org/docs/apache-airflow/1.10.10/concepts.html#trigger-rules) to enable downstream tasks to run regardless of if an earlier task fails.

johnclary (commented):
👍

dags/atd_executive_dashboard_row_weekly_summary.py
Charlie-Henry: ah yes you did comment this before, good idea. I'll see if I can get something to work

PR #198, opened 2024-01-08
ROW Queries changing trigger rule

Charlie-Henry:
## Associated issues
@johnclary suggested to change the [trigger rule](https://airflow.apache.org/docs/apache-airflow/1.10.10/concepts.html#trigger-rules) for these DAGs so they run regardless if the upstream tasks were successful or not. This could handle the case that we could have one query fail to run and we will still want the others to try.

I've changed the trigger rule from `all_success`(default) to `all_done` which triggers the DAG if all parents are done with their execution. That way we still trigger the DAG if things fail but also should not bomb the AMANDA database. 

## Associated repo
https://github.com/cityofaustin/atd-executive-dashboard

## Testing

**Steps to test:**
Easy way to test this is to not connect to the VPN and try to trigger this DAG. The first `get_env_vars` task should succeed if your 1pass secrets are set locally. The `amanda_` and `ex_permits_issued` tasks should fail but after the three retries are exhausted, the DAG should continue on to the remaining tasks.

I'm not sure how best to test the slack notifications, but I assume it will notify each time a task comes up 🟥 (failed). This could create a lot of noise but I'm not sure if there's a way around that? [Callbacks docs](https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/logging-monitoring/callbacks.html)

Right-most column is what I get when applying these trigger rules:
Screenshot 2024-01-08 at 11 17 15 AM

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_executive_dashboard_row_weekly_summary.py
+++ b/dags/atd_executive_dashboard_row_weekly_summary.py
@@ -114,6 +114,7 @@ with DAG(
         mount_tmp_dir=False,
         retries=3,
         retry_delay=duration(seconds=60),
+        trigger_rule="all_done",
     )
 
     t2 = DockerOperator(
@@ -127,6 +128,7 @@ with DAG(
         mount_tmp_dir=False,
         retries=3,
         retry_delay=duration(seconds=60),
+        trigger_rule="all_done",
     )
 
     t3 = DockerOperator(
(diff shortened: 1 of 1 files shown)


chiaberry (approved):
🚢 🚣🏼 
tested 🟩 🟥 🟩

johnclary (approved):
Nice!! It's cool to see this in practice. Will be thinking of other opportunities to use this 🙌

PR #199, opened 2024-01-10
Adjust restart policy and nix VSCode settings file

frankhereford:
## Associated issues

_none_

### Restart policy

This PR is to start a discussion mainly around the idea of changing our restart policies for airflow images. The idea is to:

1) Maintain the restart behavior for airflow on our production machine, which is to start anytime the docker service starts.
1) Change the restart behavior of having our local stacks, so that airflow doesn't fire up anytime we reboot our dev computers or restart our docker daemon. 

I'm proposing we switch to `restart: unless-stopped` from `restart: always`.

`restart: unless-stopped` means to restart unless someone uses `docker compose stop` on the compose stack, which we'll never do intentionally on our production machine. So, on data03, `unless-stopped` is effectively `always`.

However, it plays out a little differently on our local machines because we don't always want to be running the stack and we might restart our computers from time to time. When we're done working on the airflow local stack, we do a `docker compose stop` to stop our local stack, and it will stay down, but with the `restart: always` policy in effect, it is configured to restart when the docker service daemon restarts, which means airflow fires up automatically on reboot, which is not great because it consumes resources on dev machines (cpu/battery/ram etc).

I think if `restart` was replaced with a over-descriptive key name, it would be something like `should_container_start_on_docker_service_start`.

### VSCode configurations

I want to set some VSCode configurations that I don't think anyone else is going to be interested in, such as a per-repo color scheme to help me differenciate my various vscode windows. There's nothing in the configuration that isn't already left-over frank config stuff, so I don't think this is a deal at all.

## Associated repo

airflow

## Testing

Really, the main thing we're looking for here is a non-event. 

* Start the local stack, 
* stop the local stack, 
* and reboot your computer. 
* When it comes back up, do a `docker ps`. 
* Did airflow start on boot? Yes: ❌; No: ✅

---
#### Ship list
- [ ] Code reviewed 


docker-compose.yaml
```diff
@@ -266,6 +266,6 @@ volumes:
   knack-certs:
     driver: local
     driver_opts:
-      type: 'none'
-      o: 'bind'
+      type: "none"
+      o: "bind"
```
frankhereford, line 270: YAML linting

johnclary (approved):
I tested this by making sure the docker stack came back up after i force quit Docker. It did, which is what we wan to happen on prod.

👍 Thanks!

mddilley (changes requested):
Thanks, Frank! I read through the changes and your description. It all makes sense to me! 🚢 🚀

_Edit: Sorry - no changes requested. Hit the wrong button_ 🙃

mddilley (approved):
It works great if you select the correct radio - ship ship ship! ✅

chiaberry (approved):

PR #201, opened 2024-01-11
Update traffic signals frequency to every 5 min

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/15211

the mmc is going to be able to denote if a signal has gone DARK
so now this must be updated more frequently

## Testing

**Steps to test:**


Charlie-Henry (approved):
let's do it! 🧊 🚥

mddilley (approved):
⏳ ✅ 🚢

PR #203, opened 2024-01-22
Sync Moped and Data Tracker projects

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/8536

This PR adds a DAG to run the ETL that adds and updates Moped project data into the Knack Data Tracker `projects` table.

## Associated repo
https://github.com/cityofaustin/atd-moped/pull/1250

## Testing

**Steps to test:**
1. Start the local Airflow stack
2. Find a [staging Moped](https://moped.austinmobility.io/moped/) project and find its matching row in the [projects table of the test Data Tracker app](https://builder.knack.com/atd/test--austin-transportation-data-tracker--30-aug-2022/records/objects/object_201?builder_table_sort=field_3858%7Cdesc) by the Moped project ID. If it doesn't have a matching row in Knack, then this DAG will create it. Note that the project title in Moped matches the **Name** column in the Knack table.
3. Update the title of the Moped project through the staging Moped editor
4. Run the DAG added in this PR
5. You should see the DAG log that it updated at least one record (or created one if your record isn't synced yet). Find your row again in the Knack table, and it should be updated with your new title.
6. All of the staging projects are already synced to the test Knack app linked above as of creation of this PR. You can check out the row in the projects table and compare to the projects in staging Moped.

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/dags/atd_moped_data_tracker_sync.py
+++ b/dags/atd_moped_data_tracker_sync.py
@@ -32,11 +32,11 @@ REQUIRED_SECRETS = {
         "opitem": "Moped Hasura Admin",
         "opfield": f"{DEPLOYMENT_ENVIRONMENT}.Admin Secret",
     },
-    "KNACK_APP_ID": {
+    "KNACK_DATA_TRACKER_APP_ID": {
         "opitem": "Knack AMD Data Tracker",
         "opfield": f"{DEPLOYMENT_ENVIRONMENT}.appId",
     },
-    "KNACK_API_KEY": {
+    "KNACK_DATA_TRACKER_API_KEY": {
         "opitem": "Knack AMD Data Tracker",
         "opfield": f"{DEPLOYMENT_ENVIRONMENT}.apiKey",
     },
@@ -62,7 +62,7 @@ with DAG(
         task_id="data_tracker_sync",
         image=docker_image,
         auto_remove=True,
-        command=f"python data_tracker_sync.py --start {date_filter_arg}",
+        command=f"python data_tracker_sync.py {date_filter_arg}",
         environment=env_vars,
         tty=True,
         force_pull=True,
--- a/dags/atd_moped_data_tracker_sync.py
+++ b/dags/atd_moped_data_tracker_sync.py
@@ -44,7 +44,7 @@ with DAG(
     tags=["repo:atd-moped", "moped", "agol"],
     catchup=False,
 ) as dag:
-    docker_image = "atddocker/atd-moped-etl-arcgis:production"
+    docker_image = "atddocker/atd-moped-etl-data-tracker-sync:production"
 
     date_filter_arg = get_date_filter_arg()
 
@@ -54,7 +54,7 @@ with DAG(
         task_id="data_tracker_sync",
         image=docker_image,
         auto_remove=True,
-        command="python data_tracker_sync.py",
+        command=f"python data_tracker_sync.py --start {date_filter_arg}",
         environment=env_vars,
         tty=True,
         force_pull=True,
--- a/dags/atd_moped_data_tracker_sync.py
+++ b/dags/atd_moped_data_tracker_sync.py
@@ -45,7 +45,7 @@ REQUIRED_SECRETS = {
 
 with DAG(
     dag_id="atd_moped_data_tracker_sync",
-    description="sync Moped projects data to Knack Data Tracker projects table",
+    description="sync Moped project data to Knack Data Tracker projects table",
     default_args=DEFAULT_ARGS,
     schedule_interval="0 6 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
     dagrun_timeout=duration(minutes=30),
@@ -58,14 +58,11 @@ with DAG(
 
     env_vars = get_env_vars_task(REQUIRED_SECRETS)
 
-    # Run the script in test mode if we're developing locally
-    test_flag = "--test" if DEPLOYMENT_ENVIRONMENT == "development" else ""
-
     t1 = DockerOperator(
         task_id="data_tracker_sync",
         image=docker_image,
         auto_remove=True,
-        command=f"python data_tracker_sync.py --start {date_filter_arg} {test_flag}",
+        command=f"python data_tracker_sync.py --start {date_filter_arg}",
         environment=env_vars,
         tty=True,
         force_pull=True,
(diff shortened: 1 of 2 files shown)


dags/atd_moped_data_tracker_sync.py
```diff
@@ -0,0 +1,72 @@
...
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.apiKey",
+    },
+}
+
+
+with DAG(
+    dag_id="atd_moped_data_tracker_sync",
+    description="sync Moped project data to Knack Data Tracker projects table",
+    default_args=DEFAULT_ARGS,
+    schedule_interval="0 6 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
mddilley: This schedule is TBD so I'll adjust as needed.

johnclary: Considering this going to be a really light load, i was thinking we'd run it hourly.

It does seem like overkill, since 99% of the changes to projects are not going to be the pieces of data that we sync (project id, title, signal components). I will check with AMD folks 🙏

dags/atd_moped_data_tracker_sync.py
```diff
@@ -0,0 +1,72 @@
...
+
+    t1 = DockerOperator(
+        task_id="data_tracker_sync",
+        image=docker_image,
+        auto_remove=True,
+        command=f"python data_tracker_sync.py {date_filter_arg}",
+        environment=env_vars,
+        tty=True,
+        force_pull=True,
+        mount_tmp_dir=False,
```
johnclary, line 70: it would be good to add this `docker_conn_id` [arg](https://github.com/cityofaustin/atd-airflow/blob/7d4b1ee6d46632b336ee9b7dbb12ba448bff1e73/dags/atd_executive_dashboard_row_weekly_summary.py#L146) as well.

chiaberry (approved):
I at first couldnt get my projects to update, but then I changed the date flag to be earlier, and my changes showed up. I think I was having issues with the timezone mismatching with my local airflow and what was coming from staging.

dags/atd_moped_data_tracker_sync.py
mddilley: Just made this update and we can adjust if needed in the future. 🙏

dags/atd_moped_data_tracker_sync.py
```diff
@@ -0,0 +1,72 @@
...
+
+    date_filter_arg = get_date_filter_arg()
+
+    env_vars = get_env_vars_task(REQUIRED_SECRETS)
+
+    t1 = DockerOperator(
+        task_id="data_tracker_sync",
+        image=docker_image,
+        auto_remove=True,
+        command=f"python data_tracker_sync.py {date_filter_arg}",
```
frankhereford, line 66: Looking at the code this invokes, it looks like this may need a `-d` or `--date` to comply to the format `argparse` wants. Here's the code in moped: https://github.com/cityofaustin/atd-moped/blob/main/moped-etl/data-tracker-sync/data_tracker_sync.py#L232-L238.

frankhereford (changes requested):
## Review

Thank you @mddilley . Also, thanks for the really clear instructions! They are top-shelf.

I think this needs to have a `--date` added to the invocation line in the DAG. And even if it got merged like this right now, it would still work, it'd just do a whole sync each time, I think. I'll request changes, but please override it if that's helpful in regards to the release because this fails to safe.


## Observation

Like @chiaberry, I was having trouble getting a sync to go over until I adjusted the look-back time to something presumably old enough to find a record to sync. Once that occurred, it was working great and continued to work great even as I adjusted back the look-back interval to what's in the PR.

I guess, put another way, I think we may have a one-time edge case where we'll have to adjust that interval when we deploy to production to look-back time amount maybe 10 or maybe more days, do a sync and then undo that change. At that point, it'll be smooth sailing. I looked quickly at the code in the Airflow that actually interacts with that date, but I'm not positive the ETL is using the date right now at all.. so 🤷 😵‍💫 I wouldn't worry about this edge case right now, I can't reproduce it.

dags/atd_moped_data_tracker_sync.py
mddilley: @frankhereford Thanks for the thorough review! I should have mentioned in the PR description that I wrote the Moped script to use [this existing helper](https://github.com/cityofaustin/atd-airflow/blob/production/dags/utils/knack.py) that returns the `-d` flag along with the last run date (or today's date if there hasn't been a run yet).

~I do agree that there is an edge case around when the ETL sees updates in Moped, but I think that the hourly schedule should be in our favor for catching them on the next run. I'm going to re-request review and see what you think about that.~ I just put together what you are saying about not catching updates. Since we are looking back only up until the last successful run, we will continue to miss updates in the future if they have already been missed. I'll take some time this morning to think through this. 🤔

johnclary: thanks for all the attention to this date-sync business! 

the first time the production-deployed DAG runs, it will run with the fallback date arg equal to [`now()`](https://github.com/cityofaustin/atd-airflow/blob/9942821b9e0eeada005db65f8040b51397a689e9/dags/utils/knack.py#L23). after that, it will run with the `prev_start_date_success` context variable.

so, once this is a deployed to prod, i think we want to manually trigger the DAG to establish the `prev_start_date_success`, and then someone should manually run the ETL locally without a `--date` arg, so as to bring the whole of Moped into sync with Data Tracker. and then we should be covered going forward. does that sound right?

as an aside—yeesh, [the docstring](https://github.com/cityofaustin/atd-airflow/blob/9942821b9e0eeada005db65f8040b51397a689e9/dags/utils/knack.py#L20) in this helper function is wrong. it returns `-d <timestamp>`  not `<timestamp>` :/

mddilley: This has been really helpful to think through, and I think the records fall into three categories (so far):
- Moped projects not synced to Knack (covered by script)
- Moped projects synced to Knack and updated since the last successful run (covered)
- Moped project already synced to Knack and updated before this script existed and before the look-back time of the very first run (not covered)

So, I agree that we need to run this once with a look-back time far back enough to cover that third category. Something like what [this helper](https://github.com/cityofaustin/atd-airflow/blob/production/dags/utils/time.py) does.

mddilley: jinx @johnclary 😄

mddilley:
@frankhereford @johnclary That all sounds right to me. Thank y'all for the help on this! 🙏 I'll make the suggested changes, and I'll add the manual ETL run to catch all previous update to the launch instructions and assign it to myself.

dags/atd_moped_data_tracker_sync.py
mddilley: added ✅ 🙏

dags/atd_moped_data_tracker_sync.py
johnclary: thanks Mike!!

frankhereford (approved):

PR #212, opened 2024-02-27
Return Traffic Incidents schedule to every 5 mins

Charlie-Henry:
We were noticing that the traffic incidents dataset would return nothing at unusual times (like rush hour) then return several incidents only minutes later. 

Turns out, we were hitting the DB at exactly the wrong time (sometimes). According to the logs on airflow, at  :09, :24, :39, :54 past the hour we were getting zero rows from the traffic incidents CTM oracle DB. I tested this and found that at these times, the rows are deleted then replaced a few seconds later. I guess this is how they remove stale incidents from this view. Moving the schedule to every 5 minutes should allow us to still get every incident but also avoid these times where get zero rows back from the DB. 

## Testing

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_traffic_incident_reports.py
+++ b/dags/atd_traffic_incident_reports.py
@@ -77,7 +77,7 @@ with DAG(
     dag_id="atd_traffic_incident_reports",
     description="wrapper etl for atd-traffic-incident-reports docker image connects to oracle db and updates postrgrest and socrata with incidents",
     default_args=DEFAULT_ARGS,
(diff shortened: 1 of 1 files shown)


chiaberry (approved):

mddilley (approved):
Thanks for the background on this change 🙏 🚢

frankhereford (approved):
Wow, what are the odds. Feels like they should be wrapping their delete & repopulate in a transaction so it's atomic, but oh well. Great 🕵️‍♂️ work.

PR #214, opened 2024-03-05
do a full replace of School Zone Beacon Zone dataset

chiaberry:
Switching the dag to a full replace, there are 174 records total in this view

https://github.com/cityofaustin/atd-data-tech/issues/15943


Charlie-Henry (approved):
🚢  zones<sup>2</sup> 🚢

PR #215, opened 2024-03-05
purchase request copier

Charlie-Henry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/16060

## Associated repo

## Testing

**Steps to test:**
Will likely be a non-op if you run this without creating a copy request by following the steps [here](https://github.com/cityofaustin/atd-knack-services/pull/133)

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_knack_purchase_request_copier.py
+++ b/dags/atd_knack_purchase_request_copier.py
@@ -1,0 +1,62 @@
+import os
+
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration
+
+from utils.onepassword import get_env_vars_task
+from utils.knack import get_date_filter_arg
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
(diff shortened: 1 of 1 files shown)


chiaberry (approved):

johnclary:
@frankhereford @mddilley just a friendly FYI that we are deploying a DAG that runs every minute 😬 

i guess we should just see how this goes, but part of me wonders if we should modify the ETL to just run on a loop for like 12 hours before exiting. and the DAG's job would just be to spin up a new version of it on that cadence 🤔

frankhereford:
@johnclary wrote:

> just a friendly FYI that we are deploying a DAG that runs every minute 😬

Interesting! I'm totally game to see what happens. Hopefully that overhead firing up that docker container isn't that bad. I'm optimistic. 

To your idea about the 12 hour run time - When we're tempted to simulate a normal daemon process serving a on-demand service by running an ETL in a 1-minute loop, we should consider that airflow isn't the right place for it, and we should make a `systemctl` controlled service. I think John's arriving at the same conclusion here with his 12-hour idea, we're just using different controlling frameworks (airflow vs. `systemctl`). 

I wish there were an easy way to not have to poll for work to do for this process and have it able to respond to "something needs processing" events. Knack doesn't have anything in the way of web hooks or what-have-you that I am aware of, so I don't see a clear way to get that going though.

PR #217, opened 2024-03-20
Reduce the number of CR3s processed per invocation of the process by airflow

frankhereford:
## Associated issues

_None_, see [here](https://austininnovation.slack.com/archives/CM9MK950S/p1710947182415969). 

## Note

This is just a pitch for a possible solution. It's totally OK if we don't go this way, no problem at all.

## Associated repo

`atd-vz-data`

## Testing

none, code review only

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/vz_cr3_extract_ocr_narrative.py
+++ b/dags/vz_cr3_extract_ocr_narrative.py
@@ -66,7 +66,7 @@ with DAG(
         image="atddocker/atd-vz-cr3-extract:production",
         api_version="auto",
         auto_remove=True,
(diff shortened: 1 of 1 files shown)


mddilley (approved):
I think this is worth a shot too based on your linked thread, and we can always change it back! Thanks for looking into this DAG and updating. 🙏

chiaberry (approved):
a little springtime diet, sure

johnclary:
thanks, Frank!

PR #219, opened 2024-03-29
DAG for Employees to TPW Hire App

Charlie-Henry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15510

## Associated repo
https://github.com/cityofaustin/atd-knack-services/

## Testing

**Steps to test:**
Feel free to run this locally. You will not break anything.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_knack_employees_tpw_hire.py
+++ b/dags/atd_knack_employees_tpw_hire.py
@@ -106,7 +106,6 @@ with DAG(
         command=f"python ./atd-knack-services/services/records_to_postgrest.py -a {app_name_dest} -c {container_dest} {date_filter_arg}",
         environment=env_vars_t2,
         tty=True,
-        force_pull=True,
         mount_tmp_dir=False,
     )
 
@@ -118,7 +117,6 @@ with DAG(
         command=f"python ./atd-knack-services/services/records_to_knack.py -a {app_name_src} -c {container_src} --app-name-dest {app_name_dest} {date_filter_arg}",
         environment=env_vars_t3,
         tty=True,
-        force_pull=True,
         mount_tmp_dir=False,
     )
     date_filter_arg >> t1 >> t2 >> t3
--- a/dags/atd_knack_employees_tpw_hire.py
+++ b/dags/atd_knack_employees_tpw_hire.py
@@ -1,0 +1,124 @@
+# test locally with: docker compose run --rm airflow-cli dags test atd_knack_employees_tpw_hire
+
+import os
+
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration, now
+
+from utils.onepassword import get_env_vars_task
+from utils.slack_operator import task_fail_slack_alert
+from utils.knack import get_date_filter_arg
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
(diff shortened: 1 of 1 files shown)


dags/atd_knack_employees_tpw_hire.py
```diff
@@ -0,0 +1,124 @@
...
+
+    t2 = DockerOperator(
+        task_id="tpw_hire_employees_to_postgrest",
+        image=docker_image,
+        docker_conn_id="docker_default",
+        auto_remove=True,
+        command=f"python ./atd-knack-services/services/records_to_postgrest.py -a {app_name_dest} -c {container_dest} {date_filter_arg}",
+        environment=env_vars_t2,
+        tty=True,
+        force_pull=True,
```
chiaberry: do these need to be force_pull True after the first one?

chiaberry (approved):
tpw higher and higher

mddilley (approved):
I checked out the related PR and this diff, and the code looks good! I couldn't run locally on my Mac because the image on DockerHub isn't multi-architecture, but that won't stop it from running on atddata03. 🚢 🚀

frankhereford:
@mddilley wrote:

> I checked out the related PR and this diff, and the code looks good! I couldn't run locally on my Mac because the image on DockerHub isn't multi-architecture, but that won't stop it from running on atddata03. 🚢 🚀

I set it up so that it can be built locally for either arch in the last few days, but I didn't extend that work to include the workflow in the knack services repo. I did [spin off an issue](https://github.com/cityofaustin/atd-data-tech/issues/16596) to add the platform designation to the `buildx` command though, and I've added that to our next planning meeting. Thanks!

dags/atd_knack_employees_tpw_hire.py
```diff
@@ -0,0 +1,122 @@
...
+    default_args=DEFAULT_ARGS,
+    schedule_interval="30 1 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    tags=["repo:atd-knack-services", "knack", "hr-manager", "tpw-hire", "banner"],
+    catchup=False,
+) as dag:
+    docker_image = "atddocker/atd-knack-services:production"
+    app_name_src =  "hr-manager" 
+    container_src = "view_684"
+    app_name_dest = "tpw-hire"
+    container_dest = "view_148"
```
frankhereford, line 82: It feels a little like a anti-pattern to have the configuration in the orchestration configuration, but then again, here's as good as anywhere. It is where the program is kicked off from after all. I mostly wanted to flag it since it's a bit of a departure from what we've done in the past -- do you think 1PW is maybe a better place for these? I'm genuinely not sure and curious about your thoughts.

frankhereford (approved):
🚢🚢🚢🚢

dags/atd_knack_employees_tpw_hire.py
chiaberry: at lot of our knack dags follow this pattern, with the app_name and the container as vars. Its the pattern from airflow 1 too, so when the dags were brought over, they brought that pattern with them.

johnclary: yeah, i'm not sure where else to stick these generic CLI args. it would be kinda nice if we could easily look to like one big config file that has all of these across all knack services DAGs 🤷

Charlie-Henry: +1 what Chia said. If we want to start a new pattern we should probably refactor all of the knack DAGs in a separate PR. To me it doesn't feel like a 1PW thing because it's just part of the params to the command to the docker image but I don't have strong feelings either way

PR #205, opened 2024-02-01
temporarily turn off full replace for knack signals

chiaberry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/15525, sorta

## Testing

turning off the full replace flag for the rest of today, since its already done a full replace and this runs every 5 min

---
#### Ship list
- [x] Code reviewed 
- [ ] Product manager approved
- [ ] Add note to 1PW secrets moved to API vault and check for duplicates

PR #221, opened 2024-04-18
Once a month dag to fully replace signals

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/15525

Our knack datasets do a full replace on the first day of the month, but the knack signals dag runs every 5 minutes. So every 5 minutes on the first of the month, we would try to do a full replace, which takes more than 5 minutes. So failures all day. 

I turned off the monthly full replace (https://github.com/cityofaustin/atd-airflow/pull/205), so this PR is to bring back a full replace once a month.

## Associated repo

This is equivalent to the [existing knack_signals dag](https://github.com/cityofaustin/atd-airflow/blob/production/dags/atd_knack_signals.py), but the run schedule is only once on the first of the month, with no date flag so its always a full replace. 

I am not even sure if we need to do full replaces once a month. The other signals dag runs every 5 minutes, trying to see how to pause that one while this one runs.....

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_knack_signals_full_replace.py
+++ b/dags/atd_knack_signals_full_replace.py
@@ -1,0 +1,112 @@
+import os
+
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration, now
+
+from utils.onepassword import get_env_vars_task
+from utils.slack_operator import task_fail_slack_alert
+from utils.knack import get_date_filter_arg
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 1, 1, tz="America/Chicago"),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "execution_timeout": duration(minutes=15),
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+REQUIRED_SECRETS = {
+    "KNACK_APP_ID": {
+        "opitem": "Knack AMD Data Tracker",
+        "opfield": f"production.appId",
+    },
+    "KNACK_API_KEY": {
+        "opitem": "Knack AMD Data Tracker",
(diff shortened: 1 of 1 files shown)


Charlie-Henry (approved):
A little curious what will happen on the 1st of May but let's see 🤞

dags/atd_knack_signals_full_replace.py
```diff
@@ -0,0 +1,112 @@
...
+        "opfield": "production.password",
+    },
+}
+
+
+with DAG(
+    dag_id=f"atd_knack_signals_full_replace",
+    description="Load a full replace of signals (view_197) records from Knack to Postgrest to AGOL and Socrata",
+    default_args=DEFAULT_ARGS,
+    schedule_interval="28 2 1 * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
mddilley, line 68: @chiaberry One thing that crossed my mind when you mentioned figuring out how to pause the other DAG is the [`max_active_runs` setting](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/dag/index.html#airflow.models.dag.DagModel.max_active_runs) for a DAG. I'm a little weary of having multiple DAGs for different versions of the same process, and I'm wondering if there is some combination of a schedule (the complexity here might be a stretch for cron expressions but idk), branching logic or something to toggle the full replace arg, and using the max runs setting to make sure that no more than one DAG run occurs at the same time to let the full replace play out before the next run starts.

I'm know that you've put time into this already so maybe it would be a time sink, and we can always think about it more while we test this out.

johnclary: i am also wary of running multiple instances of this ETL at the same time, and agree with Mike that it would be a good idea to handle this with DAG code. 

one other wrinkle, iirc, is that the effect of a full replace is that the modified date of every record in the postgres db is going to be updated to the current date, and so every single time the DAG runs for the rest of the day it is going to process all records, because the postgres > socrata ETL queries by date, not timestamp. 

i don't know, you might want to altogether abandon running a full replace on this dataset. the current ETL is not optimized to deal with situation.

PR #225, opened 2024-05-02
Adjust DAG to use new arguments of metadata populator

frankhereford:
## Associated issues

This PR is part of closing https://github.com/cityofaustin/atd-data-tech/issues/16746.

## Associated repo & PR

[atd-vz-data](https://github.com/cityofaustin/atd-vz-data) and in particular, [this sibling PR](https://github.com/cityofaustin/atd-vz-data/pull/1436).

## Testing
* Spin up your VZ stack, at least enough to get a DB and graphql-engine instance running. 
  * For linux users (or those who are not using docker desktop), you may have to make considerations for the missing `host.docker.internal` hostname used in the DAG's development configuration. 
* Spin up the local airflow stack and execute the DAG.
  * Did it run? 
  * You're looking for a line in the logs similar to: `INFO - Items to process: 0`
* Look at the new schedule and see if you like it.
  * The goal is to get them all processed as soon as reasonable after CR3s are downloaded.
  * It's modified to run in a way that processes all pending operations, instead of doing it batch by batch. Because of this, I'm changing the schedule to run at the top of the hour, from 8AM to noon. To prevent the very unlikely case of it running longer than an hour and getting into the situation to where we have two going at once, I've set the `execution_timeout` to 50 minutes, just in case. 
  * Under this regime, at the top of the next hour after the CR3s are downloaded, we will rapidly get caught up and begin showing those CR3s on the VZE.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/vz_populate_cr3_metadata.py
+++ b/dags/vz_populate_cr3_metadata.py
@@ -11,7 +11,7 @@ default_args = {
     "email_on_failure": False,
     "email_on_retry": False,
     "retries": 0,
-    "execution_timeout": duration(minutes=5),
+    "execution_timeout": duration(minutes=50),
     "on_failure_callback": task_fail_slack_alert,
 }
 
@@ -48,7 +48,9 @@ REQUIRED_SECRETS = {
 @dag(
     dag_id="vz-populate-cr3-metadata",
     description="A DAG which populates CR3 metadata in the VZDB",
-    schedule_interval="0 9 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    schedule_interval=(
+        "0 8-12 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None
+    ),
     catchup=False,
     tags=["repo:atd-vz-data", "vision-zero", "ocr", "cr3"],
     default_args=default_args,
@@ -71,7 +73,10 @@ def populate_cr3_metadata():
         image="atddocker/vz-cr3-metadata-pdfs:production",
         auto_remove=True,
         entrypoint=[
-            "/app/populate_cr3_file_metadata.py -t 5 -a -v"
+            "/app/populate_cr3_file_metadata.py",
+            "-t 20",
+            "-a",
+            "-v",
         ],  # five threads, do all pending actions instead of batching them, and be verbose in logging.
         tty=True,
         force_pull=True,
--- a/dags/vz_populate_cr3_metadata.py
+++ b/dags/vz_populate_cr3_metadata.py
@@ -70,7 +70,9 @@ def populate_cr3_metadata():
         environment=env_vars,
         image="atddocker/vz-cr3-metadata-pdfs:production",
         auto_remove=True,
-        entrypoint=["/app/populate_cr3_file_metadata.py -t 5 -a"],
+        entrypoint=[
+            "/app/populate_cr3_file_metadata.py -t 5 -a -v"
+        ],  # five threads, do all pending actions instead of batching them, and be verbose in logging.
         tty=True,
         force_pull=True,
     )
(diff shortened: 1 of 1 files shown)


dags/vz_populate_cr3_metadata.py
```diff
@@ -72,7 +70,7 @@ def get_env_vars():
         environment=env_vars,
         image="atddocker/vz-cr3-metadata-pdfs:production",
         auto_remove=True,
-        entrypoint=["/app/populate_cr3_file_metadata.py"],
+        entrypoint=["/app/populate_cr3_file_metadata.py -t 5 -a"],
```
chiaberry: I may be reviewing this prematurely, but I went to check what -a means in the script, and I saw this comment: https://github.com/cityofaustin/atd-vz-data/pull/1436/files#r1566222905

should this include -v?

frankhereford: Naw, never too early to review! Thank you @chiaberry - I think that's a good addition, `-v`, to get more verbose logs. I'll add that right now. https://github.com/cityofaustin/atd-airflow/pull/225/commits/c04ab9a7fe02ef699e3e48e124d8bae67bec891c

chiaberry: I was just quoting your previous comment about airflow having -v and not -a!

frankhereford:
@johnclary & @Charlie-Henry -- thank you both for being willing to fit this review into your day today. 🙏🙏🙏🙏

dags/vz_populate_cr3_metadata.py
```diff
@@ -48,9 +48,9 @@
 @dag(
     dag_id="vz-populate-cr3-metadata",
     description="A DAG which populates CR3 metadata in the VZDB",
-    schedule_interval="*/5 8-10 * * *"
-    if DEPLOYMENT_ENVIRONMENT == "production"
-    else None,
+    schedule_interval=(
```
chiaberry, line 51: So it tries to process everything at 8am, and if it completes it under 50 min, all green squares, then it tries again to pick up new records (?) at 9a, 10a, 11a?

frankhereford: That is correct -- essentially, it's a non-op every time it runs except for the very first time it runs after CR3s have been downloaded by the dev doing that. I have gone with this try-every-hour thing because it's not known when the dev is going to do the downloading.

And if it happens to get fired up during the very short window of CR3 downloads, that's fine, it'll get the tail of the group at the top of the next hour.

chiaberry: ah! so if someone doesn't run the cr3 downloads before 11a CT then they would have to go in and trigger this dag themselves

frankhereford: You got it, 💯. Very technically, they have until noon to download CR3s, but anytime after that, yea, they should trigger this to get the CR3s to show on the VZE.

johnclary: would there be any downside of running this every hour from 8a-5p central? just so we don't have to remember to run it manually

frankhereford: No, not really, it's a nothing of a non-op to do that. I did the noon thing purely to stay aligned to the way it used to be where it was morning only. I can shift it to hourly during the whole business day.

Us devs won't let this encourage us to do cr3 downloads later in the day 😉 

https://github.com/cityofaustin/atd-airflow/pull/225/commits/9f9dd6eee0295c14584d6e4d9bd22a930ea098dc

johnclary (approved):
👍 looks great. if we can extend that schedule to cover the whole day, even better 🙏 🚢

dags/vz_populate_cr3_metadata.py
johnclary: 😉

chiaberry (approved):
i have not run this, but I 👍🏼

johnclary:
nice—thanks again!

frankhereford:
Thanks everyone for the discussion and for the super fast reviews on this today. I'm going to merge and deploy now.

PR #230, opened 2024-05-22
Task not function in task definition

Charlie-Henry:
Passed the function not the **task** from my last PR! This is what I get for not testing...

## Testing

**Steps to test:**

---
#### Ship list
- [x] Code reviewed 

### DIFF
--- a/dags/atd_knack_inventory_transactions.py
+++ b/dags/atd_knack_inventory_transactions.py
@@ -90,4 +90,4 @@ with DAG(
         mount_tmp_dir=False,
     )
 
-    get_date_filter_arg >> t1 >> t2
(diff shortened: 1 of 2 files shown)


frankhereford (approved):
I *have* spun up local airflow and I observe that these now compile. Thanks for this fix Charlie. 

I'm going to be more careful in the future about putting my ✅ on DAGs without at least spinning up the stack to see that they compile, which I failed to do on the previous PR. 😢  

Thank you for the very quick fix!

chiaberry (approved):

mddilley (commented):
thanks Charlie! i also dropped the ball on review on the last one 🫠

PR #231, opened 2024-05-22
PR in two fixes which have been made in situ in production; fix knack utility and cron job definition

frankhereford:
## Associated issues

/none/

These changes are in production already. I had promised to PR this in last sprint, and I forgot. I found the outstanding changes again last night, so here we are -- better late than never! Sorry for the delay on delivering this PR to y'all.

## Background for `dags/utils/knack.py`

Through a very informative and interesting conversation with @johnclary, the root cause of this situation was identified, and we're able to revert back to this ever-so-slightly changed code.

## Background for `toolbox/log-purge/log-purge.cron`

At some point in the recent past, we discovered that our cron job definition was malformed because it was to be used in the `/etc` definitions of jobs, where the user has to be specified for the daemon to know who to run the job as. We had used the format you use in `crontab -e` which intrinsically knows that the user to use is "you."

## Associated repo

`atd-airflow`

## Testing

These are in production, and have been for a few weeks, so we genuinely get this one for free.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/utils/knack.py
+++ b/dags/utils/knack.py
@@ -21,9 +21,7 @@ def get_date_filter_arg(should_replace_monthly=False, **context):
     """
 
     today = now()
-    prev_start_date = (
-        context.get("prev_start_date_success").isoformat() or today.isoformat()
-    )
+    prev_start_date = context.get("prev_start_date_success") or today
 
     if should_replace_monthly and today.day == 1:
         return ""
--- a/dags/utils/knack.py
+++ b/dags/utils/knack.py
@@ -1,5 +1,5 @@
 from airflow.decorators import task
-from pendulum import now
+from pendulum import now, parse, DateTime
 
 
 @task(
@@ -20,7 +20,19 @@ def get_date_filter_arg(should_replace_monthly=False, **context):
         Str or None: the -d flag and ISO date string or None
     """
     today = now()
-    prev_start_date = context.get("prev_start_date_success") or today.isoformat()
+    prev_start_date = context.get("prev_start_date_success")
+
+    if isinstance(prev_start_date, DateTime):
+        prev_start_date = prev_start_date.isoformat()
+    elif prev_start_date:
+        try:
+            prev_start_date = parse(prev_start_date).isoformat()
+        except ValueError:
+            prev_start_date = today.isoformat()
+    else:
+        prev_start_date = today.isoformat()
+
     if should_replace_monthly and today.day == 1:
         return ""
+
     return f"-d {prev_start_date}"
--- a/dags/utils/knack.py
+++ b/dags/utils/knack.py
@@ -19,18 +19,11 @@ def get_date_filter_arg(should_replace_monthly=False, **context):
     Returns:
         Str or None: the -d flag and ISO date string or None
     """
-    today = now()
-    prev_start_date = context.get("prev_start_date_success")
 
-    if isinstance(prev_start_date, DateTime):
-        prev_start_date = prev_start_date.isoformat()
-    elif prev_start_date:
-        try:
-            prev_start_date = parse(prev_start_date).isoformat()
-        except ValueError:
-            prev_start_date = today.isoformat()
-    else:
-        prev_start_date = today.isoformat()
+    today = now()
+    prev_start_date = (
+        context.get("prev_start_date_success").isoformat() or today.isoformat()
+    )
 
     if should_replace_monthly and today.day == 1:
         return ""
--- a/dags/utils/knack.py
+++ b/dags/utils/knack.py
@@ -28,4 +28,4 @@ def get_date_filter_arg(should_replace_monthly=False, **context):
     if should_replace_monthly and today.day == 1:
         return ""
 
-    return f"-d {prev_start_date}"
+    return f"-d {prev_start_date.isoformat()}"
--- a/dags/utils/knack.py
+++ b/dags/utils/knack.py
@@ -1,5 +1,5 @@
 from airflow.decorators import task
-from pendulum import now, parse, DateTime
+from pendulum import now
 
 
 @task(
--- a/toolbox/log-purge/log-purge.cron
+++ b/toolbox/log-purge/log-purge.cron
@@ -3,7 +3,7 @@
 # time greater than 30 days, and the second command removes all empty directories. Removed
 # files and directories are logged into /var/log which will get rotated by the system.
 
-0 2 * * * find /usr/airflow/atd-airflow/logs -type f -mtime +30 -exec rm -v {} \; \
+0 2 * * * root find /usr/airflow/atd-airflow/logs -type f -mtime +30 -exec rm -v {} \; \
 >> /var/log/airflow-logs-purge.log && \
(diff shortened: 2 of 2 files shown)


Charlie-Henry (approved):
Makes sense. Too bad pendulum made this more complicated than it probably had to be.

dags/utils/knack.py
```diff
@@ -20,7 +20,19 @@ def get_date_filter_arg(should_replace_monthly=False, **context):
...
-    prev_start_date = context.get("prev_start_date_success") or today.isoformat()
+    prev_start_date = context.get("prev_start_date_success")
+
+    if isinstance(prev_start_date, DateTime):
+        prev_start_date = prev_start_date.isoformat()
+    elif prev_start_date:
+        try:
+            prev_start_date = parse(prev_start_date).isoformat()
+        except ValueError:
+            prev_start_date = today.isoformat()
```
johnclary: @frankhereford i'm a bit confused about this `elif` block. the Airflow [template string docs](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html) suggest that `prev_start_date_success` is either going to be a `DateTime` instance or `None`, but this would suggest it may also sometimes be a string?

johnclary: ah, is it that the context variable is always a string or `None`, and that string is no longer in ISO format? e.g. airflow broke the shape of their own context vars by upgrading pendulum? if that's the case—yikes.

either way, its seems like there must be part of this if block that will never execute 🤷

frankhereford: Yea, that's a great question John. The reason for these hoops is because the date we were getting out of the `context.get()` function was a string. We had previous dates coming out that we had to handle in addition to the form of new ones we were creating.

frankhereford: Airflow handled the transition correctly, I believe, but relied on the stack to be upgraded in such a way that a specific minor version not be skipped entirely, which we did end up skipping. I'm not saying they couldn't have left more handlers in place for longer or forever, but I can see that they didn't want to support something they said they were no longer supporting. They were very clear that they were intent on moving to Pendulum 3, and we just missed the whole thing by doing such a big upgrade jump.

frankhereford: @johnclary wrote: 

> either way, its seems like there must be part of this if block that will never execute 🤷

Forgive me if I'm missing something here -- i'm a little confused -- but I read this as:

If we get a `DateTime` out of the context, use it
Else, if we get something else out of the context, try to parse it as a string into a DateTime, and if that fails, just use now. 
And finally, if we get nothing out of the context, use now.

Can you help me understand which part is unreachable? I'd be happy to sync with you if that's easier.

johnclary: my confusion is why `context.get("prev_start_date_success")` may sometimes return a string and sometimes a `DateTime` instance. the previous code assumed this value was always a string or `None`.  So the new version of the code would be like this:
```python
if prev_start_date:
    prev_start_date = parse(prev_start_date).isoformat()
else:
    prev_start_date = today.isoformat()
```

frankhereford: Ah, that makes so much sense!

The reason was -- just for the immediate time while we were in the transition / upgrade, we had contexts from the 2.6 era which were coming out of the context as a string, as they had been stored as a string, and we were storing back in DataTimes, which the next time the dag ran, we'd have DateTimes coming out. This "handle them both" fix was for the transition, since we missed the versions where Airflow had this "handle them both" fix internal to it.

frankhereford: If I knew that all of our DAGs which have ever run had run again since the upgrade, this code could be simplified to eliminate one of the branches of the condition, just like you've provided. However, I don't know that for sure. We could write up an issue to find out and to make that enhancement if you think it'd be a good thing to do.

johnclary: oh, the context API stores data from earlier airflow versions! that's bananas 😵 

would you mind adding a comment to the block that says something about backwards compatible support for string instances of this context var?

i'm still pretty baffled, because the release notes for [this change](https://airflow.apache.org/docs/apache-airflow/stable/release_notes.html#context-variables-prev-execution-date-success-and-prev-execution-date-success-are-now-pendulum-datetime) are under the Airflow v2 release notes.

> Context variables prev_execution_date_success and prev_execution_date_success are now pendulum.DateTime


🤔

johnclary: hmm, i guess i'm not convinced now. are you sure [this block](https://github.com/cityofaustin/atd-airflow/pull/231/files#discussion_r1610348924) doesn't get the job done?

frankhereford: Sure thing on the comment. 

Yea, I do hear you on the baffling observation you made there. I should point out that this code was born out of a trial-and-error debugging session because we had DAGs failing. I'm not sure if it's Airflow misbehaving or maybe our own code misbehaving -- no idea what the root of the strings are -- but I was seeing strings and DateTimes and this seemed reasonable to handle both.

frankhereford: @johnclary and I had an opportunity to discuss this in a huddle, and we're both interested in finding out the real details of what happened here. I think he has a fuller picture of the timeline of events after reading documentation. In terms of what we're going to do right now, I'm going to put the simplified block of code which he suggests in place, and if it works, then I'm going to revise this PR to reflect it. Thanks again to John for the discussion and the very valuable development perspective of wanting a full understanding and not acting blindly.

dags/utils/knack.py
```diff
@@ -20,7 +20,19 @@ def get_date_filter_arg(should_replace_monthly=False, **context):
...
+    else:
+        prev_start_date = today.isoformat()
+
     if should_replace_monthly and today.day == 1:
         return ""
+
     return f"-d {prev_start_date}"
```
johnclary: ah ha! so the bug in the old function was actually here. we were assuming this value was always a string, when in fact it was always served out of the airflow context  as a pendulum `DateType` instance. the string formatting here was implicitly calling the pendulum method with a breaking v3 change.

dags/utils/knack.py
```diff
@@ -19,8 +19,13 @@ def get_date_filter_arg(should_replace_monthly=False, **context):
     Returns:
         Str or None: the -d flag and ISO date string or None
     """
+
     today = now()
-    prev_start_date = context.get("prev_start_date_success") or today.isoformat()
+    prev_start_date = (
+        context.get("prev_start_date_success").isoformat() or today.isoformat()
+    )
```
johnclary: @frankhereford i think this needs a change, because ` context.get("prev_start_date_success").isoformat()` fails with `AttributeError ` when it is `None`. 

i like that you added the call to `isoformat()` in the return statement, so i think that all that's needed is to change this line to:
```python
 prev_start_date = context.get("prev_start_date_success") or today
```

frankhereford: Great catch, thank you. 💯 https://github.com/cityofaustin/atd-airflow/pull/231/commits/97460297ed77ade0bf7d2a5556dd99e832522d90

chiaberry (approved):
glad yall got to the bottom of it

johnclary (approved):
all set—thanks again!!! 🚢

PR #234, opened 2024-06-17
In situ changes from server migration for airflow

frankhereford:
## Associated issues

This PR is related to https://github.com/cityofaustin/atd-data-tech/issues/17627

## Associated repo

`atd-airflow`

## Testing

_none_

These changes were made in place during the migration to the new on-prem servers. These are already in production.

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/haproxy/haproxy-production.cfg
+++ b/haproxy/haproxy-production.cfg
@@ -7,6 +7,7 @@ global
   user haproxy
   group haproxy
   daemon
+  maxconn 4096
 
   # Default SSL material locations
   ca-base /etc/ssl/certs
--- a/toolbox/certbot/renew_domain_with_certbot.sh
+++ b/toolbox/certbot/renew_domain_with_certbot.sh
@@ -4,7 +4,7 @@
 DOMAIN=$1
 
 echo "Renewing certificate for $DOMAIN"
-export ATD_AIRFLOW_HOMEDIR="/usr/airflow/atd-airflow";
+export ATD_AIRFLOW_HOMEDIR="/srv/atd-airflow";
 
 # Load the same environment variables as the Airflow stack
 source $ATD_AIRFLOW_HOMEDIR/.env
@@ -24,7 +24,7 @@ AWS_SECRET_ACCESS_KEY=$(docker run --rm --name op \
 1password/op:2 op read op://$OP_VAULT_ID/Certbot\ IAM\ Access\ Key\ and\ Secret/accessSecret)
 
 # Now, remove the old concatenated certificates, renew the certificate, and replace with the new concatenated certificates
-CERT_PATH="/usr/airflow/atd-airflow/haproxy/ssl"
+CERT_PATH="/srv/atd-airflow/haproxy/ssl"
 cd $CERT_PATH
 rm $DOMAIN.pem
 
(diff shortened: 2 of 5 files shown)


haproxy/haproxy-production.cfg
```diff
@@ -7,6 +7,7 @@ global
   user haproxy
   group haproxy
   daemon
+  maxconn 4096
```
chiaberry, line 10: I was reading about what this does, but why did we not have it on our old configuration?

frankhereford: @chiaberry wrote:

> I was reading about what this does, but why did we not have it on our old configuration?

Great question. I'm chalking it up to fundamental differences between the linux kernel on the centos machines (3.10.0) vs RHEL (5.14.0). RHEL's configuration isn't willing to give up a bajillion potential handles for connections, so we have to tell haproxy to settle down and "only" be ready for 4K incoming requests. 

Two major linux releases is about an eon in linux development time too -- that's the biggest kernel version jump I've ever been a part of. Total Encino Man vibes.

toolbox/certbot/renew_domain_with_certbot.sh
```diff
@@ -40,3 +40,5 @@ certbot/dns-route53 certonly -n --dns-route53 -d $DOMAIN
 cat /etc/letsencrypt/live/$DOMAIN/cert.pem > $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/$DOMAIN.pem
 
 cat /etc/letsencrypt/live/$DOMAIN/privkey.pem >> $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/$DOMAIN.pem
+
+chmod a+r $ATD_AIRFLOW_HOMEDIR/haproxy/ssl/$DOMAIN.pem
```
mddilley, line 44: cool, i had to look up `a+r`. so this makes the cert file readable by all users

haproxy/haproxy-production.cfg
mddilley: thanks for that explanation. 🙏 because i have no clue, did this surface through an error? trying to learn as i go 😅

mddilley (approved):
All of this looks good to me and matches up with the other changes that I've been seeing. Thanks for updating! 🚢

toolbox/certbot/renew_domain_with_certbot.sh
frankhereford: Yea - turns out we need this on the new systems -- my guess is that it's related to there not being overlap in the UIDs used on the RHEL vs the one's in the debian based docker images? This is a wild guess, but for whatever reason, it needs to be set.

haproxy/haproxy-production.cfg
frankhereford: You bet! It did surface through an error. `haproxy` won't start at all if it isn't able to start with the configuration it has, so the container was croaking as soon as it fired up with an error that the kernel wasn't able to allocate enough handles -- by default, `haproxy` sets itself up to serve an enormous amount of traffic, like a `reddit.com` amount of traffic. `haproxy` will easily swamp a gigabit ethernet connection before it even comes close to bumping into a CPU restriction. It's just one of those "how do they do it" type of computing masterpieces.

PR #235, opened 2024-06-17
parking with no prev exec date fix

Charlie-Henry:
I had hard coded in August of 2023 for parking data when we didn't have a prev exec date. This changes it to only get the last two weeks of data if we have no prev exec date.

[airflow macros docs](https://airflow.apache.org/docs/apache-airflow/1.10.13/macros-ref.html)

## Associated issues
none!

## Associated repo
[atd-parking-data](https://github.com/cityofaustin/atd-parking-data)
## Testing

**Steps to test:**
I don't think we need to test this.

---
#### Ship list
- [ ] Code reviewed 


frankhereford (approved):
🚢

PR #237, opened 2024-06-25
Incremental updates in Moped > AGOL ETL 

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/14479

Small PR to update the frequency of this DAG and to pass the `date_filter_arg` to the updated Moped script.

How this works:
1. The updated DockerOperator command pass a `-d` flag with a timestamp with timezone date of the last successful run (looks like `python components_to_agol.py -d 2024-07-01T00:06:16.360805+00:00`)
2. The [script was updated](https://github.com/cityofaustin/atd-moped/pull/1369) to expect this flag and date to determine how far back to look for updated project components

## Associated repo
https://github.com/cityofaustin/atd-moped/pull/1369

## Testing

**Steps to test:**
I could use some feedback on this deploy plan to make sure I'm thinking about this right. 🙏

1. This is an existing DAG so it will have a `prev_start_date_success` ([docs](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html)) coming from our `get_date_filter_arg` helper when this code deploys.
3. So, to deploy this update which goes from once-a-day full replace (~8-15 min run duration recently) to every 5 min updates of only Moped projects updated since the last successful run (way faster run duration), the only risk is having the first run with the updated schedule go longer than 5 min.
4. I think that the chance is very low especially since we would deploy this updated Airflow code the same day as the Moped code which would mean 1 day of updates in projects from Moped users.
5. If something goes wrong, the script still supports a full replace so we can pause the DAG schedule, wipe the dataset by running the script locally with the full replace option, and then resume the schedule to pick up future Moped project updates.

---
#### Ship list
- [x] Code reviewed 
~- [ ] Add note to 1PW secrets moved to API vault and check for duplicates~

### DIFF
--- a/dags/atd_moped_components_to_agol.py
+++ b/dags/atd_moped_components_to_agol.py
@@ -47,7 +47,7 @@ with DAG(
     dag_id="atd_moped_components_to_agol",
     description="publish component record data to ArcGIS Online (AGOL)",
     default_args=DEFAULT_ARGS,
-    schedule_interval="30 22 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    schedule_interval="*/5 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
     dagrun_timeout=duration(minutes=30),
     tags=["repo:atd-moped", "moped", "agol"],
     catchup=False,
--- a/dags/atd_moped_components_to_agol.py
+++ b/dags/atd_moped_components_to_agol.py
@@ -48,7 +48,7 @@ with DAG(
     description="publish component record data to ArcGIS Online (AGOL)",
     default_args=DEFAULT_ARGS,
     schedule_interval="*/5 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
-    dagrun_timeout=duration(minutes=30),
+    dagrun_timeout=duration(minutes=5),
     tags=["repo:atd-moped", "moped", "agol"],
     catchup=False,
 ) as dag:
--- a/dags/atd_moped_components_to_agol.py
+++ b/dags/atd_moped_components_to_agol.py
@@ -8,6 +8,7 @@ from pendulum import datetime, duration, now
 
 from utils.onepassword import get_env_vars_task
 from utils.slack_operator import task_fail_slack_alert
+from utils.knack import get_date_filter_arg
 
 DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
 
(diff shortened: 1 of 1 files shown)


chiaberry:
I am still in the process of testing the script, but I get confused over what timezone airflow is in vs moped db. I'm thinking about when we tested the moped data tracker dag and how initially the updated at times were throwing me off

mddilley:
@chiaberry I get confused too so I double-checked that we don't override the timezone in our config but [we stick to the default of UTC](https://github.com/cityofaustin/atd-airflow/blob/7320415d05f9192abe79b4d8ab0e3feafe8ffb4b/airflow.cfg#L18) luckily and we also use UTC in Moped DB too.

johnclary:
👍 i was also thinking about timestamp, and since Airflow gives us that last run date as an ISO string with the TZ offset, we are good. we do need to be mindful running locally without a TZ offset, since the TZ is implicitly UTC unless stated otherwise

```
$python components_to_agol.py -d 2024-06-27
```

chiaberry:
oh yes, thats what it was. me running the data tracker dag locally!

chiaberry (approved):
I remembered leaving a comment about timing but completely forgot to actually hit approve 😬

roseeichelmann (approved):
this looks great and your deploy plan sounds solid to me. i didnt test running the script here but the code changes look right to me. these pair PRs are an awesome update, will be so much cleaner and efficient to have the ETL running more often but not replacing the entire dataset each time :sparkles:

dags/atd_moped_components_to_agol.py
```diff
@@ -46,20 +47,22 @@
     dag_id="atd_moped_components_to_agol",
     description="publish component record data to ArcGIS Online (AGOL)",
     default_args=DEFAULT_ARGS,
-    schedule_interval="30 22 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    schedule_interval="*/5 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
     dagrun_timeout=duration(minutes=30),
```
frankhereford: I think the timeout should be adjusted down to `minutes=5` or less to avoid getting into a situation where we have these pile up in concurrent executions if something goes sideways.

frankhereford (changes requested):
Mike, this looks great. I have not run the DAG locally here in my dev airflow, but I have checked to see that it compiles, which feels to me to cover the scope of this particular PR.

I think your deployment plan also sounds good, and I recognize the possible conflict between the change I'm requesting and the initial run that does the full replacement of all records in AGOL. 
  * I think the code that ends up in the repo should reflect the intended steady-state of the application, which is a <5 minute run-time.
  * If need be, for the initial execution, the `dagrun_timeout` can be commented out and the DAG be run manually

I hope that makes sense, and thank you again for this change!

mddilley:
@frankhereford Great catch! I just updated. 🙏🙏

frankhereford (approved):
🚀🚀🚀🚀🚀

PR #229, opened 2024-05-21
Add get_date_filter_arg to task flow definition

Charlie-Henry:
Two of our DAGs were failing intermittently when the first first task `t1` was run concurrently with `get_date_filter_arg`. Two of our DAGs were missing it in the task flow part.

https://austininnovation.slack.com/archives/C013G4RNJ01/p1708925588292689

## Associated issues
none

## Associated repo
various 

## Testing

**Steps to test:**
Kinda hard to test this as sometimes these DAGs will succeed even without this PR. 

---
#### Ship list
- [x] Code reviewed 
- [ ] Product manager approved
- [ ] Add note to 1PW secrets moved to API vault and check for duplicates

PR #194, opened 2024-01-02
Signal requests: Add `date_filter_arg` task to control flow

johnclary:
## Associated issues
- https://github.com/cityofaustin/atd-data-tech/issues/15107

## Associated repo
- atd-knack-services

## Testing

**Steps to test:**
1. Start airflow and trigger the DAG. Or just observe the new and improved graph.
Before:

Screenshot 2024-01-02 at 2 30 24 PM

After:

Screenshot 2024-01-02 at 2 43 12 PM

---
#### Ship list
- [ ] Code reviewed 
- [ ] Product manager approved
- [ ] Add note to 1PW secrets moved to API vault and check for duplicates

PR #238, opened 2024-07-15
Fix Moped AGOL DAG to wait for date arg before proceeding

mddilley:
## Associated issues
Fixes the error below that happens every once in a while because task 1 starts before getting the last run date to pass in the script invocation. I checked other DAGs to make sure this isn't happening elsewhere, and it looks like we are covered across the board.

```
components_to_agol.py: error: unrecognized arguments: None
```

See https://austininnovation.slack.com/archives/C013G4RNJ01/p1720812617018969

## Associated repo
n/a

## Testing

**Steps to test:**
n/a it is a small diff

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_moped_components_to_agol.py
+++ b/dags/atd_moped_components_to_agol.py
@@ -69,4 +69,4 @@ with DAG(
         mount_tmp_dir=False,
     )
 
-    t1
+    date_filter_arg >> t1


johnclary (approved):
🌬️ ⛵

dags/atd_moped_components_to_agol.py
```diff
@@ -69,4 +69,4 @@
         mount_tmp_dir=False,
     )
 
-    t1
+    date_filter_arg >> t1
```
frankhereford, line 72: This is a good change, establishing an order that these must be run in. Would you consider including the third task in here so that we are assured that there will always be populated secrets before starting the subsequent tasks?

frankhereford (changes requested):
I have proposed an enhancement. LMK if you think it's out of scope or wrong, and we can discuss -- and I'll flip this to an approval.

dags/atd_moped_components_to_agol.py
chiaberry: should we put `env_vars` before all of our tasks then? 

`date_filter_arg >> env_vars >> t1 >> t2` etc?

mddilley: this is a really good point! Really makes me wonder if we've seen the same type of error for 1PW secrets. I don't see any DAGs where we do this but it seems like we should, right? `get_env_vars_task` has a task decorator and also it is an async request so it seems even more susceptible to a race condition... 🤔 

I'm going to add it. 🙏

mddilley: i think we might be overlooking something. I'm looking at the graph from John's PR, and I'm seeing `env_vars` as a dependency without declaring it this way. https://github.com/cityofaustin/atd-airflow/pull/194.

I'm going to take a closer look. 🔍

chiaberry: maybe because its passed into the Docker Operator whereas the date_filter_arg is in a string? I didnt think about this difference

frankhereford: I recall that the new DAG format provides for a more "implicit" way of setting the order and dependencies. For example, before anything has been added, and it's just `date_filter_arg >> t1` is in the file, Airflow is able to figure this much about the DAG's graph:

image

mddilley: cool, thanks for pasting that in here, Frank! then, i'll keep this PR as-is since it would be a much larger change to update every DAG that uses `get_env_vars_task`. Re-requesting review 🙏

frankhereford: I think this PR is a non-change under this assumption. Backing out the change of e4ededac2b51333d52c0b59922b3deb3e5201ac7 leaves Airflow reporting this graph for the DAG. I feel like I must be overlooking something - can you point me in the right direction?

image

chiaberry (approved):
I remember Charlie fixing this before (https://github.com/cityofaustin/atd-airflow/pull/229), I'm sorry I didnt remember this when I originally reviewed the dag!

frankhereford (approved):
I don't think this change is making a difference in how airflow interprets the order of tasks, and I think we should continue to monitor for the problem. 

That said -- this PR makes the ordering more explicit and less implicit, and that is a good thing. Thank you Mike!

dags/atd_moped_components_to_agol.py
johnclary: it's really confounding :/  there's a slack thread somewhere in which i naively declare this to be an Airflow bug 🤷 

the `moped_components_to_agol` operator is getting passed the string `"None"` in  place of `{date_filter_arg}`. i'm resorting to speculation in our duck-typed world, but the [`get_date_filter_arg` ](https://github.com/cityofaustin/atd-airflow/blob/824cea830ea52ca35411b72846aef883bfdbe6ed/dags/utils/knack.py#L8) function itself cannot return `None` or `"None"` as far as i can tell, but it could use another look.

for what it's worth this patch did seem to resolve this error for the signal status publisher. it's really confounding :/

mddilley: agreed! thanks for the discussion, y'all. I'm going to merge and pull the changes on p01. I'll monitor but also pretty confident this will fix it even if we don't fully understand it.

johnclary: fwiw here's what this DAG looks like in my local `production` branch — it's showing the missing dependency on `get_date_filter_arg` 🤷 

Screenshot 2024-07-16 at 11 04 22 AM

PR #240, opened 2024-07-30
DAG updates for new VZ data model

johnclary:
## Associated issues

- https://github.com/cityofaustin/atd-data-tech/issues/18439
- https://github.com/cityofaustin/atd-data-tech/issues/18729
- https://github.com/cityofaustin/atd-data-tech/issues/13139

## Testing

1. Start your local VZ stack from `master`
2. Start your local airflow stack
3. Trigger the CRIS import DAG. It's going to fail once it encounters an extract with no PDFs directory. That'll be fixed here: https://github.com/cityofaustin/atd-vz-data/pull/1518. I think this is a good enough test of this DAG?
4. Trigger the Socrata Export DAG

---
#### Ship list
- [ ] Code reviewed 


chiaberry (approved):
tested running both dags. Got green for the socrata export and after a while timed out for the cris import one. but it ran and was chugging along so ✔️

johnclary:
thanks @chiaberry. i am realizing it's a little odd not to have a `staging` setting on these DAGs, since both of them do have staging available. i'm going to work that `DEPLOYMENT_ENVIRONMENT` handler to make that possible.

dags/vz_socrata_export.py
```diff
@@ -39,34 +31,62 @@
...
+        command=f"./socrata_export.py --people",
         environment=env_vars,
+        auto_remove=True,
         tty=True,
         force_pull=True,
-        mount_tmp_dir=False,
+        trigger_rule="all_done"  # always run this task regardless of outcome of crashes task
     )
+
+    socrata_export_crashes >> socrata_export_people
```
mddilley, line 117: hooray!!!! 🎉 🙏

dags/vz_cris_import.py
```diff
@@ -1,31 +1,79 @@
...
+
+docker_image = f"atddocker/vz-cris-import:{'production' if DEPLOYMENT_ENVIRONMENT == 'prod' else 'latest'}"
+
+
+DEFAULT_ARGS = {
+    "depends_on_past": False,
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 1,
+    "execution_timeout": duration(minutes=60),
```
mddilley, line 74: Chia's test run cleared the dev extract inbox (and so did mine) so I encountered the expected error of `OSError: No extract zips found in S3 bucket`. I noticed that the DAG hangs while it is in `up_for_retry` status. I can't remember if we will see the Slack alert at this point or only when we it truly fails so I wanted to check in case this means we would see a delayed alert.
Screenshot 2024-08-22 at 1 29 53 PM

mddilley (approved):
Moved one extract into the dev inbox, ran the import ✅, and ran the Socrata export ✅. I see the see the test datasets updated with the same crash and people counts that I saw logged in the import. So nice to have the crash and people datasets split out. 🚀 

I left one comment about the Slack alert behavior when no extract is found that I happened to run into while testing.

dags/vz_cris_import.py
johnclary: ah yes, great point. iirc we won't see an alert until retries have been expended. perhaps nixing retries altogether is the move. i'd be curious to know, at least in the early going, if retries are even needed 🤔

johnclary: i opted to set `retries` to `0` for now 🙏

PR #243, opened 2024-08-06
fix view for attachments

Charlie-Henry:
Looks like this DAG was referencing the wrong knackpy config. It is a dupe of the [marking specifications DAG](https://github.com/cityofaustin/atd-airflow/blob/137fbaffdf47fb089b46182c64df51c48ed06f0d/dags/atd_knack_markings_specifications.py#L63). This is supposed to reference [the marking attachments view](https://github.com/cityofaustin/atd-knack-services/blob/f984eb658b705e52a2049811cd7473af560e393c/services/config/knack.py#L266-L274). It was also replacing the whole AGOL layer every day due to some incorrect config [addressed in this PR](https://github.com/cityofaustin/atd-knack-services/pull/143).

[slack thread](https://austininnovation.slack.com/archives/CHZE6BC6L/p1722287448660389)

## Associated issues
n/a

## Associated repo

## Testing

**Steps to test:**
This DAG should work, but you will not see the reduced amount of AGOL record updates until we merge [the knack-services PR](https://github.com/cityofaustin/atd-knack-services/pull/143).

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_knack_markings_attachments.py
+++ b/dags/atd_knack_markings_attachments.py
@@ -1,3 +1,5 @@
+# test locally: docker compose run --rm airflow-cli dags test atd_knack_markings_attachments
+
 import os
 
 from airflow.models import DAG
@@ -17,7 +19,7 @@ DEFAULT_ARGS = {
     "email_on_failure": False,
     "email_on_retry": False,
     "retries": 0,
-    "execution_timeout": duration(minutes=30),
+    "execution_timeout": duration(minutes=60),
     "on_failure_callback": task_fail_slack_alert,
 }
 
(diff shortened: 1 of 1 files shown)


mddilley (approved):
Thanks for untangling! This view id matches the markings work attachments config from the other PR. 🚢

dags/atd_knack_markings_attachments.py
```diff
@@ -80,11 +83,11 @@
     t2 = DockerOperator(
         task_id="atd_knack_markings_attachments_to_agol",
         image=docker_image,
+        docker_conn_id="docker_default",
         auto_remove=True,
         command=f"./atd-knack-services/services/records_to_agol.py -a {app_name} -c {container} {date_filter_arg}",
         environment=env_vars,
         tty=True,
-        force_pull=True,
```
johnclary, line 87: i guess we don't need to force pull bc it's the same image that gets force pulled above 👍

johnclary (approved):
nice!

PR #246, opened 2024-08-19
Parking ETL: Remove PARD account and set 14 days as default

Charlie-Henry:
Finance pointed out we were missing a tiny amount of transactions that were not coming into Power BI. This is because some were actually still being processed after the 7 day window this ETL uses. I've expanded it to 14 days to hopefully capture these transactions. I've also dropped the `pard` user account as that one has not had any data posted to the S3 bucket in over a year now. 

## Associated issues
related to: https://github.com/cityofaustin/atd-data-tech/issues/18330

## Associated repo
https://github.com/cityofaustin/atd-parking-data

## Testing

**Steps to test:**
I would suggest not testing this as it can take a looong time to run due to the flowbird API having a 1 request per 60 seconds limit 😑 this makes it so you can step on each other's toes (or the ETL which is scheduled at 3:35AM) while testing. Maybe just post in the comments that you are beginning/end testing

if you're feeling up to it, be on the VPN and:

> $ docker compose run --rm airflow-cli dags test atd_parking_data

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_parking_data.py
+++ b/dags/atd_parking_data.py
@@ -152,8 +152,8 @@ with DAG(
 ) as dag:
     env_vars = get_env_vars_task(REQUIRED_SECRETS)
 
-    # default to the last 7 days of transactions but if we haven't ran it before, collect the last two weeks.
-    prev_exec = "{{ (prev_start_date_success - macros.timedelta(days=7)).strftime('%Y-%m-%d') if prev_start_date_success else (execution_date - macros.timedelta(days=14)).strftime('%Y-%m-%d')}}"
+    # default to the last 14 days of transactions
+    prev_exec = "{{ (prev_start_date_success - macros.timedelta(days=14)).strftime('%Y-%m-%d') if prev_start_date_success else (execution_date - macros.timedelta(days=14)).strftime('%Y-%m-%d')}}"
 
     docker_tasks = []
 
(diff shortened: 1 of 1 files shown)


dags/atd_parking_data.py
```diff
@@ -152,8 +152,8 @@
 ) as dag:
     env_vars = get_env_vars_task(REQUIRED_SECRETS)
 
-    # default to the last 7 days of transactions but if we haven't ran it before, collect the last two weeks.
-    prev_exec = "{{ (prev_start_date_success - macros.timedelta(days=7)).strftime('%Y-%m-%d') if prev_start_date_success else (execution_date - macros.timedelta(days=14)).strftime('%Y-%m-%d')}}"
+    # default to the last 14 days of transactions
+    prev_exec = "{{ (prev_start_date_success - macros.timedelta(days=14)).strftime('%Y-%m-%d') if prev_start_date_success else (execution_date - macros.timedelta(days=14)).strftime('%Y-%m-%d')}}"
```
chiaberry, line 156: should the else part be expanded to 21 days, or just remove the else?

Charlie-Henry: We still need the else to handle the case that `prev_start_date_success` is null

chiaberry: ah now i get it, i misread it.

johnclary (approved):
this is a good PR—there's a bunch of red.

📦 ➡️ ⚙️ ➡️ 🤑

chiaberry (approved):
not only did i misread the code change but then i missed hitting approve entirely. 🤕

mddilley (approved):
Thanks for the warning about testing taking a long time 🐌, I read the diff but did not run the DAG. Makes sense to me! 🧹🚢

Charlie-Henry:
ty guys!

PR #247, opened 2024-08-26
Add params to Moped to AGOL DAG

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/18469

This PR adds a new DAG parameter to control the flow of the DAG to be either incremental (current behavior) or a full data replacement (new and currently only possible with a local run). This will help us execute a full replacement to backfill new fields in the AGOL dataset when we add them in the future.

I tried and failed to use the parameter to dynamically set the execution timeout parameter in the DockerOperator task, and, the more that I read, it felt like I was use the wrong tool for the problem. Open to all feedback on the implementation since we haven't used branching before.

Here are some links:
- https://airflow.apache.org/docs/apache-airflow/2.9.0/core-concepts/dags.html#branching
- And what i read about passing data at runtime: 
   - Xcom https://www.astronomer.io/docs/learn/airflow-passing-data-between-tasks 
   - Similar issue that i ran into https://stackoverflow.com/questions/64235496/format-jinja-template-as-int-in-operator-parameter-in-airflow
      - I even tried writing awkward code using [Jinja templating](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html) like `execution_timeout=duration(minutes=int("{{ 30 if params.full_replace else 5 }}")` which Airflow wouldn't accept as a working DAG since the logic inside the string doesn't resolve until runtime so it sees a long string that is an invalid argument for `int()`

## Associated repo
https://github.com/cityofaustin/atd-moped/tree/main/moped-etl/arcgis

## Testing

**Steps to test:**
1. To test, you will have to disable the production DAG before proceeding. You will also need to fully complete these steps to make sure the dataset remains complete. After completing the steps, turn the production DAG back on to make sure we continue to pick up updated Moped components. It will take about 15 minutes.
2. Start up your local Airflow stack and then go to this DAG in the dashboard. Click the Play button to trigger a run, and you will see a new toggle labeled `full_replace`. Keep the radio toggled off and press the **Trigger** button.
3. You should see `Starting sync. Finding projects updated since <timestamp> and replacing components data...` logged and 
4. Next, trigger the run with the `full_replace` radio toggled on. This run takes much longer and fully replaces ~5600 points features, ~6200 line features, and then the combined ~11800 features to three AGOL layers.
5. You should see `Starting sync. Replacing all projects' components data...` logged.
6. After it is complete, please turn the production schedule back on. Thanks!

**Note:** The DAG defaults to `full_replace = False` so, when scheduled, the DAG will run incrementally, and a manually trigger full replace will interrupt the schedule, run the full replace, and revert to incremental runs afterwards. The scheduled when I tested locally looked like:
Screenshot 2024-08-26 at 3 33 23 PM

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/atd_moped_components_to_agol.py
+++ b/dags/atd_moped_components_to_agol.py
@@ -48,10 +48,10 @@ REQUIRED_SECRETS = {
 @task(
     task_id="get_args",
 )
-def get_args(arg_override, **context):
+def get_args(params, **context):
     """Task to get a date filter based on previous success date. If there
-    is no prev success date, today's date is returned. If arg_override is provided,
-    it will be returned instead of the date filter.
+    is no prev success date, 1970-01-01 is returned. If arg_override is provided,
+    it will be returned instead of the date filter arg.
 
     Args:
         arg_override (string): if provided, the arg string will be used instead of the date filter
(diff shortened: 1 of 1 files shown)


dags/atd_moped_components_to_agol.py
```diff
@@ -43,30 +45,94 @@
 }
 
 
+@task(
+    task_id="get_args",
+)
+def get_args(params, **context):
```
mddilley, line 50: This new helper is very specific to this DAG. We used to recycle `get_date_filter_arg` which was written for our `atd-knack-services` DAGs and is a better fit there since this one never used the `should_replace_monthly` option  it has.

chiaberry (approved):
This is great! I want to copy this pattern for other datasets that may benefit from an option to toggle a full replace
Screenshot 2024-08-28 at 11 53 02 AM

dags/atd_moped_components_to_agol.py
```diff
@@ -43,30 +45,94 @@
...
+        command=f"python components_to_agol.py {args}",
         environment=env_vars,
         tty=True,
         force_pull=True,
         mount_tmp_dir=False,
+        execution_timeout=duration(minutes=5),
     )
 
-    date_filter_arg >> t1
+    args >> branch >> [full, incremental]
```
frankhereford, line 137: Thank you for using meaningful names for the tasks instead of the `t1`, `t2`, etc that the old docs used. This is so much more readable.

dags/atd_moped_components_to_agol.py
```diff
@@ -4,7 +4,9 @@
 
 from airflow.models import DAG
 from airflow.operators.docker_operator import DockerOperator
-from pendulum import datetime, duration, now
+from airflow.decorators import task
+from airflow.models import Param
+from pendulum import datetime, duration, now, parse
```
frankhereford: looks like `now` isn't used, maybe it can go?

frankhereford: Similarly, while it's not changed as part of this PR, can you remove the import of `get_date_filter_arg` as a tag along change?

frankhereford (approved):
Thanks for this Mike, and I'm sorry it's hung out here so long in review! This worked great -- 

Well, the DAG worked great, but AGOL was sure slow today, so I had to extend the timeout to 20 minutes on my local copy to test. If that doesn't bother you and sounds OK -- then this is ready to ship!

dags/atd_moped_components_to_agol.py
mddilley: thanks for catching these, Frank! removing them now 🙏

mddilley:
going to merge this one! thank y'all for the reviews!!!

dags/atd_moped_components_to_agol.py
johnclary: late to the party, but it is very awesome to see this in action and know that we have a pattern for it now! 🚀

PR #249, opened 2024-08-28
Schedule Maximo DAG later

Charlie-Henry:
We were getting intermittent connecting to the DB failures on this DAG. I think I timed this poorly and am likely hitting the DB when it is being rebuilt every AM. 

## Testing

**Steps to test:**

---
#### Ship list
- [ ] Code reviewed 


johnclary (approved):
yep, this checks out: 6am is later than 4am 🚢

chiaberry (approved):
agree with john, 6 is after 4

PR #254, opened 2024-10-09
Work Zone Data Feed Airflow deployment

Charlie-Henry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/18131

## Associated repo
https://github.com/cityofaustin/dts-work-zone-data-feed

## Testing

**Steps to test:**
If you're NOT on arm64 (Apple silicon) you can ignore these two steps:
1. Comment out `force_pull=True` on line 100
2. `docker pull --platform linux/amd64 atddocker/dts-work-zone-data-feed:production`
☝️ I did try to google this but wasn't able to find where to add that platform arg when pulling in the DockerOperator?

```
docker compose run --rm airflow-cli dags test dts_work_zone_data_feed
```
Check that it completes: 
> **work_zone_data_publishing** ran successfully!

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/dts_work_zone_data_feed.py
+++ b/dags/dts_work_zone_data_feed.py
@@ -1,0 +1,106 @@
+# test locally with: docker compose run --rm airflow-cli dags test dts_work_zone_data_feed
+
+import os
+
+from airflow.decorators import task
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration, now
+
+from utils.onepassword import get_env_vars_task
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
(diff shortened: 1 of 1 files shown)


dags/dts_work_zone_data_feed.py
```diff
@@ -0,0 +1,106 @@
...
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 1, 1, tz="America/Chicago"),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "execution_timeout": duration(minutes=10),
```
johnclary, line 22: Just double-checking: this is enough time?

johnclary (approved):
it's passing the eyeball test for me 🚢

dags/dts_work_zone_data_feed.py
Charlie-Henry: I think so, usually takes me less than a minute to run. Also it'll be running hourly so we probably should put a cap on things to monitor it.

PR #257, opened 2024-11-12
Add retry behavior to VZ socrata export

johnclary:
## Associated issues
- https://austininnovation.slack.com/archives/CM9MK950S/p1731421398439679

## Associated repo
- Vision Zero

## Testing

🤞 

---
#### Ship list
- [ ] Code reviewed 

### DIFF
--- a/dags/vz_socrata_export.py
+++ b/dags/vz_socrata_export.py
@@ -74,7 +74,7 @@ DEFAULT_ARGS = {
     "depends_on_past": False,
     "email_on_failure": False,
     "email_on_retry": False,
-    "retries": 0,
+    "retries": 2,
     "execution_timeout": duration(minutes=60),
     "on_failure_callback": task_fail_slack_alert,
 }


chiaberry (approved):
i did a quick search to see if the retries could only retry the tasks that failed, since export_people doesn't depend on the outcome of export_crashes. I didnt find the answer to that, but I did see that theres a "retry_delay" [option ](https://airflow.apache.org/docs/apache-airflow/1.10.3/_api/airflow/operators/#package-contents), if the request is timing out, would giving it some time before rerunning be a good idea?

Charlie-Henry (approved):
why are we truncating the dataset daily and rebuilding it? This is caused us some headaches in the past whenever socrata bombs and we aren't able to rebuild the dataset.

Charlie-Henry:
The "new" dataset management API lets you push updates to a draft dataset then `apply` those edits. Maybe this is something to look into and let us should catch these errors before we publish them to the prod dataset?
 
https://github.com/socrata/socrata-py?tab=readme-ov-file#updating-a-dataset

mddilley (approved):
🚢 😎

mddilley:
I like Chia and Charlie's suggestions. I'm also wondering how [the recently discovered PG sequence behavior](https://austininnovation.slack.com/archives/CM9MK950S/p1730742492917289?thread_ts=1730740273.235829&cid=CM9MK950S) would affect any kind of incremental updates to this ETL. I guess that we could rely on the CRIS crash ids but I can't remember if there are any gotchas with them.

johnclary:
> The "new" dataset management API lets you push updates to a draft dataset then apply those edits.

That's a really get solution to this problem! I am a bit torn between learning this new API versus putting in the effort to make it possible to do incremental loading with this dataset. The blockers to incremental loading are:
- include deleted/temp in dataset and filter them out of the VZV, because incremental loading can't detect when a record needs to be removed from the target dataset
- set up cascading update triggers so that people + unit updates propagate to the crash-level. although we might be able to do this within the SQL view 🤔 

in any case, i will open an issue 🙏

johnclary:
@chiaberry thanks for the suggestion about retry today—TIL. looking at the docs, the default delay is apparently enabled at 5 minutes, so i think we're good?

PR #260, opened 2024-12-04
Update readme and update Airflow version

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/19797

## Associated repo
n/a

## Testing

**Steps to test:**
1. Follow the steps in the "Testing a new Airflow version" section in the diff
2. Also, verify the version of the `apache-airflow` library in use in the footer

---
#### Ship list
- [x] Code reviewed 
~- [ ] Add note to 1PW secrets moved to API vault and check for duplicates~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM apache/airflow:2.9.0
+FROM apache/airflow:2.10.3
 
 USER root
 RUN apt-get update
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ Follow these steps to update the Airflow docker step. Reasons for doing this inc
 - In the [docker-compose.yaml](./docker-compose.yaml), replace `image: atddocker/atd-airflow:production` with `build: .`
 - Build the Docker images locally:
 ```shell
-$ docker compose build
+$ docker compose build --no-cache
 ``` 
 - Bring up the services and check the logging for errors and see that everything runs as expected:
 ```shell
--- a/README.md
+++ b/README.md
@@ -14,17 +14,23 @@ The stack is composed of:
 
 ## Table of Contents
 
-- [Getting Started](#getting-started)
-  - [Developing a new DAG](#developing-a-new-dag)
-  - [Tags](#tags)
-  - [Moving to production](#moving-to-production)
-- [Utilities](#utilities)
-  - [1Password utility](#1password-utility)
-  - [Slack operator utility](#slack-operator-utility)
-- [Useful Commands](#useful-commands)
-- [Updating the stack](#updating-the-stack)
-- [HAProxy and SSL](#haproxy-and-ssl)
-  - [HAProxy operation](#haproxy-operation)
+- [DTS Airflow](#dts-airflow)
+  - [Table of Contents](#table-of-contents)
+  - [Getting Started](#getting-started)
+    - [Developing a new DAG](#developing-a-new-dag)
+    - [Tags](#tags)
+    - [Moving to production](#moving-to-production)
+  - [Utilities](#utilities)
+    - [1Password utility](#1password-utility)
+    - [Slack operator utility](#slack-operator-utility)
+  - [Useful Commands](#useful-commands)
+  - [Updating the stack](#updating-the-stack)
+    - [Update Process](#update-process)
+      - [Testing a new Airflow version](#testing-a-new-airflow-version)
+      - [Update the production stack after merge](#update-the-production-stack-after-merge)
+  - [HAProxy and SSL](#haproxy-and-ssl)
+    - [HAProxy operation](#haproxy-operation)
+  - [Ideas](#ideas)
 
 ## Getting Started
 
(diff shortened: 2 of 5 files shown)


mddilley:
@frankhereford I'm requesting your review first because I'm looking at this as a rough draft. Do the docs make sense? Am I missing something about the previous method of updating files on the server vs. pulling in changes from this repo?

mddilley:
Forgot to add to feel free to push changes to this branch! 🙏

frankhereford:
edit for posterity: Mike and i observed that the most current version of `apache-airflow` was not in use on the updated image. I ran down the rabbit hole to figure out why so that we could have a complete upgrade.

original message: @mddilley 👋 I have run down the problem. It was in the `pip` solver that has to figure out what actual set of libraries to run given the various requirements put upon it. For some reason beyond me, it decides to stick with older version of implied libraries instead of installing the latest and greatest. I think it's some combination of the usage of `--upgrade-strategy eager` when running `pip` or perhaps this alternative solver via `--use-feature=2020-resolver`. 

Anyway - this fixes it with the minimum knock on updates required. Additionally, major version bumps have occurred for the `docker` and `apache-airflow-providers-slack` libraries from the python library repository. We need to incorporate those, but as a major version bump, they will require some updates to our usage. 

It's my opinion that those major updates need to be addressed as part of this update, but we're a little too far into opinion territory for me to push changes to do so into this branch beyond the commented out lines in our `requirements.txt`. In the interest of maybe putting a pin in this issue as fully complete, I'm going to branch off yours and see if I can bring those libraries up to date. That way we can talk about it without me having muddied the water in your PR here. 

edit: ☝️ this actually turned out really fruitful. https://github.com/cityofaustin/atd-airflow/pull/261

Thanks!

dags/utils/slack_operator.py
```diff
@@ -67,7 +67,7 @@ def task_fail_slack_alert(context):
 def task_success_slack_alert(context):
     slack_webhook_token = BaseHook.get_connection(SLACK_CONN_ID).password
     slack_msg = """
-            :white_check_mark: Task Sucessfully Completed.
+            :white_check_mark: Task Successfully Completed.
```
frankhereford, line 63: Tag along spelling correction here

requirements.txt
```diff
@@ -1,3 +1,9 @@
+httpx>=0.25.0
```
frankhereford: This is an implied library shared between `apache-airflow-providers-slack` and some other library in use in the image. It's pinned here to force the `pip` solver to have a route to land on `apache-airflow` version `2.10.3`.

README.md
```diff
@@ -181,26 +187,39 @@ Follow these steps to update the Airflow docker step. Reasons for doing this inc
...
+#### Testing a new Airflow version
 
 - Read the "Significant Changes" sections of the Airflow release notes between the versions in question: https://github.com/apache/airflow/releases/
   - Apache Airflow is a very active project, and these release notes are pretty dense. Keeping a regular update cadence will be helpful to keep up the task of updating airflow from becoming an "information overload" job.
-- Snap a backup of the Airflow postgreSQL database
+- Create a local branch with the [Dockerfile](./Dockerfile) modified to the version you intend to test
+- In the [docker-compose.yaml](./docker-compose.yaml), replace `image: atddocker/atd-airflow:production` with `build: .`
+- Build the Docker images locally:
+```shell
+$ docker compose build
```
frankhereford: I think it's worth mentioning here the flag `--no-cache` to give people a route to build things anew, which is important for images that include `apt` commands in their Dockerfile.

frankhereford (approved):
Thanks for this Mike! I have added a handful of spelling errors to tag-along in your PR, and I have solved the `pip` solver mystery. I am eager to get your feedback, re this solution.

I am going to offer a PR off this PR tonight, hopefully, that allows for this update to bring a few of in-house required libraries up a major version to more fully complete the issue of this PR. I think this is right to do because this work is ultimately an upgrade task. I may not be successful, and I'll report back to you, @mddilley, on the outcome. Thanks!

README.md
```diff
@@ -14,17 +14,23 @@ The stack is composed of:
...
+    - [Tags](#tags)
+    - [Moving to production](#moving-to-production)
+  - [Utilities](#utilities)
+    - [1Password utility](#1password-utility)
+    - [Slack operator utility](#slack-operator-utility)
+  - [Useful Commands](#useful-commands)
+  - [Updating the stack](#updating-the-stack)
+    - [Update Process](#update-process)
+      - [Testing a new Airflow version](#testing-a-new-airflow-version)
+      - [Update the production stack after merge](#update-the-production-stack-after-merge)
```
frankhereford, line 30: Mike - thanks for this additional documentation. 🙌

requirements.txt
```diff
@@ -1,3 +1,9 @@
...
 onepasswordconnectsdk==1.*
+
+# the following two libraries have a major version bump; we need to update our usage, at least for the slack provider
```
mddilley: Thanks for figuring out that it is finally time to give the Slack provider usage some attention. 🙏 Pretty good that we made it all this way since it was written for v1! I'll take a look at that PR next.

Edit: just saw this again i lol'd because it looks like i'm thanking myself. Thanks, Frank!

README.md
mddilley: cool, thanks! i just added this.

requirements.txt
mddilley: this is interesting and unexpected, and I'm curious how we could find/debug if we run into a similar problem in the future. I wouldn't know where to start - do you have a brief version of how you made it to this solution?

frankhereford: TLDR: I broke the build process up into pieces, did parts of it manually, observed where it went sideways and gave it hints on what I wanted.

👇 Opinion time --

Very generally, we tend to be aggressive in how we pin versions in our `requirements.txt`, `packages.json`, etc. It's a double edged blade -- you get assurance that you won't get tagged by a breaking change, but it becomes easy to fall behind and slip out of that zone where you're running the same stuff as everyone else which tends to work together because they are temporally correlated. 

👇 Fact time --

Influenced by this opinion, I suspected it was something we had in our config, and to confirm that, I spun up our airflow image including overriding the `entrypoint` of the image so that it simply dropped me into a shell without anything else happening. I was able to observe that installed version of `apache-airflow` was `9.3.0`. 

So I built my own image of plain airflow without any of our extensions to the image, and it had the right version in it, so I then knew that it was us, which I had suspected re the opinion above. So, I disabled any of our python related extensions in the Dockerfile, spun up a "airflow + our extensions - python stuff" image and hand installed the packages, again working form a shell where I had again overloaded the entrypoint. 

It was through this that I was able to directly observe the `pip` solver going south on me, and then it was a matter of installing our packages one a time to see where it got hung up. Turns out that we have two libraries that we install that themselves need `httpx` and based on how they are pinned (or not) in their respective `requirements` files, I determined that it was the solver that wasn't doing what I expected. With that, I pinned this `httpx` where I wanted it, and then it solved "correctly."

I'm not sure if this is a useful description of a general solution for python library wackiness, but it is the solution for this particular issue, at least for me.

Charlie-Henry (approved):
Did we decide to remove the `$`'s from the README commands going forward? I always have to go back and delete those 🥴 

Upgrades worked for me!
Screenshot 2024-12-12 at 3 47 57 PM

mddilley:
@Charlie-Henry good call! i do remember that conversation. i just removed the new `$`s and a couple old ones too. Thanks!

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM apache/airflow:2.9.0
+FROM apache/airflow:2.10.3
```
johnclary, line 1: Do we want to consider pinning to minor release instead of patch release? so `apache/airflow:2.10`  instead of `apache/airflow:2.10.3`?

Edit: I see that the Airflow docker repository does not follow the tagging convention needed to enable this :/

requirements.txt
johnclary: Thanks for working through this, Frank. I'm not sure if I'm fully following the issues we're running into here, but fwiw I tried clearing the `requirements.txt` of all dependencies except `onepasswordconnectsdk`, and the image built normally.

When i run `docker run -it --rm apache/airflow:2.10.3 bash` and then use `pip freeze` to inspect the python env, I see:

```
apache-airflow==2.10.3
apache-airflow-providers-slack==8.9.1
docker==7.1.0
httpx==0.23.3
```

Sorry if this is already a known dead-end, but it looks to me like the Airflow base image may already have this problem solved 🤷

README.md
```diff
@@ -181,26 +187,39 @@ Follow these steps to update the Airflow docker step. Reasons for doing this inc
...
+- Create a local branch with the [Dockerfile](./Dockerfile) modified to the version you intend to test
+- In the [docker-compose.yaml](./docker-compose.yaml), replace `image: atddocker/atd-airflow:production` with `build: .`
+- Build the Docker images locally:
+```shell
+docker compose build --no-cache
+``` 
+- Bring up the services and check the logging for errors and see that everything runs as expected:
+```shell
+docker compose up
+```
```
johnclary, line 207: thanks for these docs! i know that we have at least one test DAG (the Slack notifier) and maybe others. Should we mention specific DAGs that can be tested?

requirements.txt
frankhereford: Interesting - thanks for testing this out. I'm seeing something similar. I took it on faith that since we had those extra libraries in our own `requirements.txt` that we used to extend the airflow image, we would continue to need them. I'm paring down our additional requirements, and we can let airflow manage this. Thanks again. https://github.com/cityofaustin/atd-airflow/pull/260/commits/1ebfd61b173cda584101a7ff8a0166f8a6433d98

johnclary: Same—maybe Airflow changed the default python packages that are bundled in the image. I doubt we would have set up that requirements file without good reason to do so.

README.md
mddilley: i think that is a great idea - i'll add the name of the DAG to this section

mateoclarke (approved):
👍

PR #261, opened 2024-12-10
New version of slack library & enhanced slack error message

frankhereford:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/19797 for the new slack library and https://github.com/cityofaustin/atd-data-tech/issues/20070 for the improved notifications.

I kinda got on a roll and thought I'd play around with the notifications while I was in there, so that's what's up with this unexpected addition to this PR. I think it's worth it though. Here's an example:

image

LMK if I need to split up this PR into two -- thank you!

## Testing

* Switch your local stack to local build mode and build
* Spin it all up
* Make sure that you have a slack integration in `Admin` => `Connections`, it's in 1pw
* Fire off the slack notification tester DAG
* Edit the slack notification test DAG itself and play with the `byline` key to edit the DAG error byline
* 👀 Try out the optional `icon` key for instant DAG visibility
* Observe the way identities are encoded for mentions
* Observe how it decodes cron schedules
  * `Every 15 minutes, between 09:00 AM and 04:59 PM, Monday through Friday, March through September` is rendered from `*/15 9-16 * 3-9 1-5` for example
* Consider how exception classes could be used to better categorize error modes

---
#### Ship list
- [ ] Code reviewed 


dags/utils/slack_operator.py
```diff
@@ -15,74 +18,125 @@ def get_central_time_exec_data(context):
...
         dag=context.get("task_instance").dag_id,
-        ti=context.get("task_instance"),
         exec_date=get_central_time_exec_data(context),
         log_url=context.get("task_instance").log_url,
     )
     failed_alert = SlackWebhookOperator(
         task_id="slack_critical_failure",
-        http_conn_id="slack",
-        webhook_token=slack_webhook_token,
+        slack_webhook_conn_id=SLACK_CONN_ID,
```
frankhereford, line 86: part of the major library version bump

dags/utils/slack_operator.py
```diff
@@ -15,74 +18,125 @@ def get_central_time_exec_data(context):
 
 
 def task_fail_slack_alert_critical(context):
-    slack_webhook_token = BaseHook.get_connection(SLACK_CONN_ID).password
```
frankhereford, line 18: part of the major library version bump

dags/utils/slack_operator.py
```diff
@@ -15,74 +18,125 @@ def get_central_time_exec_data(context):
...
         dag=context.get("task_instance").dag_id,
-        ti=context.get("task_instance"),
         exec_date=get_central_time_exec_data(context),
         log_url=context.get("task_instance").log_url,
     )
     success_alert = SlackWebhookOperator(
         task_id="slack_success",
-        http_conn_id="slack",
-        webhook_token=slack_webhook_token,
+        slack_webhook_conn_id=SLACK_CONN_ID,
```
frankhereford, line 171: part of the major library version bump

dags/utils/slack_operator.py
```diff
@@ -15,74 +18,125 @@ def get_central_time_exec_data(context):
...
+*Execution Time*: `{exec_date}`
+*Schedule*: `{schedule_description}`
+*Try Number*: `{try_number} of {max_tries}`
+*Duration*: `{duration} seconds`
+*Operator*: `{operator}`
+*Exception Type*: `{exception_type}`
+*Exception Message*: `{exception_message}`
+<{log_url}|*Log URL*>
+"""
+
```
frankhereford: I thought I'd take a stab in the dark at what an enhanced notification could look like. I used the o1 model to crank out some overly verbose boilerplate and trimmed and modified it to what I think we've been envisioning.

I have no attachment to any of this, but I thought it was interesting enough to bring up for PR review. If anyone wants to modify it, please feel free to commit on this branch, 💯

frankhereford:
@mddilley Mike, I requested your review first, and if you like any or all of this, we can go from there to bring it to the group for their approval. I'm pretty excited about the possibilities this demonstrates for what we can do with our slack notifications, even if the ultimate product we come up with looks entirely different than this.

dags/utils/slack_operator.py
```diff
@@ -15,74 +18,125 @@ def get_central_time_exec_data(context):
...
         task=context.get("task_instance").task_id,
         dag=context.get("task_instance").dag_id,
-        ti=context.get("task_instance"),
         exec_date=get_central_time_exec_data(context),
         log_url=context.get("task_instance").log_url,
     )
     failed_alert = SlackWebhookOperator(
         task_id="slack_critical_failure",
-        http_conn_id="slack",
-        webhook_token=slack_webhook_token,
```
mddilley, line 35: 🙌 🙏 Nice to address this warning that we see in prod right now: `Provide webhook_token as hook argument deprecated by security reason and will be removed in a future releases. Please specify it in Slack Webhook connection.`

dags/utils/slack_operator.py
```diff
@@ -15,74 +18,125 @@ def get_central_time_exec_data(context):
...
+
+    # Convert schedule_interval to human-readable format
+    if schedule_interval is None:
+        schedule_description = "None"
+    elif isinstance(schedule_interval, str):
+        try:
+            schedule_description = get_description(schedule_interval)
+        except Exception:
+            schedule_description = f"Cron expression: {schedule_interval}"
+    elif isinstance(schedule_interval, datetime.timedelta):
```
mddilley: Nice catch that schedules can come in [multiple data types](schedule_interval) 🙏 Might be good to extract this logic into a helper so it could have a docstring to go along with it.

dags/test_slack_notifier.py
```diff
@@ -32,6 +38,8 @@ def task_fail():
     tags=["slack"],
     catchup=False,
 ) as dag:
+    dag.importance = "Critical DAG, must fix ASAP. Call Bob A. and Alice B. for help. Also, <@UMS32US1E>"
```
mddilley: This is cool as a demo of how to extend the DAG object. I think that if we decide to implement something like this that we should explore a custom DAG object so we document how these extra attributes are used and include them in the DAG config. Not suggesting any further work here - I stashed your idea in https://github.com/cityofaustin/atd-data-tech/issues/20008

mddilley (commented):
Cool stuff with cool demos of what is possible! I think we should open up review to the rest of the team! I am concerned about the merge of this PR being held back by discussion so I think that we should move decisions/ideas into https://github.com/cityofaustin/atd-data-tech/issues/20008 if needed so we can get these important updates into production asap along with the schedule descriptions that we scoped already. 🚀

Charlie-Henry (approved):
@frankhereford I wonder if there's a way to to flag notifications that come from our local stack versus the production stack? Maybe we can use the environment var?

Task failed successfully!
https://austininnovation.slack.com/archives/C013G4RNJ01/p1734040744884339

frankhereford:
@Charlie-Henry wrote:

> I wonder if there's a way to to flag notifications that come from our local stack versus the production stack?

This is a solid idea, thank you. I'm going to make it print out the environment if it's not production.

dags/utils/slack_operator.py
frankhereford: Good idea - I totally agree. https://github.com/cityofaustin/atd-airflow/commit/cf088a7faa108b349f5fd3e357c613d8e600d091 & https://github.com/cityofaustin/atd-airflow/commit/0cf80bb563a855dc39a448512ee2fe56b3cce3ca

dags/utils/slack_operator.py
johnclary: ah, so we have `task_fail_slack_alert_critical` and `task_fail_slack_alert`, but we're not using `task_fail_slack_alert_critical` anywhere?

dags/utils/slack_operator.py
```diff
@@ -1,88 +1,171 @@
...
+
+def format_schedule(schedule_interval):
+    """
+    Convert a schedule interval to a human-readable format.
+
+    This function handles different types of schedule intervals:
+    - None: Returns "None" as the description.
+    - str: Assumes the string is a cron expression and tries to convert it to a human-readable format.
+           If conversion fails, it returns the cron expression as is.
+    - datetime.timedelta: Converts the timedelta to a human-readable string, breaking it down into days, hours, minutes, and seconds.
```
johnclary: awesome docstring 🙏 would you mind breaking these lines to keep them under 90ish chars?

dags/utils/slack_operator.py
```diff
@@ -1,88 +1,171 @@
...
+                part = f"{period_value} {period_name}{'s' if period_value > 1 else ''}"
+                parts.append(part)
+        schedule_description = "Every " + ", ".join(parts)
+    else:
+        schedule_description = str(schedule_interval)
+    return schedule_description
+
 
 def get_central_time_exec_data(context):
-    from pendulum import timezone
```
johnclary, line 10: my memory on this is pretty vague, but I think we stashed the `import` in here because it's good to be mindful of what libs are imported from within the `dags` directory, because Airflow actually runs all of the DAG code every time it checks for new dags. 

probably not a big deal—and we should have added a comment here in the first place!

dags/utils/slack_operator.py
```diff
@@ -1,88 +1,171 @@
...
+        schedule_interval: The schedule interval to format. It can be None, a string (cron expression), or a datetime.timedelta.
+
+    Returns:
+        str: A human-readable description of the schedule interval.
+    """
+    if schedule_interval is None:
+        schedule_description = "None"
+    elif isinstance(schedule_interval, str):
+        try:
+            schedule_description = get_description(schedule_interval)
```
johnclary, line 39: very cool!

johnclary (commented):
Awesome stuff, Frank! I can't wait for our new error messages. Thanks for bringing us these major improvements 🚀  🚢

dags/utils/slack_operator.py
frankhereford: I'm sure you're right - thank you for saying this! When I saw the import down inside the function, I didn't understand the implications of it being there, but this makes sense and is a TIL. I'll move this back down here, 100%.

frankhereford: https://github.com/cityofaustin/atd-airflow/pull/261/commits/39ac9a93e5b7564b15378769db1531abdea06146

dags/utils/slack_operator.py
frankhereford: For sure; I'll go 80 characters for the classic terminal fit. https://github.com/cityofaustin/atd-airflow/pull/261/commits/2f7c9bdd76bb26a93b93f8544a7eb8fec0f2b8e8

dags/utils/slack_operator.py
chiaberry: my memory @johnclary is that when airflow 2 was set up, the critical dag template was set up as an option/example for future us, if we felt like some dags needed a different type of alert

dags/test_slack_notifier.py
```diff
@@ -32,6 +38,8 @@ def task_fail():
     tags=["slack"],
     catchup=False,
 ) as dag:
+    dag.byline = "Critical DAG, must fix ASAP. Call Bob A. and Alice B. for help. Also, <@UMS32US1E>"
```
chiaberry, line 42: ok so this is tagging you in the slack message, how did you know `<@UMS32US1E>` means`@frank` ?

chiaberry: answering my own question to say its someones member ID! i see theres an issue for future refinements, but could you add a quick comment here saying something like tag users by member id?

frankhereford: I'm not sure if people will know to look here. I'll add something to the README though - does that sound ok?

chiaberry: I was thinking here because when reading the line of code I didn't know it was tagging you without having seen it in the airflow channel.
if there is going to be readme updates in a future branch, it can go there then. its a cool feature!

frankhereford: Ah, gotcha -- that isn't clear. I will add a comment about who that particular UID is, and I also added some docs here: https://github.com/cityofaustin/atd-airflow/pull/261/commits/2a10c45ff9d4a1e9fb689c490c4b8111215eb7dc.

frankhereford: Thanks Chia -- how about this: https://github.com/cityofaustin/atd-airflow/pull/261/commits/d0aa04ae2db29953851c3e27ea35bfece4343186

README.md
```diff
@@ -157,6 +157,8 @@ The Slack operator utility makes use of the integration between the Airflow and
 
 To configure the Slack operator in your local instance, from the Airflow UI go to **Admin** > **Connections** and choose **Slack API** as the **connection type**. You can find the remaining settings in 1Password under the **Airflow - Slack Bot** item.
 
+The stack offers the ability to define a `byline` in the Slack message. This, if defined, this message will be shown prominently in the slack error notification. You may notify individuals or channels using the slack Advanced Formatting addressing scheme documented [here](https://api.slack.com/reference/surfaces/formatting#advanced).
```
frankhereford, line 161: Real change here, the rest is linting

dags/utils/slack_operator.py
mddilley: +1 - we were lucky that the v1 Slack notifier code was able to be copy/pasted into the new v2 stack. I've never seen us use the critical alert either.

frankhereford:
I've extracted [the slack library update portions](https://github.com/cityofaustin/atd-airflow/pull/262) of this PR and sent that to Mike for approval to get it into his PR and from there, it'll find its way into `production`.

PR #262, opened 2024-12-16
Update slack notification utility library to use new major version of python library

frankhereford:
## Associated issues

This work is part of the upgrade to airflow in https://github.com/cityofaustin/atd-data-tech/issues/19797.

## Associated repo

`atd-airflow` itself

## Testing

Spin up the local stack and fire off the test slack notification. You will need a slack integration which you can find in 1PW.

---

#### Ship list
- [ ] Code reviewed 


mddilley (approved):
Tested and this looks great! 🙏

PR #263, opened 2024-12-19
Update the readme after I ran through the Airflow maintenance tasks

mddilley:
## Associated issues
No issue - these changes came from me walking through the readme with help from Frank, and us realizing it needed a few changes.

Frank helped with some initial editing and fixed outdated parts too, and I'm requesting more reviews so y'all know about these updated steps too.

## Associated repo
n/a

## Testing

**Steps to test:**
1. None - just reading! 🙏

---

#### Ship list

- [x] Code reviewed
- [ ] ~add note to 1PW secrets moved to API vault and check for duplicates~

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This repository hosts Data & Technology Service's [Airflow](https://airflow.apac
 
 The production Airflow instance is available at `https://airflow.austinmobility.io/`. It requires COA network access.
 
-Our Airflow instance is hosted on `atd-data03` at `/usr/airflow/atd-airflow`. Local development is available, and instructions are below.
+Our Airflow instance is hosted on `dts-int-data-p01` and can be found in `/srv/atd-airflow`. Local development is available, and instructions are below.
 
 The stack is composed of:
 
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ git pull;
 
 ```
 
-#### If there has been a change to the Airflow Docker image like when updating the Airflow verson:
+#### If there has been a change to the Airflow Docker image like when updating the Airflow version:
 
 The production Airflow deployment uses a second Docker compose file which provides haproxy configuration overrides. To start the production docker compose stack use you must load both files in order:
 
--- a/README.md
+++ b/README.md
@@ -222,19 +222,25 @@ Follow these steps to update the Airflow docker step. Reasons for doing this inc
 - Create a local branch with the [Dockerfile](./Dockerfile) modified to the version you intend to test
 - In the [docker-compose.yaml](./docker-compose.yaml), replace `image: atddocker/atd-airflow:production` with `build: .`
 - Build the Docker images locally:
+
 ```shell
 docker compose build --no-cache
-``` 
+```
+
 - Bring up the services and check the logging for errors and see that everything runs as expected:
+
 ```shell
 docker compose up
-``` 
+```
+
 - Check if you can reach the Airflow dashboard at `http://localhost:8080`
 - If any updates affect the Slack notifier, see the [instructions to test it](#slack-operator-utility)
 - Bring down the services:
+
 ```shell
 docker compose down
 ```
+
 - In the [docker-compose.yaml](./docker-compose.yaml), switch `build: .` back to `image: atddocker/atd-airflow:production`
 - Push your branch and create a PR for review
 - After approval, merge and update the stack using the instructions in the [Moving to production section](#moving-to-production)
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ The stack is composed of:
     - [Tags](#tags)
     - [Moving to production](#moving-to-production)
       - [If there has been a change to the code in the repo:](#if-there-has-been-a-change-to-the-code-in-the-repo)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -88,6 +89,8 @@ Never commit directly to the `production` branch. Commit your changes to a devel
 
 Once merged, you will need to connect to our production Airflow host on the COA network, then pull down your changes from Github. Airflow will automatically load any DAG changes within five minutes. Activate your DAG through the Airflow web interface at `https://airflow.austinmobility.io/`.
 
+#### If there has been a change to the code in the repo:
```
mddilley, line 92: After I wrote this, I realized there could be occasion where both a GitHub and Docker pulls are needed, and these steps have overlap for the two separate tasks. Maybe I am overthinking it? 🤷

README.md
```diff
@@ -100,16 +103,37 @@ cd /srv/atd-airflow;
...
+# dts-int-data-p01
+
+# become the superuser
+su -;
+
+# enter into the production airflow directory
+cd /srv/atd-airflow;
+
+# pull the fresh production image
+docker pull atddocker/atd-airflow:production
```
frankhereford: There couldn't be a less of a nothing request, but will you stick semicolons on lines of bash commands that are missing them? It'll do the right thing based on new lines, but should someone suck all the commands up into one line, which isn't uncommon, those semicolons break up the commands.

README.md
```diff
@@ -100,16 +103,37 @@ cd /srv/atd-airflow;
...
+
+# stop the Docker stack
+docker compose stop
+
+# start the Docker stack
 docker compose -f docker-compose.yaml -f docker-compose-production.yaml up -d
+
```
frankhereford, line 137: Maybe add a line here with the `docker compose logs -f` to tail the logs? `ctrl-c` breaks you out of that log tailing.

README.md
```diff
@@ -213,22 +237,7 @@ docker compose down
 ```
 - In the [docker-compose.yaml](./docker-compose.yaml), switch `build: .` back to `image: atddocker/atd-airflow:production`
 - Push your branch and create a PR for review
-
-#### Update the production stack after merge
```
frankhereford, line 217: 🧹 🧹🧹🧹

great call, this section was outdated and needed to go

README.md
```diff
@@ -20,14 +20,15 @@ The stack is composed of:
     - [Developing a new DAG](#developing-a-new-dag)
     - [Tags](#tags)
     - [Moving to production](#moving-to-production)
+      - [If there has been a change to the code in the repo:](#if-there-has-been-a-change-to-the-code-in-the-repo)
```
frankhereford, line 23: Like, last night, since we were doing something that was purely a stack update via the image, we could have left the changes in the repo un-pulled down onto `01`, but I think we'd just be leaving ourselves a tiny foot gun there .. I can't think of a situation where we'd want to update the image but leave outdated `Dockerfile` or compose files in production.

I guess what I'm getting at -- I think a `git pull` is prudent both when you're doing a "normal" update where you have changed or added a dag, and also we should include a `git pull` when the stack is being updated, just so that it's up to date with HEAD on the origin. 

Maybe change this line on 23 to read "Steps for all updates, both to DAGs and to the stack" or something like that?

README.md
frankhereford: This `docker pull` line does exactly what needs to be done, but what would you think of embracing the `compose` nature of the stack and doing a `docker compose pull;` instead? That way, should we do something wonky like rename the image or whatever, this documentation will stay evergreen and correct.

frankhereford (approved):
Thank you for this Mike!! I'm very grateful for you updating this documentation because it was getting old in the tooth and the portions you identified were totally out of date.

I've added a few ideas and a few suggestions, but as is where is, this is a marked improvement and is shippable. Thank you!

README.md
mddilley: yes! i will update to match the rest of the commands 🙏

README.md
mddilley: yeah! that was super helpful to watch as the stack was coming up. i'll add.

pull_request_template.md
frankhereford, line 1: Just linting here 🙏 🙏🙏🙏

README.md
frankhereford: Naw, you're totally on the right track. My take, and check my math for sure, is that a `git pull` is good in all updates and a `docker compose pull` should be done, in addition, when the image has been updated.

README.md
mddilley: ahh yes! that eliminates the weirdness that i was feeling about these two separate paths having some overlap that was only suggested. 🙏 i'll add `git pull` explicitly to the second section so it will always be in the steps that happen when the stack changes from maintenance or other needs.

README.md
mddilley: making that change too - makes sense to pull the image that we've defined 💯

README.md
mddilley: awesome - i added a `git pull` to the second set of steps too so it is done either way. 🙏

chiaberry (approved):
*chef's kiss*

mateoclarke (approved):
👍

PR #266, opened 2025-01-30
bump airflow patch version

frankhereford:
## Associated issues

This is part of the monthly upgrade routine, please see https://github.com/cityofaustin/atd-data-tech/issues/20105

## Testing

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed


chiaberry (approved):
we have built locally and tested the patch

PR #268, opened 2025-02-10
Enhance airflow error notifications in slack

frankhereford:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/20645

## Associated repo

Airflow itself

## Testing

* Spin up the local airflow stack and kick off `test_slack_notifier_development`.
* Do you see and approve of the message that popped up in slack?
* Do you approve of the additional documentation added to the [readme](https://github.com/cityofaustin/atd-airflow/tree/update-airflow-error-messages?tab=readme-ov-file#slack-operator-utility)?

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,26 @@ The Slack operator utility makes use of the integration between the Airflow and 
 
 To configure the Slack operator in your local instance, from the Airflow UI go to **Admin** > **Connections** and choose **Slack API** as the **connection type**. You can find the remaining settings in 1Password under the **Airflow - Slack Bot** item.
 
+#### Byline support
+
+The stack offers the ability to define a `byline` in the Slack message. This, if defined, this message will be shown prominently in the slack error notification. You may notify individuals or channels using the slack Advanced Formatting addressing scheme documented [here](https://api.slack.com/reference/surfaces/formatting#advanced). Very briefly, you can look for a user's member ID in their profile under the hamburger menu, and you can right click on a channel name above in the main frame of the client to get at a channel ID.
+
+#### Custom icon/emoji support
+
+Additionally, you may specify a slack emoji to replace the default `:red_circle:` emoji. This is done by specifying the `icon` property of the `dag` object. For example:
+
+```python
+with DAG(
+    dag_id=f"test_slack_notifier_{DEPLOYMENT_ENVIRONMENT}",
+    <snip>
+) as dag:
+    # @UMS32US1E is @frank, you can "Copy Member ID" from a member's profile, in the hamburger menu
+    dag.byline = "Example optional byline, which supports mentions: <@UMS32US1E>"
+    dag.icon = ":test_tube:"
+
+    <snip>
+```
+
 **To test the Slack operator locally**, see the DAG named `test_slack_notifier`.
 
 ## Useful Commands
--- a/dags/test_slack_notifier.py
+++ b/dags/test_slack_notifier.py
@@ -32,7 +32,9 @@ with DAG(
     tags=["slack"],
     catchup=False,
 ) as dag:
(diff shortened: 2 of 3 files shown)


johnclary (approved):
🤩 looks amazing — this is a major upgrade!

i actually just missed an Airflow notification today that Chia thankfully caught. i can't wait to get this into production.

thanks for the updated docs as well!
 🚢 🚢 🚢 🚢 🚢 🚢

README.md
```diff
@@ -188,6 +188,26 @@ The Slack operator utility makes use of the integration between the Airflow and
...
+#### Custom icon/emoji support
+
+Additionally, you may specify a slack emoji to replace the default `:red_circle:` emoji. This is done by specifying the `icon` property of the `dag` object. For example:
+
+```python
+with DAG(
+    dag_id=f"test_slack_notifier_{DEPLOYMENT_ENVIRONMENT}",
+    <snip>
+) as dag:
+    # @UMS32US1E is @frank, you can "Copy Member ID" from a member's profile, in the hamburger menu
```
mddilley, line 204: TIL! Thanks for this pointer to where this ID lives. 🙏

mddilley (approved):
Tested locally and I see the slick new test notification in Slack. The exception message is going to be so helpful. 🚢 🙌 😎

dags/test_slack_notifier.py
```diff
@@ -32,6 +32,10 @@ def task_fail():
     tags=["slack"],
     catchup=False,
 ) as dag:
+    # @UMS32US1E is @frank, you can "Copy Member ID" from a member's profile, in the hamburger menu
+    dag.byline = "Example optional byline, which supports mentions: <@UMS32US1E>"
```
Charlie-Henry: I do kinda wonder if we could store these member IDs in another config file? That way you could do something like this where it's a little clearer which ID belongs to who?
```
f"Example optional byline, which supports mentions: {slack_member_ids['Frank']}"
```

dags/utils/slack_operator.py
```diff
@@ -87,29 +93,55 @@ def task_fail_slack_alert_critical(context):
...
-            *Execution Time*: {exec_date}  
-            *Log URL*: {log_url} 
-            """.format(
-        task=context.get("task_instance").task_id,
-        dag=context.get("task_instance").dag_id,
-        schedule_description=schedule_description,
-        exec_date=get_central_time_exec_data(context),
-        log_url=context.get("task_instance").log_url,
-    )
+    byline = getattr(dag, "byline", None)
```
Charlie-Henry: I [tested this with no byline](https://austininnovation.slack.com/archives/C013G4RNJ01/p1739300608754489) and it adds this `None` at the top of the error message, replacing `None` with an empty string `""` should clean this up for the DAGs where we don't have a byline.

Charlie-Henry (approved):
Looks awesome, just a few little comments ty @frankhereford

dags/test_slack_notifier.py
frankhereford: Thanks for this suggestion. I have implemented it here: https://github.com/cityofaustin/atd-airflow/pull/268/commits/7e78a5e6b56a71b576e771b71d4550e36c610dd7. Initially, I was wary of this because it hard-codes in magic numbers that we have to maintain, but I can certainly see the ease of ID'ing folks making it worth it here. Thanks for the suggestion.

dags/utils/slack_operator.py
frankhereford: Great suggestion. https://github.com/cityofaustin/atd-airflow/pull/268/commits/8176eccf3827cdbdc292b6ed61cc5fddce8e8004

PR #270, opened 2025-02-19
Support python tracebacks which occur inside the docker container + refine whitespace in slack message

frankhereford:
# Enhance / fix Airflow Slack notifications

### TLDR

This updates airflow slack notifications to use the python exception we care about from inside docker.

Sorry it's long -- this describes the problem and the solution. 

## What am I up to?

Howdy - I had a couple of hours today before planning and my issues pretty wrapped up, so I put together this bug fix / enhancement PR that follows up on our recent work to improve slack notifications for airflow.

I sync'd with @mddilley and showed this to him when it was a bare-bones PR, and he gave me valuable feedback and suggested I open this up to everyone to take a look at. Thank you Mike 🙏 

## What problem are you fixing? What is the bug fix?

When I put together the enhanced slack notifications for airflow failures, I used the existing DAG that Mike originally developed for testing slack notifiers. However, this test DAG is a purely python DAG and it does not invoke a docker container like all of our "real" production DAGs do. 

The exception of the purely python DAG is accessable to airflow as an exception object, or some similar rich interface. However, docker containers only communicate to airflow in the form of the log on `stdout`/`stderr` of the container. Since we write almost all of our tooling in python, almost always the last thing that comes out of our worker containers is the /log/ of a python exception, and it's this one that we actually care about. 

If you examine what we have in production by looking at our slack notifications, you'll see `DockerContainerFailedException` and `Docker container failed: {'StatusCode': 1}` as generic errors on almost every single airflow notification.

We know a container failed -- not helpful. This PR will detect the real exception of interest and uses its type and message instead in the notification. That's the fix.

### OK, what's in here?

* A docker container in the form of a Dockerfile and a python script in the toolbox. It's a container that has one job -- to throw an exception inside the container emitting a traceback on `stdout` or `stderr`. It's available at `frankinaustin/pyfail:latest`, [here](https://hub.docker.com/r/frankinaustin/pyfail). 
  * We could move this to `atddocker` of course, but would we really want to?
* A update to the slack notification utility library to support using the type and message from `stdout` of the container, if one can be extracted.
* A new DAG which we can use for testing this work which more closely mirrors the normal pattern of DAG->docker that we use.

### How do you test this thing?

* Get up and running with the airflow local stack
* Make sure that you have a slack integration in Admin => Connections, it's in 1pw
* Run the new DAG `test_docker_failure`
* Check out the notification and in particular cross reference this to the log of the sole task of DAG you just ran. 
  * Consider the context of each python exception
* Consider: This is going to do a much better job of surfacing our varied exception types and more notably, our varied exception messages. Are you 👍 on our exception messages being in slack? As a long shot, could there be a secret in one?
    * I talked about this with Mike after he brought up the question and h/t @Charlie-Henry for posing it. I feel like it's very unlikely we'd emit a secret in an exception message, and I'd love to talk about that rational with anyone who wants to. It's a good question. 🙏 🤔


dags/test_docker_failure.py
frankhereford, line 1: TLDR: test dag that works like our docker-ized dags in production

New test DAG here. This one follows the same pattern as all our working DAGs, where they kick off a container using the `DockerOperator` task constructor. (base class? airflow operator? 🤷 whatever you want to call it on line 33)

This differs from the test slack notifier as that it's all contained within the python itself, no docker, and thus exceptions in this code are first class exceptions in airflow's python scope. 

This way, in this file, has the exception of interest inside the docker container, where airflow only really knows that it's /a/ docker error, but doesn't know anything else except that it can expect some sort of output on STDOUT from the container which it reports as the log.

dags/utils/slack_operator.py
```diff
@@ -110,6 +109,28 @@ def task_fail_slack_alert_critical(context):
...
+
+    # Find the last line (which usually contains the exception type and message)
+    last_line = traceback_text.strip().split("\n")[-1]
+
+    # Extract exception type and message
+    match = re.match(r"([\w.]+): (.*)", last_line)
+    if match:
+        return match.group(1), match.group(2)
+
+    return None, None
```
frankhereford, line 131: Since we only have the log that came out of the docker container and not the actual python context, we are going to work with what we have. This looks for the very reliably formatted traceback report, starting at the end of the log working back so we get the one that really killed us. It extracts out the type and message and returns them.

toolbox/pyfail/Dockerfile
frankhereford, line 1: simple docker container that runs a python script

toolbox/pyfail/fail_script.py
frankhereford, line 1: a python program that throws an exception, but does it a few functions deep so it's not super short and looks "normal"

dags/test_docker_failure.py
frankhereford, line 1: this is a test dag that works like our docker-ized dags in production

dags/test_docker_failure.py
```diff
@@ -0,0 +1,43 @@
...
+    schedule_interval=None,
+    tags=["repo:atd-airflow", "slack"],
+    catchup=False,
+) as dag:
+    dag.byline = f"{slack_member_ids['Frank']} test failure"
+    dag.icon = ":test_tube:"
+
+    t1 = DockerOperator(
+        task_id="docker_failure",
+        image="frankinaustin/pyfail:latest",
```
frankhereford: This published image is defined later in this PR. It uses a modern python image and extends it to run a python program, included, that throws an exception.

dags/utils/slack_operator.py
```diff
@@ -142,9 +170,7 @@ def task_fail_slack_alert(context):
 
     slack_msg = f"""
         {icon}{env_indicator} *Task failure* 
-
-        {byline}
-
+        {'\n\t\t' + byline if byline else ''}
```
frankhereford, line 173: This excessive whitespace was bugging me and this tightens that up.

dags/test_docker_failure.py
johnclary: it'd be really nice if we could find a way to keep this test DAG around without depending on your image or keeping an image in our own repository. what do you think about this trickery?!

```python
    t1 = DockerOperator(
        task_id="docker_failure",
        image="atddocker/atd-airflow:production",
        command="python -c \"raise Exception('This is a test exception')\"",
        auto_remove=True,
        tty=True,
        mount_tmp_dir=False,
    )
```

johnclary (approved):
I'm really excited about this! I'm expecting we might need to make minor adjustments to this as we observe real live errors. 

I offered a suggestion about using our existing Airflow image for the test DAG.

thanks for putting this together!

Charlie-Henry (approved):
Did not test this but I did see your test notifications in slack. I don't think it's a big deal if something gets leaked to slack as we can scrub it pretty easily. 

It might also just be good practice if we are passing secrets through URL params to do not do that 🤷.

dags/test_docker_failure.py
mddilley: I just tested John's snippet ([slack link](https://austininnovation.slack.com/archives/C013G4RNJ01/p1740766358856229)), and it worked great! It would be a nice way to avoid depending on another image.

chiaberry: I think this is neat trickery!!

dags/test_docker_failure.py
```diff
@@ -0,0 +1,43 @@
...
+
+with DAG(
+    dag_id=f"test_docker_failure",
+    description="Throws a python exception from within a docker container",
+    default_args=DEFAULT_ARGS,
+    schedule_interval=None,
+    tags=["repo:atd-airflow", "slack"],
+    catchup=False,
+) as dag:
+    dag.byline = f"{slack_member_ids['Frank']} test failure"
```
mddilley: I think that I'd feel better without the mention here so we don't 📣 you when testing.

mddilley (approved):
I tested this out, and I like it a lot. Thanks for making this upgrade to the alerts! 🐳 ℹ️ ✅

chiaberry (approved):
I agree with Mike, having your @ mention in there was nice for testing, but I dont think you need to be alerted *every* time someone uses this dag :)

dags/test_docker_failure.py
frankhereford: I hear that Mike, and I can take it out -- I guess the rational that I left it in there was a reference usage of the dict of names, but we have another real one out there now to look at in the cris import dag. I'll remove this one.

dags/test_docker_failure.py
frankhereford: Absolutely 100%, this is brilliant - thank you @johnclary - this is the route.

frankhereford: Much much better. 🙏  https://github.com/cityofaustin/atd-airflow/pull/270/commits/cbfdd611832c58ac9ffde2c7d7ccffda9607650a

dags/test_docker_failure.py
frankhereford: It's outta there! Thank you for this suggestion! https://github.com/cityofaustin/atd-airflow/pull/270/commits/d0e756138edbf276ffb3be67fbc4d55485946b23

dags/test_docker_failure.py
johnclary: nice!

PR #274, opened 2025-03-25
Update airflow docker compose file

frankhereford:
## Associated issues

tangential to https://github.com/cityofaustin/atd-data-tech/issues/21325 and also this [deep cut](https://austininnovation.slack.com/archives/CHZE6BC6L/p1741798346601379?thread_ts=1741787771.395849&cid=CHZE6BC6L). While we don't have a new airflow release to adopt, we have spoken over the last month about how we need to update our compose file to squash a number of errors we see on start-up in airflow.

## Associated repo

Airflow itself

## Wait? The redis image is not starting! Something about a RDB file?

Do a `docker compose down` to make sure that there are no ephemeral stores of the redis database hanging around. If you are a `docker compose stop` user, this may apply to you. 

## Testing

No particular thing to test, but spin it up locally and give the startup logs a once-over to see what you think. 

Does airflow work as expected for you?

Do you like the diff of the compose file?

---

#### Ship list

- [ ] Code reviewed
- [ ] ~Add note to 1PW secrets moved to API vault and check for duplicates~


chiaberry:
Does this PR get us closer to addressing https://github.com/cityofaustin/atd-data-tech/issues/21630, or does it basically accomplish the refresh? I think the only step missing would be to remove the CeleryExecutor

docker-compose.yaml
```diff
@@ -1,13 +1,15 @@
...
 x-airflow-common: &airflow-common
-  # In order to add custom dependencies or upgrade provider packages you can use your extended image.
-  # Comment the image line, place your Dockerfile in the directory where you placed the docker-compose.yaml
-  # and uncomment the "build" line below, Then run `docker compose build` to build the images.
-  # build: .
-  image: atddocker/atd-airflow:production
+  # the `image:` key here is used in production to pull our published, production airflow image.
+  # the `build:` key is used in development to build the image locally.
+  build: .
+  # image: atddocker/atd-airflow:production
```
frankhereford: `git` 😵‍💫

frankhereford:
> Does this PR get us closer to addressing [cityofaustin/atd-data-tech#21630](https://github.com/cityofaustin/atd-data-tech/issues/21630), or does it basically accomplish the refresh? I think the only step missing would be to remove the CeleryExecutor

Thanks @chiaberry for this great question. 

It does a large part of that issue, but does not do the crux of that issue. In the course of doing that issue, one would need to update the docker compose file to be up to date with what they offer in the docs and then embark on modifying it to remove the celery executor. 

This work is entirely in the scope of my assigned issue, and is something we've overlooked over the last few months -- resulting in our environment variables and other configuration parameters drifting away from the stack we were running. I called it tangential only because it's not in the same form of the work we've done previously, which was concerning the version of the airflow stack. This work is concerning the form of the airflow configuration.

chiaberry:
@frankhereford okay, just wondering if 21630 could be closed when this is merged or if that issue should stay open, and it sounds like it should stay open.

docker-compose.yaml
```diff
@@ -23,12 +25,28 @@ x-airflow-common: &airflow-common
...
+    # If you want to use it, out comment it and replace airflow.cfg with the name of your config file
+    # AIRFLOW_CONFIG: '/opt/airflow/config/airflow.cfg'
+
   volumes:
     - ${AIRFLOW_PROJ_DIR}/dags:/opt/airflow/dags
     - ${AIRFLOW_PROJ_DIR}/logs:/opt/airflow/logs
     - ${AIRFLOW_PROJ_DIR}/plugins:/opt/airflow/plugins
-    - ${AIRFLOW_PROJ_DIR}/config:/opt/airflow/config
-    - ${AIRFLOW_PROJ_DIR}/airflow.cfg:/opt/airflow/airflow.cfg
+    # - ${AIRFLOW_PROJ_DIR}/config:/opt/airflow/config
```
chiaberry: can this line be removed since its on L42 too?

chiaberry (approved):
spun up locally, signed back in and saw all the dags
I think this feels faster than before, but I could also be imagining that -- either way, 🚢 , it works!

johnclary (approved):
This spun up normally for me, although i did get a big wall of this error repeated ~300 times:

```
atd-airflow-airflow-webserver-1  | /home/airflow/.local/lib/python3.12/site-packages/azure/synapse/artifacts/models/_models_py3.py:68588 SyntaxWarning: invalid escape sequence '\d'
```

The error did not persist, so it doesn't seem like a major issue. 


Thanks for getting us updated. I'm strongly in favor of ripping celery out of our stack—looking forward to seeing that work come on deck 🚢

chiaberry:
hm weird that I didnt see that error. its related to a python version, right @frankhereford ?

frankhereford:
@johnclary wrote:

> This spun up normally for me, although i did get a big wall of this error repeated ~300 times:

That is actually a warning being emitted from python complaining about a now invalid escape sequence that python is slowly deprecating. Since 3.12, python complains about it much louder -- you'll see it all over our code bases, it's a clean-up we need to do and about a billion python libraries need to do. It'll be around a while, i'm afraid.

https://docs.python.org/3/whatsnew/3.12.html#other-language-changes

docker-compose.yaml
frankhereford: The comment on line 40-42 was from upstream, so I didn't really give it much heed -- i only left it in there to make it easier on the next person who is doing this diff operation by hand. edit: they are gone now.

I did strike line 48 though. as well, as it's a previous version of line 49, and it was clutter. 

Thanks!

docker-compose.yaml
mddilley: @frankhereford are you going to switch back to `image` here before merge?

frankhereford: Goodness, thank you Mike. yes!

docker-compose.yaml
```diff
@@ -23,12 +25,24 @@ x-airflow-common: &airflow-common
...
+    AIRFLOW__CELERY__RESULT_BACKEND: db+postgresql://airflow:airflow@postgres/airflow
+    AIRFLOW__CELERY__BROKER_URL: redis://:@redis:6379/0
+    AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION: "true"
+    AIRFLOW__CORE__LOAD_EXAMPLES: "false"
+    AIRFLOW__API__AUTH_BACKENDS: "airflow.api.auth.backend.basic_auth,airflow.api.auth.backend.session"
+
+    AIRFLOW__SCHEDULER__ENABLE_HEALTH_CHECK: "true"
+    # WARNING: Use _PIP_ADDITIONAL_REQUIREMENTS option ONLY for a quick checks
+    # for other purpose (development, test and especially production usage) build/extend Airflow image.
+    _PIP_ADDITIONAL_REQUIREMENTS: ${_PIP_ADDITIONAL_REQUIREMENTS:-}
```
mddilley, line 39: I was thrown off here for a second by the comment, but I think that I understand that this is [a place to set additional requirements if needed for testing](https://airflow.apache.org/docs/docker-stack/entrypoint.html#installing-additional-requirements). We aren't actually setting any in prod just like they recommend. TIL!

mddilley (approved):
I spun this up locally with success. ✅ Thanks for the refresh in here! 🙏 🚢

docker-compose.yaml
frankhereford: Yea, it's unusual. I think the idea behind this thing they have done is to think that it's pretty normal to need one or two extra libraries, and so they have this mechanism to install them at container start instead of forcing people to build their own images just to add one library. But -- it's on container start -- not image build -- so it's expensive and slow and repeats the install every time you start it up, ... like including when you fire up a docker image to run a DAG to test it using the CLI tooling.

PR #277, opened 2025-04-03
Change Moped x VZ schedule to daily

Charlie-Henry:
@johnclary @mddilley  I set it to 1AM daily, not sure if you have a preferred time for this that would best for VZ or moped? Maybe we should sync it up with the vz->socrata export?

## Testing

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/vz_moped_component_crashes.py
+++ b/dags/vz_moped_component_crashes.py
@@ -105,7 +105,7 @@ with DAG(
     dag_id="vz-moped-component-crashes",
     description="Populate the Moped - crash lookup table and publish the table to Socrata",
     default_args=DEFAULT_ARGS,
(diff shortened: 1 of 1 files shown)


chiaberry (approved):
👍🏼

johnclary (commented):
1am works for me — thanks for updating this!

johnclary (approved):

mddilley (approved):
Schedule works good for Moped. thanks for making this update. 🙏 🚢

frankhereford (approved):
Looks great Charlie, and I think the timing is fine. I pulled the data for the system's load over the last 24 hours, and the only time that sticks out to me as one we want to avoid is between 3AM and 4AM, and then between 7AM and 8AM. And even if you wanted to run it in those times, it wouldn't be the end of the world. Thanks!

image

PR #279, opened 2025-04-23
deleting long paused dags

Charlie-Henry:
Deleting some old DAGs that have been paused for a long time. Really no reason to keep them around cluttering things up. I think ideally paused DAGs should only be for temporary outages or for our test DAGs.

If you feel like any others should be added LMK!

## Testing

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed


frankhereford (approved):
# 🧹 🧼 ✌️  🌬️ ⛵

chiaberry (approved):
I like this, reduces the noise of paused dags in airflow

mddilley (approved):
I agree, and I'm all about reducing clutter and confusion. Thanks, Charlie!

PR #280, opened 2025-05-01
Update ETL for Traffic Detectors Socrata Dataset to refresh daily

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/22274

## Testing

**Steps to test:**'

You can run the dag locally, it now does a full replace on every run, should take under 4 minutes. 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/atd_knack_detectors.py
+++ b/dags/atd_knack_detectors.py
@@ -5,7 +5,6 @@ from airflow.operators.docker_operator import DockerOperator
 from pendulum import datetime, duration
 
 from utils.onepassword import get_env_vars_task
-from utils.knack import get_date_filter_arg
 from utils.slack_operator import task_fail_slack_alert
 
 DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT", "development")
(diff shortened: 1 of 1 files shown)


mddilley (approved):
Read the code, and the updates makes sense to me! 🙏 I didn't test locally since I can't reach the postgREST endpoints at the moment.

dags/atd_knack_detectors.py
```diff
@@ -65,24 +64,22 @@
     dag_id="atd_knack_detectors",
     description="Load detectors (view_1333) records from Knack to Postgrest to AGOL and Socrata",
     default_args=DEFAULT_ARGS,
-    schedule_interval="10 9 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    schedule_interval="10 4 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
roseeichelmann, line 67: hold my hand thru this cron syntax... run every day at 4:10 am?

dags/atd_knack_detectors.py
```diff
@@ -94,7 +91,7 @@
         image=docker_image,
         docker_conn_id="docker_default",
         auto_remove="force",
-        command=f"./atd-knack-services/services/records_to_socrata.py -a {app_name} -c {container} {date_filter_arg}",
+        command=f"./atd-knack-services/services/records_to_socrata.py -a {app_name} -c {container}",
```
roseeichelmann, line 94: so the date_filter_arg command causes it to not refresh daily? how often was it refreshing before?

dags/atd_knack_detectors.py
chiaberry: yes! I moved the time earlier because Christina / the MMC asked for it to happen overnight

dags/atd_knack_detectors.py
chiaberry: if a date_filter_arg is supplied to the scripts in knack services, it looks for records updated after the timestamp -- without a date_filter_arg the script does a full replace every time

previously it was doing a full replace on the first of the month, but records that were deleted (in knack) in the interim weren't being removed from the open data set. so there was a mismatch. now we are just doing a full replace every day.

johnclary (approved):
Haven't tested locally, but this looks great 🙏

frankhereford (approved):

PR #282, opened 2025-05-15
TDS cases to Socrata

Charlie-Henry:
Related PR: https://github.com/cityofaustin/dts-right-of-way-reporting/pull/7

## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/21684

## Associated repo
https://github.com/cityofaustin/dts-right-of-way-reporting

## Testing

**Steps to test:**
1. Connect to VPN
2. Checkout this branch
3. Trigger this DAG `dts_row_reporting`
4. Check that `tds_cases_s3` and `tds_cases_socrata` completes

---

#### Ship list

- [ ] Code reviewed


dags/dts_row_reporting.py
```diff
@@ -319,5 +319,33 @@ def get_dataset_id(env_vars):
...
+        tty=True,
+        mount_tmp_dir=False,
+        retries=3,
+        retry_delay=duration(seconds=60),
+        trigger_rule="all_done",
+    )
+
 
-    t1 >> t2 >> t3 >> t4 >> t5 >> t6 >> t7 >> t8 >> t9 >> t10 >> t11 >> t12 >> t13
+    t1 >> t2 >> t3 >> t4 >> t5 >> t6 >> t7 >> t8 >> t9 >> t10 >> t11 >> t12 >> t13 >> t14 >> t15
```
johnclary, line 351: wow—1st prize for longest DAG ever 🏆 

i see the duration is set to 30 minutes. Is that still enough headroom?

johnclary (approved):

dags/dts_row_reporting.py
mddilley: y'all - we might have some misconfigured DAGs. I was looking at recent run times of this DAG, and I found an outlier that ran 36 minutes and should have timed out. I checked the [DAG docs](https://airflow.apache.org/docs/apache-airflow/2.10.5/_api/airflow/models/dag/index.html), and I only see `dagrun_timeout` as a parameter. `execution_timeout` is in the [task docs](https://airflow.apache.org/docs/apache-airflow/2.10.5/core-concepts/tasks.html#timeouts) but we are using as a DAG param in a bunch of files.

Could someone double-check me on this? 🙏 I can create an issue to fix if needed.

Screenshot 2025-05-27 at 4 59 20 PM

mddilley (approved):
I tested in the related PR, and I read the task code here which looks great to me and matches the commands in the other PR. 🚢

chiaberry (approved):
i got my 18 green squares!

dags/dts_row_reporting.py
Charlie-Henry: Yeah I think you're right @mddilley that `execution_timeout` is a *task* level timeout. Also, looking at that particular run most of the time is queuing time. Maybe it was long because we were doing some upgrades to airflow or running something big  at the same time?

mddilley: thanks for checking that, @Charlie-Henry 🙏 , and yeah that was an unusual run and I think that 30 min seems plenty for this DAG even with the new tasks. 

I put https://github.com/cityofaustin/atd-data-tech/issues/22797 in the backlog for the parameter updates.

mddilley: actually, @frankhereford cracked the code [here](https://github.com/cityofaustin/atd-airflow/pull/285/files#r2122182621). So, it seems we would need to add `dagrun_timeout` on the DAG level to see the entire DAG timeout.

PR #284, opened 2025-05-29
set 311 reporting DAG schedule to run twice daily

Charlie-Henry:
We usually have a 24+ hour lag in data from these reports we pull from 311. I ran it this afternoon and found fresher data. I think it either takes a long time for these reports to run so we aren't catching it at 2AM or the time 311 sends new data is not set to midnight (which is what we were told).

Sets the schedule to run twice daily at 2:36 AM and 1:36PM. I will reflect a similar refresh time for[ Power BI](https://app.powerbigov.us/links/RqF4l_7Mgj?ctid=5c5e19f6-a6ab-4b45-b1d0-be4608a9a67f&pbi_source=linkShare) as well.

[slack thread](https://austininnovation.slack.com/archives/C050E76BUQH/p1748543712377749)

## Associated repo
https://github.com/cityofaustin/dts-311-reporting

## Testing

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/dts_csr_report_publishing.py
+++ b/dags/dts_csr_report_publishing.py
@@ -145,7 +145,7 @@ with DAG(
     dag_id="dts_csr_report_publishing",
     description="Downloads reports of 311 service requests for TPW and publishes it in a Socrata dataset.",
     default_args=default_args,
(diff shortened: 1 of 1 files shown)


chiaberry (approved):
double the fun

mddilley (approved):
🚢 🙏

PR #285, opened 2025-06-02
Airflow self-maintenance DAG to prune the DB and purge old logs

frankhereford:
## Associated issues

This PR aims to [double](https://media.licdn.com/dms/image/v2/C4E12AQGqU2GQU51ksg/article-cover_image-shrink_600_2000/article-cover_image-shrink_600_2000/0/1520220037113?e=2147483647&v=beta&t=O0TkCgiA2qsaLIrkNHRNuwolr4aBJSZO6xQzGhz3Stk) [whammy](https://www.youtube.com/watch?v=tMoL61B91fA) two issues:

* https://github.com/cityofaustin/atd-data-tech/issues/20644
* https://github.com/cityofaustin/atd-data-tech/issues/20629

## One PR? Two Issues?

Both issues call for closely related functionality, of which, one task should not be done without the other, generally. Additionally, the tasks called for in both issues are very IO intensive, and structuring them in a single DAG allows for that work to be serialized, which might as well be a requirement on our on-prem server. We bog down real-good under high IO.

## Associated repo

Airflow itself

## Testing

* Normal routine; spin up the stack and fire off the DAG.
  * There is no outside docker task or anything; everything is contained in this python file.
* When you start the DAG in the airflow UI, did you see the parameterization?
  * Do you like the defaults set in the DAG code, for when we turn it on and let it do its thing automagically?
  * Do you see that you're unable to specify an interval shorter than 15 days to clean out?
* Do you like the inline docstrings? 🤖 
* Do you mind that I do not constrain the execution time for this DAG? If one were to let a ton of logs accumulate, the execution time here is 1) long, and 2) unknown. 

### Practical Reality for Final Testing

I have no reason to believe that this DAG will have any problems when we install it on `01`, but this DAG is very meta and relies on the environment outside of the stack itself, such as on-disk file permissions, linux vs. mac docker services, and the very kernel itself. 

As a result, when this has the approvals to go in, I will test it myself on production prior to merging it in to see if there are any other wrinkles to iron out and do so.

I know it feels a little backwards -- approval then more testing, but i think that's OK since the final testing needs to happen in the production environment. 

Please let me know if you have any reservations about that -- thanks.

---

#### Ship list

- [ ] Code reviewed


dags/airflow_db_prune_logs_rotate.py
```diff
@@ -0,0 +1,140 @@
...
+from airflow.decorators import dag, task
+from airflow.models import Param
+from airflow.utils.log.logging_mixin import LoggingMixin
+
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = os.getenv("ENVIRONMENT")
+
+
+default_task_args = {
```
frankhereford, line 15: We generally call these `default_args` and I've always thought that they were default arguments to the `@dag` decorator .. not true. They are default task arguments. I renamed this symbol to reflect that a-ha moment.

dags/airflow_db_prune_logs_rotate.py
```diff
@@ -0,0 +1,140 @@
...
+    cmd = "find /opt/airflow/logs/ -type d -empty -delete"
+    logger = LoggingMixin().log
+    logger.info(f"Running command: {cmd}")
+    return cmd
+
+
+@dag(
+    dag_id="airflow_purge_logs_prune_database",
+    default_args=default_task_args,
+    schedule_interval="0 0 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
frankhereford, line 125: We have used 2AM in the past, and I shifted that to midnight because the unit we're defining our interval to purge back from is 'days', so why not use the cut-off between one day and the next.

Also, we don't have a whole bunch of stuff running then. I'd say, if not midnight, I'd pick 5AM because that is my mental model of when a day starts. 🤷

dags/airflow_db_prune_logs_rotate.py
```diff
@@ -0,0 +1,140 @@
...
+    tags=["repo:atd-airflow", "airflow", "maintenance"],
+    catchup=False,
+    params={"days_back_to_prune": Param(default=30, type="integer", minimum=15)},
```
frankhereford, line 128: `minimum=15`

You can't pick a number of days less that 15 for which you want to keep. The idea being that we always will have two weeks and change worth of logs, even if someone goes in there and cleans up everything they could.

dags/airflow_db_prune_logs_rotate.py
```diff
@@ -0,0 +1,140 @@
...
+    Generate the bash command to delete Airflow log files older than a given number of days.
+
+    Args:
+        day_interval (int): Number of days; logs older than this will be deleted.
+
+    Returns:
+        str: Bash command string.
+    """
+    cmd = (
+        f'find /opt/airflow/logs/ -type f -name "*.log" -mtime +{day_interval} -delete'
```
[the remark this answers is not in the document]
frankhereford: Denied. The path specified is actually drawn from our `docker-compose.yaml` file and it's right, and it'll stay right.

dags/airflow_db_prune_logs_rotate.py
```diff
@@ -0,0 +1,140 @@
...
+
+@task.bash(task_id="airflow_log_dir_cleanup")
+def log_dir_cleanup_bash() -> str:
+    """
+    Generate the bash command to delete empty directories in the Airflow logs directory.
+
+    Returns:
+        str: Bash command string.
+    """
+    cmd = "find /opt/airflow/logs/ -type d -empty -delete"
```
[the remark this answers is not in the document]
frankhereford: Denied also. Specified as part of our stack in our compose file.

dags/airflow_db_prune_logs_rotate.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    Retrieve the number of days to prune back from the Airflow database.
+
+    Args:
+        days_back_to_prune (int): Number of days to look back for pruning.
+
+    Returns:
+        int: The number of days to prune back.
+    """
+    prune_before_days = int(days_back_to_prune)
```
johnclary: i love that you added type hints 🥇

i am slightly thrown off that you're using `int()` on a param declared as `int`. maybe the input param should be type `str` and the function should have a `TypeError` handler for if the `int()` call raises? 

i'm also not sure when it makes sense to use a task versus a simple helper function in an Airflow DAG. this feels borderline to me? But I get that a separate task enables you to isolate errors down to this specific step.

dags/airflow_db_prune_logs_rotate.py
```diff
@@ -0,0 +1,140 @@
+import os
+import pendulum
```
johnclary: It'd be a nice touch to import the specific pendulum submodules here (or, for super-optimization, in the function that uses them), since these imports are constantly executed by the Airflow DAG importer service thingy.

I'm also noticing that 0% of our dags do this for the `os` import, where we could go with `from os import getenv` 😐

dags/airflow_db_prune_logs_rotate.py
johnclary: good call!

chiaberry (approved):
🚢 

I like that we can specify if we want to prune more than what is our default, test steps check out and the docstrings are ✨

dags/airflow_db_prune_logs_rotate.py
mddilley: +1 - i had some [confusion about this recently](https://github.com/cityofaustin/atd-airflow/pull/282#discussion_r2110381177)! Total TIL for me. 🙏

dags/airflow_db_prune_logs_rotate.py
```diff
@@ -0,0 +1,140 @@
...
+    (
+        prune_before_days
+        >> clean_before_timestamp
+        >> db_clean
+        >> log_file_cleanup
+        >> log_dir_cleanup
+    )
+
+
+dag = airflow_purge_logs_prune_database()
```
mddilley: I could be wrong but I'm not sure if the assignment to `dag` is needed here. It could be a convention that I don't know about. Either way, excited about the TaskFlow API in here. 🚀

dags/airflow_db_prune_logs_rotate.py
```diff
@@ -0,0 +1,140 @@
...
+    log_file_cleanup = log_file_cleanup_bash(prune_before_days)
+    log_dir_cleanup = log_dir_cleanup_bash()
+
+    (
+        prune_before_days
+        >> clean_before_timestamp
+        >> db_clean
+        >> log_file_cleanup
+        >> log_dir_cleanup
+    )
```
mddilley, line 149: This part of the DAG is meant to spread out these tasks, right? Might be worth a comment that this is on purpose. I went full-TaskFlow mode and thought it was unneeded but then I remembered your comment about some of this being I/O-intensive.

mddilley (approved):
I tested this out with the default parameter and a custom one too. Worked as expected each time, and I think the defaults and schedule are great. Thanks for packaging up a refresher on the TaskFlow API in this one too. 🙌

dags/airflow_db_prune_logs_rotate.py
```diff
@@ -0,0 +1,140 @@
...
+    Returns:
+        str: Bash command string.
+    """
+    cmd = "find /opt/airflow/logs/ -type d -empty -delete"
+    logger = LoggingMixin().log
+    logger.info(f"Running command: {cmd}")
+    return cmd
+
+
+@dag(
```
johnclary, line 122: you asked wether or not I think we should set a max duration for this DAG, and I think it would be good to do that and it set it to a value that would be cause for concern. an hour?

johnclary (approved):
Works like a charm! It's so awesome to have this in place. I left some minor feedback, which I hope is helpful. Thanks, Frank! 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

dags/airflow_db_prune_logs_rotate.py
frankhereford: That is a good call, thank you for framing it like that. I am completely convinced. 

Under this logic, I think it depends on if we're going to schedule to run on its own or leave it for ad-hoc usage. Depending which way we go, I'd say:

* 90 minutes if we leave it manual, which feels like cleaning out 6 months of logs to me, or
* 5 minutes if we turn it on the daily schedule coded into the DAG today

I'm going to code it up for 5 minutes in the morning and ask the group for our preference at the next opportune time.

edit: Or, spitballing, if I can, I'm only going to enforce the 5 minute limit when it's run via schedule and let it run to completion if it's kicked off manually.

dags/airflow_db_prune_logs_rotate.py
frankhereford: > i love that you added type hints 🥇
> 
> i am slightly thrown off that you're using `int()` on a param declared as `int`. maybe the input param should be type `str` and the function should have a `TypeError` handler for if the `int()` call raises?

These are good observations -- thank you.

This is me being slap-dash and moving on as soon as something seemed like it was working to my satisfaction. To me, I didn't baulk at it because `int()`ing an `int` is a non-op for free and I don't have a fantastic handle on how the typing system works in python. 

> i'm also not sure when it makes sense to use a task versus a simple helper function in an Airflow DAG. this feels borderline to me? But I get that a separate task enables you to isolate errors down to this specific step.

👋 This is going to sound hand-wavy because it totally is, but it's a mental model that has worked pretty good for me -- i'm furthest from an expert.

I think that airflow's `task()` decorator dramatically alters the function that it decorates. I think it causes a `@task()`d function to return special Airflow objects which encapsulate the returned values. I also think it intercepts and interprets that data coming in as arguments, if available. Lastly, I think it can figure out the DAG node graph by tracking functions' output objects and where that data gets fed into other functions' input arguments. 🔮 

I've found that feeding a regular function's return value, that is to say a fundamental data type of data -- I've had trouble with tasks kicking off and not receiving the input, or other strange behavior like that. As if the execution was out-of-order.

Like I said, super hand-wavy, I don't have firm examples, but I always seem to end up coming back to this "make all your DAG parts into a task" and "simple tasks are good tasks" and let airflow handle whatever dark magic it wants to handle in its own black box. It's doing a lot and these DAGs end up running in something that only is meant to "look" like normal python execution flow, but is actually much more complex under the hood. As if the "passing" of data was an actionable event in its own right, and airflow uses it.

🤷 I'm def rambling, but this is my best guess to date. I think the way it works is the way to go, but I'm open to experimentation and new data, 💯

dags/airflow_db_prune_logs_rotate.py
frankhereford: You're right on the money, both on your reasoning and that it needs some clarification. I wouldn't have remembered why that was done like that if you hadn't reminded me just now. Great call.

https://github.com/cityofaustin/atd-airflow/pull/285/commits/3245662260a7fbbe4dc634050d98eb3c60cfbaac

dags/airflow_db_prune_logs_rotate.py
frankhereford: good call, https://github.com/cityofaustin/atd-airflow/pull/285/commits/929f7522519497301ad635e52ccf3f9ba693828a, I left it there thinking about how there is this object we get back that I think we can interact with, but we just never do. 

But yes - totally unused in this context - it's out of there

dags/airflow_db_prune_logs_rotate.py
frankhereford: It sure would, you're right. Boom: https://github.com/cityofaustin/atd-airflow/pull/285/commits/c5f6a57681e94b709558e5bc807d5a5938ee2000.

frankhereford: 🛸 @johnclary As a thought experiment, check this out -- it's like we're living in the Jetsons.  

https://github.com/cityofaustin/atd-airflow/pull/289

🤖 This took 30 seconds to write the prompt (seen in the PR) & maybe 20-25 minutes to run. `gemini-2.5-pro`, thinking model. One big agentic pass over it. I didn't do anything, this is how it came out of the machine. 🪄

I'm not advocating that we should pursue merging this, but I could easily be convinced to. It's zero shot ... but it compiles. I have this huge artifact where it explained its thought process for each DAG, it's dumbly transparent. 

At this point, testing is the bulk of the work. Half-jokingly, maybe we can write test fixtures with the robots? 

# 👍 👎 ⚖️ 🤔 

It's an interesting thought experiment.

dags/airflow_db_prune_logs_rotate.py
johnclary: > I've found that feeding a regular function's return value, that is to say a fundamental data type of data -- I've had trouble with tasks kicking off and not receiving the input, or other strange behavior like that

Same here—thanks for explaining that thinking. sounds good to me to leave it as-is 👍 

To clarify my point about the typing, the issue is that `days_back_to_prune` is actually passed to this function as a `str`. Here's a printout from `type(days_back_to_prune)`:

```
[2025-06-10, 13:28:57 UTC] {logging_mixin.py:190} INFO - Type of days_back_to_prune is: <class 'str'>
```

So it seems that although you define this as an `int` in the [parameter definition ](https://github.com/cityofaustin/atd-airflow/blob/c5f6a57681e94b709558e5bc807d5a5938ee2000/dags/airflow_db_prune_logs_rotate.py#L121), that simply determines the Airflow UI input for this param. It is provided to the DAG as a string. Which is to say that `int(days_back_to_prune)` is not a non-op, it's coercing a string to text.

dags/airflow_db_prune_logs_rotate.py
johnclary: I think setting to a shorter limit is good. If we somehow accumulate more logs that can be run on our regular schedule, that's probably a problem that will take fixing outside of this dag 🤷

dags/airflow_db_prune_logs_rotate.py
johnclary: Neat! Pretty nice to have a second hand for this kind of a task. I'm happy to review that!

dags/airflow_db_prune_logs_rotate.py
frankhereford: https://github.com/cityofaustin/atd-airflow/pull/285/commits/0e5e2708e84cf368c3ef3cbf16f3fc002e28bc64

Bummer, I didn't see a obvious method to control Airflow from the DAG that would allow the `dagrun_timeout` to be set conditionally, so I set it at 10 minutes. I agree, a shorter interval is better, and we can adjust once we get more experience with it.

Thank you for your suggestion 🙏

johnclary: 10 min sounds good!!

dags/airflow_db_prune_logs_rotate.py
frankhereford: https://github.com/cityofaustin/atd-airflow/pull/285/commits/1f87f73a2ca7ecdf903dc33cb8d794ad138b7eb5

💡Thank you for helping me see what is going on. I had mentally latched onto parameter definition - you're so right.  

I updated the task to `try`/`except` to catch a value error on the int() and stop execution. Thanks!

johnclary: Looks great—thank you!!

PR #287, opened 2025-06-06
development services DAG

Charlie-Henry:
I went with a little different of a DAG template here as there are so many very similar tasks repeated over and over. Idk if it makes things easier to read or not? 

## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/21684

## Associated repo
https://github.com/cityofaustin/atd-knack-services/

## Testing

**Steps to test:**
1. Trigger this DAG! You do have to be on VPN or have your IP whitelisted for postgrest
2. Check if it completes
3. check if my 1pass item for "Knack Development Services (TDS)" passes the vibe test
---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/atd_knack_development_services.py
+++ b/dags/atd_knack_development_services.py
@@ -77,7 +77,7 @@ with DAG(
     description="Downloads Knack data for several objects then publishes it to socrata datasets.",
     default_args=DEFAULT_ARGS,
     schedule_interval="0 2 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
-    tags=["repo:atd-knack-services", "knack", "socrata", "tds", "development services"],
+    tags=["repo:atd-knack-services", "knack", "socrata", "tds", "development-services"],
     catchup=False,
 ) as dag:
     docker_image = "atddocker/atd-knack-services:production"
@@ -86,7 +86,7 @@ with DAG(
 
     app_name = "development-services"
 
-    commmands = [
+    commands = [
         {
             "task_id": "tia_mitigations_to_postgrest",
             "command": f"./atd-knack-services/services/records_to_postgrest.py -a development-services -c view_2814 {date_filter_arg}",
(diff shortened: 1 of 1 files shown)


dags/atd_knack_development_services.py
```diff
@@ -0,0 +1,182 @@
...
+        trigger_rule="all_done",
+    )
+
+
+with DAG(
+    dag_id="atd_knack_development_services",
+    description="Downloads Knack data for several objects then publishes it to socrata datasets.",
+    default_args=DEFAULT_ARGS,
+    schedule_interval="0 2 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    tags=["repo:atd-knack-services", "knack", "socrata", "tds", "development services"],
```
frankhereford: TBH, I like the white space in the `development services` tag, and it seems to work fine in the UI, but it may be worth it to comply with the existing pattern in case we ever want to export these tags or use them in some outside context. Not a big deal - please disregard if you want.

dags/atd_knack_development_services.py
```diff
@@ -0,0 +1,182 @@
...
+    catchup=False,
+) as dag:
+    docker_image = "atddocker/atd-knack-services:production"
+    env_vars = get_env_vars_task(REQUIRED_SECRETS)
+    date_filter_arg = get_date_filter_arg(should_replace_monthly=True)
+
+    app_name = "development-services"
+
+    commmands = [
```
frankhereford: Typo here -- so many commands in here it snuck in a third m.

My main point though -- this pattern here where you list all these up and then `chain()` them in sequence is so crisp. It reads easily, easy to expand or work on, it's real fine.

frankhereford (approved):
Looks great Charlie, it's neat to see a new pattern in here. I pointed out one typo, and other than that, this is gold. Thanks!

chiaberry (approved):
entry in 1pw passes vibe test, I think this way of writing the dag with so many tasks is clean, but then its harder to see that this should [get new first place for longest DAG](https://github.com/cityofaustin/atd-airflow/pull/282#discussion_r2109932676)

Charlie-Henry:
> entry in 1pw passes vibe test, I think this way of writing the dag with so many tasks is clean, but then its harder to see that this should [get new first place for longest DAG](https://github.com/cityofaustin/atd-airflow/pull/282#discussion_r2109932676)

maybe a refactor is in order for that one too 😅

PR #293, opened 2025-06-30
Mention Chia when the dag fails and add execution timeout

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/23342 

and slack thread: https://austininnovation.slack.com/archives/CHZE6BC6L/p1751299211984369

### DIFF
--- a/dags/atd_knack_banner.py
+++ b/dags/atd_knack_banner.py
@@ -2,11 +2,10 @@ from os import getenv
 
 from airflow.models import DAG
 from airflow.operators.docker_operator import DockerOperator
-from airflow.decorators import task
 from pendulum import datetime, duration
 
 from utils.onepassword import get_env_vars_task
-from utils.slack_operator import task_fail_slack_alert
+from utils.slack_operator import task_fail_slack_alert, slack_member_ids
 
 DEPLOYMENT_ENVIRONMENT = getenv("ENVIRONMENT", "development")
 
@@ -17,7 +16,7 @@ DEFAULT_ARGS = {
     "email_on_failure": False,
     "email_on_retry": False,
     "retries": 0,
-    "retry_delay": duration(minutes=5),
+    "execution_timeout": duration(minutes=30),
     "on_failure_callback": task_fail_slack_alert,
 }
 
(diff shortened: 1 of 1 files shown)


dags/atd_knack_banner.py
```diff
@@ -2,11 +2,10 @@
 
 from airflow.models import DAG
 from airflow.operators.docker_operator import DockerOperator
-from airflow.decorators import task
```
chiaberry, line 5: this wasnt being used

dags/atd_knack_banner.py
```diff
@@ -48,10 +47,11 @@
     default_args=DEFAULT_ARGS,
     schedule_interval="45 7 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
     dagrun_timeout=duration(minutes=30),
-    tags=["repo:atd-knack-banner", "knack"],
+    tags=["repo:atd-knack-banner", "knack", "hr"],
     catchup=False,
 ) as dag:
     docker_image = f"atddocker/atd-knack-banner:production"
+    dag.byline = f"{slack_member_ids['Chia']}"
```
chiaberry, line 54: 📣

Charlie-Henry (approved):
image

dags/atd_knack_banner.py
```diff
@@ -17,7 +16,7 @@
     "email_on_failure": False,
     "email_on_retry": False,
     "retries": 0,
-    "retry_delay": duration(minutes=5),
+    "execution_timeout": duration(minutes=30),
```
mddilley, line 19: Should this be longer if the whole DAG was timing out in 30 minutes? Ignore me if that was related to a bug that was making the runs longer than normal.

btw I found this little gem in the [Airflow v3 docs](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/dag/index.html) which is not in our current version docs. 🫠 Doesn't seem like it would be new behavior so probably them just improving their docs 🚧 

> dagrun_timeout – Specify the duration a DagRun should be allowed to run before it times out or fails. Task instances that are running when a DagRun is timed out will be marked as **skipped**.

mddilley (approved):
Looks great - good to see that this will not fail silently anymore! 🙌 📣

dags/atd_knack_banner.py
chiaberry: I was thinking about that, usually it shouldnt take more than a few minutes, but this is timing out because its trying to replace every record. I decided to keep it at 30 because if its taking longer than 30 that should mean something unexpected is happening

PR #300, opened 2025-07-24
📢  Stacked exception enunciator for airflow slack integration

### LINKED ISSUE #22992: Enhance airflow exception enunciator to handle chained python exceptions
In a recent slack conversation, [here](https://austininnovation.slack.com/archives/CHZE6BC6L/p1749226876196089?thread_ts=1749179975.812389&cid=CHZE6BC6L), @johnclary observed that it would be good to make our airflow error slack system catch a particular error related to a timeout. 

I poked around in the airflow utility we have, and I was reminded that it parses up from the bottom until it finds an exception, which it reports.

Upon looking at the actual logs for the error, I observed that what is going on is an exception is being thrown, and then:

```
During handling of the above exception, another exception occurred:
```

From there, it prints out the second exception. It's the second exception that is picked up by the airflow slack utility, masking the first, more important exception. 

[I proposed](https://austininnovation.slack.com/archives/CHZE6BC6L/p1749228117128269?thread_ts=1749179975.812389&cid=CHZE6BC6L) that we compile a list of tuples representing all the exemptions found in the log, not just the last one. This would be something like  `[(ExcemptionType, ExcemptionMessage), ...]`. Then pass this list into the slack message creator, and enhance it to print out the list in slack / human friendly way, and we'll be able to see the actual root problem in many of our ETLs.

> This would give us observability “through” all the docker.errors.NotFound: 404 Client Error for http+docker://localhost/v1.49/containers/d4a77f6b787e74e7b666c85167dcaee9ddfa2933460685edfa324ee8a348838f/stop: Not Found ("No such container: d4a77f6b787e74e7b666c85167dcaee9ddfa2933460685edfa324ee8a348838f") we get, i think

I'd venture to say this is a large `2`, small `3`. 

cc @mddilley

frankhereford:
# Associated issues

This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/22992.

# Associated repo

Airflow itself

# Testing

### Testing this is a real ~time~ mind warp. It's just a jump to the left...

Here's the problem: It's very challenging to test this PR in a single pass because you can either be testing the parsing capabilities by artificially injecting crafted logs to simulate one of the practically infinite ways in which an ETL can fail, or you can craft fake ETLs that fail in very predictable, uninteresting, non-varying ways. You either give up simulating the event's occurrence or you give up simulating the event's output.

So we're going to do both, and I'm going to propose a final stage of testing after approvals are gathered.

### First, let's try out a hand-crafted DAG that fails in a predictable way:

* Spin up your local stack and get up to date with this branch. 
* Fire up the test fixture DAG `test_docker_failure`. This one has been enhanced to fire up a docker container that fails in such a way it emits a 3-layer-deep stacked exception. 
* Take a look at the code in this DAG and the error in Airflow and the way it's reported in slack. 
* Think about the stack and how it has failed. Think about what errors happened inside of docker (from airflow's perspective) and which didn't.
* Do you like how this is reported in slack? Can you see what exceptions came out of our code and which came out of airflow's code?

### Second, let's take a whirl at putting some real world failures into our failure parsing gizmo:

* Get on slack and find a real error that is complicated, or not. You can pick any error. This is testing the real world robustness of the parser and how it handles whatever wacky thing we can dig up. 
  * Understand that this is a system that is entirely based on heuristics and it is always, absolutely, going to be be possible to craft up an error message that fails as it goes through the parser. This is about making one that is "good enough", not one that is perfect. 
    * I do think this one is pretty good though; I think it rides the fine line between simple and comprehensive well. 
* Follow the "View Task Log" link from slack, and you'll be dropped on the log page of the particular task in the particular DAG that you picked. Look in this log and copy the entire contents out of the log section between the following lines. Do not include the following lines. We're looking for the logging that *comes out of our ETL* and not out of the airflow system or the airflow UI.

Between this line:
```
[2025-07-23, 13:41:16 CDT] {docker.py:345} INFO - Starting docker container from image atddocker/some-dts-thing:production
```

and the line:

```
[2025-07-23, 13:41:29 CDT] {taskinstance.py:1226} INFO - Marking task as FAILED. dag_id=whatever-dag, task_id=whatever-task, run_id=scheduled__2025-07-23T16:40:00+00:00, execution_date=20250723T164000, start_date=20250723T184115, end_date=20250723T184129
```
* Now that you've got that copied, we need to remove all the logging cruft that airflow has decorated the output with. Go to your favorite LLM provider and give it the log and ask it to do the following:

> Remove timestamps, file/line references, and log level prefixes from this log output and return as clean text in a code block.

* Now that you have a cleaned up log, which represents the actual, literal bytes of text that came out of the docker container's `STDOUT`, as it would be presented to the slack operator's code, replace the `None` with a multiline (`"""` ... `"""`) assignment into `DEBUG_LOG_TEXT` variable at the top of the `slack_operator.py` file. 
* Fire off the same test DAG from above, but this time, you should see your hand-picked error represented in slack, as the operator will use what you provided in the debug variable instead of what actually came out of the docker container. 
  * **NB**: your UI in airflow will still be the actual error from docker, only the *slack* warning message will represent the error you prepared.
* Did it work? Can you break the parser? 
* Review the code and consider the methodology
  * Do you have any ideas on how this could be done better or more robustly? 

#### You made it to the end!
# 🙌

### DIFF
--- a/dags/test_docker_failure.py
+++ b/dags/test_docker_failure.py
@@ -21,19 +21,52 @@ DEFAULT_ARGS = {
 
 with DAG(
     dag_id=f"test_docker_failure",
-    description="Throws a python exception from within a docker container",
+    description="Throws stacked python exceptions from within a docker container",
     default_args=DEFAULT_ARGS,
     schedule_interval=None,
     tags=["repo:atd-airflow", "slack"],
     catchup=False,
 ) as dag:
-    dag.byline = f"Test failure in a docker container"
+    dag.byline = f"Test stacked exceptions in a docker container"
     dag.icon = ":test_tube:"
 
     t1 = DockerOperator(
         task_id="docker_failure",
         image="atddocker/atd-airflow:production",
-        command="python -c \"raise Exception('This is a test exception')\"",
+        command=[
+            "python",
+            "-c",
+            """
+import sys
+import traceback
+
+def outer_function():
+    try:
+        middle_function()
+    except Exception as e:
+        print("Caught exception in outer_function: " + str(e))
+        raise RuntimeError("Failed in outer function") from e
+
+def middle_function():
+    try:
+        inner_function()
+    except Exception as e:
+        print("Caught exception in middle_function: " + str(e))
+        raise ValueError("Failed in middle function") from e
+
+def inner_function():
+    print("About to raise ConnectionError")
+    raise ConnectionError("Connection failed")
+
+if __name__ == "__main__":
+    try:
+        outer_function()
+    except Exception as e:
+        print("Final exception caught at top level:")
+        traceback.print_exc()
+        sys.exit(1)
+""",
+        ],
         docker_conn_id="docker_default",
         auto_remove="force",
         tty=True,
--- a/dags/utils/slack_operator.py
+++ b/dags/utils/slack_operator.py
@@ -276,11 +276,6 @@ def task_fail_slack_alert(context):
     if not airflow_already_found:
         all_exceptions.append(airflow_exception)
 
-    # Use the last exception as primary for backward compatibility
-    if all_exceptions:
-        exception_type = all_exceptions[-1][0]
-        exception_message = all_exceptions[-1][1]
-
     # Extract additional information
     dag = context.get("dag")
     dag_id = task_instance.dag_id
@@ -305,6 +300,8 @@ def task_fail_slack_alert(context):
     exceptions_text = ""
     if len(all_exceptions) == 1:
         # Single exception - use original format with source
+        exception_type = all_exceptions[-1][0]
+        exception_message = all_exceptions[-1][1]
         source = all_exceptions[0][2] if len(all_exceptions[0]) > 2 else "Unknown"
         exceptions_text = f"*Exception Type*: `{exception_type}` _(from {source})_\n        *Exception Message*: `{exception_message}`"
     else:
--- a/dags/utils/slack_operator.py
+++ b/dags/utils/slack_operator.py
@@ -109,6 +109,41 @@ def extract_exception_from_log(log_text):
     return None, None
 
 
+def extract_all_exceptions_from_log(log_text):
+    import re
+
+    # Find all occurrences of 'Traceback (most recent call last):'
+    exceptions = []
+    traceback_pattern = "Traceback (most recent call last):"
+    
+    start_pos = 0
+    while True:
+        traceback_start = log_text.find(traceback_pattern, start_pos)
+        if traceback_start == -1:
+            break
+        
+        # Find the end of this traceback by looking for the next traceback or end of text
+        next_traceback = log_text.find(traceback_pattern, traceback_start + len(traceback_pattern))
+        if next_traceback == -1:
+            traceback_text = log_text[traceback_start:]
+        else:
+            traceback_text = log_text[traceback_start:next_traceback]
+        
+        # Find the last line of this traceback (which contains the exception)
+        lines = traceback_text.strip().split("\n")
+        if len(lines) > 1:
+            last_line = lines[-1]
+            
+            # Extract exception type and message
+            match = re.match(r"([\w.]+): (.*)", last_line)
+            if match:
+                exceptions.append((match.group(1), match.group(2)))
+        
+        start_pos = traceback_start + len(traceback_pattern)
+    
+    return exceptions if exceptions else [(None, None)]
+
+
 def task_fail_slack_alert(context):
 
     task_instance = context.get("task_instance")
@@ -119,14 +154,23 @@ def task_fail_slack_alert(context):
         str(exception) if exception else "No exception message available"
     )
 
+    # Extract all exceptions from logs if available
+    all_exceptions = []
     if exception and hasattr(exception, "logs") and exception.logs:
         logs = exception.logs
-        parsed_exception_type, parsed_exception_message = extract_exception_from_log(
-            "\n".join(logs)
-        )
-        if parsed_exception_message and parsed_exception_type:
-            exception_type = parsed_exception_type
-            exception_message = parsed_exception_message
+        all_exceptions = extract_all_exceptions_from_log("\n".join(logs))
+        
+        # If we found exceptions in logs, use them; otherwise keep the original exception
+        if all_exceptions and all_exceptions[0][0] is not None:
+            # Use the last exception as the primary one for backward compatibility
+            exception_type = all_exceptions[-1][0]
+            exception_message = all_exceptions[-1][1]
+        else:
+            # Fallback to original exception if no parsed exceptions found
+            all_exceptions = [(exception_type, exception_message)]
+    else:
+        # No logs available, use the original exception
+        all_exceptions = [(exception_type, exception_message)]
 
     # Extract additional information
     dag = context.get("dag")
@@ -148,6 +192,17 @@ def task_fail_slack_alert(context):
     if DEPLOYMENT_ENVIRONMENT != "production":
         env_indicator = f" *{DEPLOYMENT_ENVIRONMENT.capitalize()} Environment*"
 
+    # Format all exceptions for display
+    exceptions_text = ""
+    if len(all_exceptions) == 1:
+        # Single exception - use original format
+        exceptions_text = f"*Exception Type*: `{exception_type}`\n        *Exception Message*: `{exception_message}`"
+    else:
+        # Multiple exceptions - list them all
+        exceptions_text = f"*Exceptions Found ({len(all_exceptions)} total)*:"
+        for i, (exc_type, exc_msg) in enumerate(all_exceptions, 1):
+            exceptions_text += f"\n        {i}. *{exc_type}*: `{exc_msg}`"
+
     slack_msg = f"""
         {icon}{env_indicator} *Task failure* 
         {'\n\t\t' + byline if byline else ''}
@@ -156,8 +211,7 @@ def task_fail_slack_alert(context):
         *Execution Time*: `{exec_date}`
         *Schedule*: `{schedule_description}`
         *Duration*: `{duration} seconds`
-        *Exception Type*: `{exception_type}`
-        *Exception Message*: `{exception_message}`
+        {exceptions_text}
         <{log_url}|*View Task Log*>
     """
 
--- a/dags/utils/slack_operator.py
+++ b/dags/utils/slack_operator.py
@@ -112,36 +112,85 @@ def extract_exception_from_log(log_text):
 def extract_all_exceptions_from_log(log_text):
     import re
 
-    # Find all occurrences of 'Traceback (most recent call last):'
     exceptions = []
-    traceback_pattern = "Traceback (most recent call last):"
     
-    start_pos = 0
-    while True:
-        traceback_start = log_text.find(traceback_pattern, start_pos)
-        if traceback_start == -1:
-            break
+    # Split log into lines for easier processing
+    lines = log_text.split('\n')
+    
+    # Track whether we're in a traceback section
+    in_traceback = False
+    current_traceback_lines = []
+    
+    for line in lines:
+        line = line.strip()
         
-        # Find the end of this traceback by looking for the next traceback or end of text
-        next_traceback = log_text.find(traceback_pattern, traceback_start + len(traceback_pattern))
-        if next_traceback == -1:
-            traceback_text = log_text[traceback_start:]
-        else:
-            traceback_text = log_text[traceback_start:next_traceback]
-        
-        # Find the last line of this traceback (which contains the exception)
-        lines = traceback_text.strip().split("\n")
-        if len(lines) > 1:
-            last_line = lines[-1]
+        # Start of a new traceback
+        if "Traceback (most recent call last):" in line:
+            # Process previous traceback if we have one
+            if in_traceback and current_traceback_lines:
+                exception = _extract_exception_from_traceback_lines(current_traceback_lines)
+                if exception:
+                    exceptions.append(exception)
             
-            # Extract exception type and message
-            match = re.match(r"([\w.]+): (.*)", last_line)
-            if match:
-                exceptions.append((match.group(1), match.group(2)))
-        
-        start_pos = traceback_start + len(traceback_pattern)
+            # Start new traceback
+            in_traceback = True
+            current_traceback_lines = [line]
+            
+        elif in_traceback:
+            # Check if this line ends the current traceback
+            # Lines that typically end a traceback: empty lines, INFO logs, or new sections
+            if (line == "" or 
+                "INFO -" in line or 
+                "ERROR -" in line or 
+                "WARNING -" in line or
+                "The above exception was the direct cause of the following exception:" in line):
+                
+                # Process current traceback before ending
+                if current_traceback_lines:
+                    exception = _extract_exception_from_traceback_lines(current_traceback_lines)
+                    if exception:
+                        exceptions.append(exception)
+                
+                # Reset for potential next traceback
+                if "The above exception was the direct cause of the following exception:" in line:
+                    # This indicates chained exceptions, stay in traceback mode
+                    current_traceback_lines = []
+                else:
+                    # End of traceback section
+                    in_traceback = False
+                    current_traceback_lines = []
+            else:
+                # Add line to current traceback
+                current_traceback_lines.append(line)
     
+    # Process final traceback if we ended while in one
+    if in_traceback and current_traceback_lines:
+        exception = _extract_exception_from_traceback_lines(current_traceback_lines)
+        if exception:
+            exceptions.append(exception)
+    
+    print("Found exceptions: ", exceptions)
     return exceptions if exceptions else [(None, None)]
+
+
+def _extract_exception_from_traceback_lines(traceback_lines):
+    import re
+    
+    # Look for the exception line (usually the last non-empty line)
+    for line in reversed(traceback_lines):
+        line = line.strip()
+        if line and not line.startswith('File ') and not line.startswith('Traceback'):
+            # Try to match exception pattern: ExceptionType: message
+            match = re.match(r'^([\w.]+):\s*(.*)', line)
+            if match:
+                return (match.group(1), match.group(2))
+            
+            # Sometimes exceptions don't have messages, just the type
+            match = re.match(r'^([\w.]+)$', line)
+            if match:
+                return (match.group(1), "")
+    
+    return None
 
 
 def task_fail_slack_alert(context):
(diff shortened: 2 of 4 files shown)


dags/utils/slack_operator.py
```diff
@@ -87,28 +87,6 @@ def get_central_time_exec_data(context):
     return local_tz.convert(execution_date_timestamp).format("MM/DD/YYYY hh:mm:ss A")
 
 
-def task_fail_slack_alert_critical(context):
```
frankhereford, line 90: I'm tagging-along this change. This function was unused, and it certainly wasn't hurting anything, but I feel like our general policy is to remove stuff that isn't running. LMK if you want this to stay in, I'm totally cool with that too.

dags/utils/slack_operator.py
```diff
@@ -87,28 +87,6 @@ def get_central_time_exec_data(context):
...
-    )
-    failed_alert = SlackWebhookOperator(
-        task_id="slack_critical_failure",
-        slack_webhook_conn_id=SLACK_CONN_ID,
-        message=slack_msg,
-        username="airflow",
-    )
-    return failed_alert.execute(context=context)
-
-
```
frankhereford, line 111: I'm tagging-along this change. This function was unused, and it certainly wasn't hurting anything, but I feel like our general policy is to remove stuff that isn't running. LMK if you want this to stay in, I'm totally cool with that too.

dags/utils/slack_operator.py
```diff
@@ -131,6 +113,123 @@ def extract_exception_from_log(log_text):
...
+    "Airflow" for exceptions after that line.
+    """
+    
+    exceptions = []
+    lines = log_text.split('\n')
+    
+    # Find the "Task failed with exception" dividing line
+    task_failed_line_idx = None
+    for i, line in enumerate(lines):
+        if "Task failed with exception" in line:
```
frankhereford: This particular string is the boundary between:

* Our code: exceptions that are coming out of our containerized, inside-the-black-box python apps that do our ETL work. 
* Not our code: exceptions that are coming out of Airflow itself, such as its code that starts and manages docker containers.

dags/utils/slack_operator.py
```diff
@@ -131,6 +113,123 @@ def extract_exception_from_log(log_text):
...
+            task_failed_line_idx = i
+            break
+    
+    # Look for all traceback sections
+    traceback_starts = []
+    for i, line in enumerate(lines):
+        if "Traceback (most recent call last):" in line:
```
frankhereford: This line is the marker for the start of a traceback.

dags/utils/slack_operator.py
```diff
@@ -131,6 +113,123 @@ def extract_exception_from_log(log_text):
...
+        # Determine source based on position relative to "Task failed with exception"
+        if task_failed_line_idx is None:
+            source = "ETL"  # Default to ETL if no dividing line found
+        elif start_idx < task_failed_line_idx:
+            source = "ETL"
+        else:
+            source = "Airflow"
+        
+        # Find the exception line for this traceback
+        exception_line = _find_exception_line_in_traceback(lines, start_idx)
```
frankhereford: The "Exception Line" is the marker for the end of a traceback. This is the tough one. In general, they are of the form:

```
<Exception Type>: <Exception Message>
```

Not a lot to hang onto there in terms of writing heuristics, so we're pretty much looking for that colon as an indicator of the final printout of the exception type and message. 

One snag is that the Exception message is not required to be a single line (or even have a colon!), so in the handful of places where we are fancy and printing formatted exception messages in custom handlers, we will only get the first line of the exception in slack. 

I don't think it's worth it to try to refine the parser further to get optionally multiline exception messages. The whole thing is already a little black-magic'y as it's parsing things that were not meant to be parsed, so let's go KISS principals.

dags/utils/slack_operator.py
```diff
@@ -131,6 +113,123 @@ def extract_exception_from_log(log_text):
...
+    
+    print(f"Found exceptions: {exceptions}")
+    return exceptions if exceptions else [(None, None, "Unknown")]
+
+
+def _find_exception_line_in_traceback(lines, traceback_start_idx):
+    """
+    Find the actual exception line (final line) in a traceback section.
+    """
+    # Start from the traceback line and look forward
```
frankhereford: In the containing block that called this function, we were looking backwards-up-the-log, backwards-in-time looking for exception beginnings because we want them last-printed, first-processed, but once we're inside and looking to find the end of an exception, we have to turn around and start parsing forwards-down-the-log.

dags/utils/slack_operator.py
```diff
@@ -131,6 +113,123 @@ def extract_exception_from_log(log_text):
...
+    while i < len(lines):
+        line = lines[i].strip()
+        
+        # Stop if we hit another traceback
+        if "Traceback (most recent call last):" in line:
+            break
+            
+        # Stop if we hit certain log markers that indicate end of traceback
+        if (line.startswith('During handling of the above exception') or
+            line.startswith('The above exception was the direct cause')):
```
frankhereford: This is a list of the ways that python reports the relationship between the blocks of code involved in a trace back up the execution stack reported as part of an exception.

These are the only two individual examples of these relationship indicators that I'm aware of, but if you know more of them, please let me know. Thanks! 🐍

dags/utils/slack_operator.py
```diff
@@ -131,6 +113,123 @@ def extract_exception_from_log(log_text):
...
+        line.startswith('^') or
+        'Traceback' in line or
+        line.startswith('During handling') or
+        line.startswith('The above exception')):
+        return False
+    
+    # Look for Python exception patterns
+    # Must start with a capital letter followed by word characters, dots, underscores
+    # Common exception endings but not required
+    exception_pattern = r'^[A-Z][A-Za-z0-9_.]*(?:Error|Exception|Warning|Timeout)?(?::\s|$)'
```
frankhereford: In the comment above around line 151, I referred to the single colon that we are using to identify the end of a traceback and it's tucked away here in this regex. 

`(?::\s|$)` means "Please use this logical grouping for boolean operations and do not capture the value for use outside of the expression. Match a colon followed by a character of whitespace or match the end of a line."

dags/utils/slack_operator.py
```diff
@@ -131,6 +113,123 @@ def extract_exception_from_log(log_text):
...
+def _parse_exception_line(line):
+    """Parse an exception line into (type, message)"""
+    import re
+    
+    line = line.strip()
+    
+    # Pattern for ExceptionType: message
+    match = re.match(r'^([A-Z][A-Za-z0-9_.]*)\s*:\s*(.*)', line)
+    if match:
+        return match.group(1), match.group(2)
```
frankhereford: Given that we've made it this far, we know we're dealing with an exception type line, so let's look for the first colon and use that to delimit the exception type from the exception message. This makes further, subsequent colons that pop up in the exception message get handled correctly and not used as a delimiting marker.

dags/utils/slack_operator.py
```diff
@@ -141,14 +240,40 @@ def task_fail_slack_alert(context):
...
+    # First, get exceptions from Docker container logs if available
+    if DEBUG_LOG_TEXT is not None:
+        # DEBUG MODE: Use the debug log text instead of actual logs
+        print(f"DEBUG MODE: Using debug log text for parsing")
+        parsed_exceptions = extract_all_exceptions_from_log(DEBUG_LOG_TEXT)
+        
+        # Add parsed exceptions (filter out None entries)
+        for exc in parsed_exceptions:
+            if exc[0] is not None:
+                all_exceptions.append(exc)
```
frankhereford: So, testing this is a real mind-bender. I assert that it's not possible to plan errors and therefore, I am unable to give you testing instructions that actually simulate the execution stack or contextual environment this code needs to work in. 

So, given this, I'm calling this part of the system the best place to inject a crafted log and have it handled by the parsing system and used in a slack warning message. Should I leave it in after we test this PR? I'm thinking yes, due to the very unusual considerations around testing this code.

dags/utils/slack_operator.py
```diff
@@ -170,6 +295,21 @@ def task_fail_slack_alert(context):
...
+        source = all_exceptions[0][2] if len(all_exceptions[0]) > 2 else "Unknown"
+        exceptions_text = f"*Exception Type*: `{exception_type}` _(from {source})_\n        *Exception Message*: `{exception_message}`"
+    else:
+        # Multiple exceptions - list them all with sources
+        exceptions_text = f"*Exceptions Found ({len(all_exceptions)} total)*:"
+        for i, exc_tuple in enumerate(all_exceptions, 1):
+            exc_type = exc_tuple[0]
+            exc_msg = exc_tuple[1] 
+            source = exc_tuple[2] if len(exc_tuple) > 2 else "Unknown"
+            exceptions_text += f"\n        {i}. *{exc_type}* _(from {source})_: `{exc_msg}`"
```
frankhereford: And finally, we add support for printing out human readable messages from the list of tuples each representing a parsed exception.

I think the `from <source>` is a valuable addition because it makes it clear what errors came out of our code and what came out of airflow's code. The airflow's code exception is still very important to report, as it may include information about how the stack failed, such as external return codes or what-have-you.

dags/utils/slack_operator.py
```diff
@@ -11,6 +11,10 @@
 
 DEPLOYMENT_ENVIRONMENT = getenv("ENVIRONMENT", "development")
 
+# DEBUG: Set this to a string containing log text to test exception parsing
+# When None, normal operation resumes
+DEBUG_LOG_TEXT = None
```
frankhereford: I comment on the inclusion of the DEBUG mechanism later in this file around line 247.

dags/test_docker_failure.py
```diff
@@ -21,19 +21,52 @@
...
+    print("About to raise ConnectionError")
+    raise ConnectionError("Connection failed")
+
+if __name__ == "__main__":
+    try:
+        outer_function()
+    except Exception as e:
+        print("Final exception caught at top level:")
+        traceback.print_exc()
+        sys.exit(1)
```
frankhereford, line 67: Here's an inception moment with a python program wrapped in a bash program stored inside another python program with a docker layer in the middle somewhere. 😵‍💫 I think it's still plenty readable and understandable though.

This is an extension of the simulation of a python stack failing while running inside a docker container running inside airflow's python execution stack -- and it's extended in such a way that it fails with chained exceptions.

This is about as close to simulating the way most of our ETLs operate and fail as I think we can reasonably do without building some external infrastructure to have our test ETL interact with.

dags/utils/slack_operator.py
```diff
@@ -131,6 +113,123 @@ def extract_exception_from_log(log_text):
...
+        
+        # Find the exception line for this traceback
+        exception_line = _find_exception_line_in_traceback(lines, start_idx)
+        
+        if exception_line:
+            exc_type, exc_msg = _parse_exception_line(exception_line)
+            if exc_type:
+                exceptions.append((exc_type, exc_msg, source))
+    
+    print(f"Found exceptions: {exceptions}")
```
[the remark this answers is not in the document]
frankhereford: 🙈 Thank you copilot .. [34fbfc6](https://github.com/cityofaustin/atd-airflow/pull/300/commits/34fbfc659e9dfeea3a5223d5cbf15ce5255b4695)

dags/utils/slack_operator.py
```diff
@@ -141,14 +240,40 @@ def task_fail_slack_alert(context):
...
     )
 
-    if exception and hasattr(exception, "logs") and exception.logs:
+    # Extract all exceptions from logs if available
+    all_exceptions = []
+    
+    # First, get exceptions from Docker container logs if available
+    if DEBUG_LOG_TEXT is not None:
+        # DEBUG MODE: Use the debug log text instead of actual logs
+        print(f"DEBUG MODE: Using debug log text for parsing")
```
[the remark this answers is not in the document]
frankhereford: This actually is a helpful message to have show up in airflow's logs when you have overloaded log values and this printing method here accomplishes that. I don't want to introduce logging or a logging system that might conflict or be harder to parse into these logs.

dags/utils/slack_operator.py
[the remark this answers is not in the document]
frankhereford: I feel like the pattern is documented. If someone really needs a blow-by-blow of how the regex works, they should drop it into a LLM at that time, not have it in this code.

dags/utils/slack_operator.py
```diff
@@ -141,14 +240,40 @@ def task_fail_slack_alert(context):
...
+        # Add parsed exceptions (filter out None entries)
+        for exc in parsed_exceptions:
+            if exc[0] is not None:
+                all_exceptions.append(exc)
+    
+    # Always add the Airflow-level exception as well if not already found
+    airflow_exception = (exception_type, exception_message, "Airflow")
+    # Check if we already have this exception from parsing
+    airflow_already_found = any(exc[0] == exception_type and exc[1] == exception_message 
+                                for exc in all_exceptions)
```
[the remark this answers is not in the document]
frankhereford: This is fine. If we have a list of exceptions long enough for this to matter, we have much larger problems.

dags/utils/slack_operator.py
```diff
@@ -170,6 +295,21 @@ def task_fail_slack_alert(context):
...
+        # Single exception - use original format with source
+        source = all_exceptions[0][2] if len(all_exceptions[0]) > 2 else "Unknown"
+        exceptions_text = f"*Exception Type*: `{exception_type}` _(from {source})_\n        *Exception Message*: `{exception_message}`"
+    else:
+        # Multiple exceptions - list them all with sources
+        exceptions_text = f"*Exceptions Found ({len(all_exceptions)} total)*:"
+        for i, exc_tuple in enumerate(all_exceptions, 1):
+            exc_type = exc_tuple[0]
+            exc_msg = exc_tuple[1] 
+            source = exc_tuple[2] if len(exc_tuple) > 2 else "Unknown"
```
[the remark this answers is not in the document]
frankhereford: KISS. This is a 3 element tuple, this is OK IMHO.

dags/utils/slack_operator.py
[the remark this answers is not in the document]
chiaberry: I was going to ask more about reasons to not introduce a logging system, but this is the only time anything is printed or logged in this file. And based on my understanding (and your comment below) this is only used for testing. I am conflicted about keeping the DEBUG_LOG_TEXT mode in the dag after the PR is approved because I think we will get better test failures with our real dags, so I have my doubts how often we will need this. But it is useful to keep in mind, so maybe a compromise would be a comment with a link back to this PR?

dags/utils/slack_operator.py
```diff
@@ -141,14 +239,40 @@ def task_fail_slack_alert(context):
...
+    # Check if we already have this exception from parsing
+    airflow_already_found = any(exc[0] == exception_type and exc[1] == exception_message 
+                                for exc in all_exceptions)
+    if not airflow_already_found:
+        all_exceptions.append(airflow_exception)
+    
+    # Use the last exception as primary for backward compatibility
+    if all_exceptions:
+        exception_type = all_exceptions[-1][0]
+        exception_message = all_exceptions[-1][1]
```
chiaberry: can you elaborate more what "primary for backward compatibility" means? 

This is replacing the default `type(exception).__name__ if exception else "Unknown"` with the last exception from the group, is it so that if there is only one log in the group, we dont show the original default, yeah?

chiaberry (approved):
I reviewed the code and tested the sample test_docker_failure, and saw the ✨ stacked exceptions, really nice! 

I am trusting your testing during development with experimenting with different logs and I completely agree that it is always going to be possible to come up with logs that would not fit perfectly in this parser. 

You mentioned deploying this feature branch as production before merging it to our production branch in order to further refine it, what do you think about just merging it to production (once you get more approvals of course) and then seeing what refinement it could use after being out in the wild for a few weeks? 

we luckily have only been getting a handful of failures in the airflow channel, so i dont know how long it would take until we get new and novel failure logs. And if we are running off a feature branch it could potentially block deploying dags that are merged to production. 

Its going to be great to get more than `Docker container failed: {'StatusCode': 1}`! 🎊

dags/utils/slack_operator.py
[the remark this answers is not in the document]
frankhereford: Thanks for this thoughtful question Chia. 

I think, for one, I could have done a better job naming this variable, because it does make it sound like it's concerning logging output, but really, it's concerning getting an existing log /into/ the system to be processed, as part of debugging. 

I'm becoming less attached to having it in here:
  * One would have to figure out /why/ it's in there, and that's not terribly obvious
  * There's no reason why I wouldn't expect the next dev to have any issue figuring out how to inject a log without this breadcrumb. 

~I'm going to take that out, and I'll edit this comment with a SHA when I'm done.~

When testing is done, as it's a part of the testing process, I'll take it out and comment here with a SHA to indicate that. 

Thank you!

dags/utils/slack_operator.py
frankhereford: ### Changes made

Thanks for calling out this confusing placement for these assignments. I've pulled them down into the block that uses them. It handles the case when there is only one exception discovered in the airflow log. The variables are only used once, so it's debatable if they should even be there, but I like them because `all_exceptions[-1][1]` doesn't offer much explanation about what data one could expect to find in that space. 


```
    if len(all_exceptions) == 1:
        # Single exception - use original format with source
        exception_type = all_exceptions[-1][0]
        exception_message = all_exceptions[-1][1]
        source = all_exceptions[0][2] if len(all_exceptions[0]) > 2 else "Unknown"
        exceptions_text = f"*Exception Type*: `{exception_type}` _(from {source})_\n        *Exception Message*: `{exception_message}`"
```

https://github.com/cityofaustin/atd-airflow/pull/300/commits/bb19f536a05037083ce03da33a508dd56e6f205d

### Answering your questions

To your question about the comment -- The behavior that is being maintained for backwards compatibility is the system's printing of a single exception correctly. It doesn't use a list format, doesn't include an exception number and correctly refers to the exception in the singular.

I think your explanation about how those assignments are overwriting the default exception message and type from when they were set at the top of the function is spot on, 💯.

frankhereford:
@chiaberry wrote in a review comment:

> You mentioned deploying this feature branch as production before merging it to our production branch in order to further refine it, what do you think about just merging it to production (once you get more approvals of course) and then seeing what refinement it could use after being out in the wild for a few weeks?
> 
> we luckily have only been getting a handful of failures in the airflow channel, so i dont know how long it would take until we get new and novel failure logs. And if we are running off a feature branch it could potentially block deploying dags that are merged to production.

I'm totally 👍 on skipping the last pass at refinement of the heuristics in production before merger. I was hedging against the situation where a parser that I thought would be good in the general case ended up being only good in the cases that I envisioned. I'll mark that out of the main comment in this PR. 🙏

dags/utils/slack_operator.py
mddilley: Sounds good to me. I think the ability to Slack mention devs is a better alternative to this critical alert. Thanks for the cleanup! 🧹

dags/utils/slack_operator.py
mddilley: Sounds good to me too! You and Chia mentioned it elsewhere in this PR, and I agree that we can put more work into this in the future if more types of common errors emerge that aren't bubbling up to the Slack messages.

dags/utils/slack_operator.py
```diff
@@ -141,14 +244,37 @@ def task_fail_slack_alert(context):
         str(exception) if exception else "No exception message available"
     )
 
-    if exception and hasattr(exception, "logs") and exception.logs:
+    # Extract all exceptions from logs if available
```
mddilley: Since you have such nice helpers elsewhere in this code, what do you think about creating a couple more for handling the extraction and formatting of the exceptions so not so much of this code in this function is about munging the output of the log-handling? Maybe something like:

```python
all_exceptions = get_all_exceptions(...)
exceptions_text = format_exceptions(all_exceptions)
```

Also, a module for the log parsing functions could be a nice touch.

dags/test_docker_failure.py
mddilley, line 1: Nice! Might be good to mentio this new DAG in the readme part about the Slack notifier since we currently mention the `test_slack_notifier` DAG in there.

Also, I noticed the **Slack operator utility** section mentions the critical alert that is removed in this PR. Could you remove that? 🙏

mddilley (approved):
Looks and works great! I added a few comments and some suggested changes to the readme. I agree with Chia that we should get it straight into prod after merge and then tweak if needed. Very fun one to think through!! 🚢

dags/utils/slack_operator.py
frankhereford: Yea - I think that's a great suggestion. https://github.com/cityofaustin/atd-airflow/pull/300/commits/781e13adb3b43f99392d6541ce950b8f1db17ce0 & https://github.com/cityofaustin/atd-airflow/pull/300/commits/95108548479b357c848aaa0a8ecf584151cf8554

dags/test_docker_failure.py
frankhereford: Thanks for the great suggestion Mike -- https://github.com/cityofaustin/atd-airflow/pull/300/commits/2e3d0a8dcd4c2d597baa8c49a81212df27aaab82.

PR #305, opened 2025-08-25
Ignore date filter arg SMO signals

Charlie-Henry:
Discovered that new/updated records can sometimes slip through the cracks and not get updated in Postgrest, which can [cause errors](https://austininnovation.slack.com/archives/C013G4RNJ01/p1756099913053879). Removing the date filter arg from the records_to_postgrest task here fixed it for me, and only took a few seconds longer to run.

## Associated repo
https://github.com/cityofaustin/atd-knack-services/blob/production/services/records_to_knack.py

## Testing

**Steps to test:**

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/atd_knack_artbox_signals.py
+++ b/dags/atd_knack_artbox_signals.py
@@ -97,7 +97,7 @@ with DAG(
         image=docker_image,
         docker_conn_id="docker_default",
         auto_remove="force",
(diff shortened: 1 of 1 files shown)


chiaberry (approved):
Thanks for tracking down the issue!

PR #307, opened 2025-10-01
[Collector] Upgrade Airflow platform to major version 3

frankhereford:
# Airflow major version upgrade to `v3.1.6` up from the 2.x family. 

## Issue
[#25197](https://github.com/cityofaustin/atd-data-tech/issues/25197): Deploy Airflow 3 to the on prem server

## Branch `frank/airflow-three` vs. `frank/dag-refactor` 

This PR references the `airflow-three` branch, which contains the work that upgrades the Airflow ETL platform itself. In a separate branch / PR, you'll find `frank/dag-refactor`, [here](https://github.com/cityofaustin/atd-airflow/pull/308), which contains the work to make our DTS in-house DAGs run on Airflow 3. 

DAGs are moved, temporarily, out of `./dags/` and into `./not-yet-upgraded-dags/` in this branch, as they prevent airflow from operating in their normal location. The `dag-refactor` brings them back into the normal location and includes the updates that enable them to work in the new environment. 

## Notable changes

* Airflow 2x → 3.1.6
* Postgres 13 → 18 (+5 versions!)
* ~Redis 7.2 → 8.4~ No longer needed, used only by Celery. 👋 
* haproxy 2.6 → LTS (currently: 3.2.10)

## Upgrade opinion

This PR has an update to the underlying postgres server, and it's 5 major versions, so to transforming that dataset along through 5 version jumps to get here is going to be a task. I think we should should start with a fresh database and not worry about bringing our old one along with us. We purge our records frequently and automatically, and it's very easy to configure the instance once it's running:

* Add a connection for slack named `slack` via the web UI
* Add individual user accounts

This proposal only works with the caveat that we are able to back-specify the `last run` date for a DAG.

## Changes to how our DAGs check the last successful run end date

The way we accessed this previously does not work under airflow 3. There is a 1:1 replacement though, and you can look at [this new test DAG](https://github.com/cityofaustin/atd-airflow/blob/frank/airflow-three/dags/log_last_run_time.py) to see how it works and interact with it. You can see it start to pick up the last run date on the second run after a fresh database reset. 

This "last run end date" can be set manually, and being able to do so is important when firing up a new database underneath airflow such as this upgrade.

```bash
# with the stack running, this will get you a `psql` shell
docker compose exec --user airflow airflow-dag-processor airflow db shell;
```
```sql
INSERT INTO dag_run (
  dag_id,
  run_id,
  logical_date,
  data_interval_start,
  data_interval_end,
  run_after,
  queued_at,
  start_date,
  end_date,
  state,
  run_type,
  triggered_by,
  triggering_user_name,
  conf,
  context_carrier,
  span_status
)
VALUES (
  '<DAG ID>',
  (now() AT TIME ZONE 'UTC') - interval '1 hour',
  (now() AT TIME ZONE 'UTC') - interval '1 hour',
  (now() AT TIME ZONE 'UTC') - interval '1 hour',
  (now() AT TIME ZONE 'UTC') - interval '1 hour',
  (now() AT TIME ZONE 'UTC'),
  (now() AT TIME ZONE 'UTC'),
  (now() AT TIME ZONE 'UTC') + interval '1 second',
  'success',
  'manual',
  'UI',
  'admin',
  '{}'::jsonb,
  'ended'
);
```

## Testing

* Spin it up locally and use it however you want
  * To enable your local stack to send slack messages, you can add a "Connection" namd `slack` using the info in 1PW `Airflow - Slack Bot` and that should get that going.
  * Maybe kick off some of the handful of test DAGs
  * Maybe find a DAG you know and port it over to this platform if you want hard-mode


config/airflow.cfg
```diff
@@ -0,0 +1,2314 @@
...
+#
+default_timezone = utc
+
+# The executor class that airflow should use. Choices include
+# ``LocalExecutor``, ``CeleryExecutor``,
+# ``KubernetesExecutor`` or the full import path to the class when using a custom executor.
+#
+# Variable: AIRFLOW__CORE__EXECUTOR
+#
+executor = LocalExecutor
```
frankhereford, line 51: 👋 🥬 

This removes our use of the celery executor and brings the local executor on board in its place.

config/airflow.cfg
```diff
@@ -0,0 +1,2314 @@
...
+# used in your environment, this list controls who can access the environment.
+# 
+# List of user-role delimited with a comma. Each user-role is a colon delimited couple of username and
+# role. Roles are predefined in simple auth managers: viewer, user, op, admin.
+#
+# Example: simple_auth_manager_users = bob:admin,peter:viewer
+#
+# Variable: AIRFLOW__CORE__SIMPLE_AUTH_MANAGER_USERS
+#
+simple_auth_manager_users = admin:admin
```
frankhereford, line 69: In local dev mode, this is the account you'll use to access the airflow web UI as any personalized account would get wiped away anytime the local dev database state is reset. In production, this account exists only momentarily and has the password changed at the same time that dev's accounts are created.

config/airflow.cfg
```diff
@@ -0,0 +1,2314 @@
...
+#
+# Variable: AIRFLOW__CORE__MAX_ACTIVE_TASKS_PER_DAG
+#
+max_active_tasks_per_dag = 2 
+
+# Are DAGs paused by default at creation
+#
+# Variable: AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION
+#
+dags_are_paused_at_creation = True
```
frankhereford, line 111: This is what we want right? We could have them start their schedule immediately as soon as we `git pull` them onto the machine if we wanted.

dags/utils/slack_operator.py
```diff
@@ -119,8 +120,9 @@ def task_fail_slack_alert(context):
     log_url = task_instance.log_url
     duration = getattr(task_instance, "duration", "Not available")
 
-    schedule_interval = dag.schedule_interval if dag else None
-    schedule_description = format_schedule(schedule_interval)
+    schedule_description = (
+        dag.timetable.summary if dag and hasattr(dag, "timetable") else "Not available"
```
frankhereford: I don't think it's going to matter to us, but here is info on the [Timetable](https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/timetable.html) representation of a schedule. Our cron expressions are 👍.

Here we're using a new method from a `timetable` to get at original expression before the conversion into a `timetable`. Maybe a little brittle but i think it will work as long as we keep using cron expressions.

dags/utils/slack_operator.py
```diff
@@ -110,17 +110,37 @@ def build_exception_text(all_exceptions):
...
+    duration = getattr(task_instance, "duration", None)
+    if duration is not None:
+        return duration
+
+    start_date = getattr(task_instance, "start_date", None)
+    end_date = getattr(task_instance, "end_date", None)
+    if start_date and end_date:
+        return (end_date - start_date).total_seconds()
+
+    return None
```
frankhereford, line 123: I don't see the `duration` value being populated into the `duration`, so there's some additional logic here to back that up.

dags/utils/slack_operator.py
```diff
@@ -145,32 +165,11 @@ def task_fail_slack_alert(context):
...
-    failed_alert = SlackWebhookOperator(
-        task_id="slack_failure",
-        slack_webhook_conn_id=SLACK_CONN_ID,
-        message=slack_msg,
-        username="Airflow Alert",
-    )
-    return failed_alert.execute(context=context)
-
-
-def task_success_slack_alert(context):
```
frankhereford, line 157: I removed this function. we didn't use it anywhere I could see and it's gotten pretty diverged from our other alert function.

dags/test_docker_failure.py
frankhereford, line 1: I (🤖) did bring refactor our two slack testing DAGs to use the decorator syntax of the TaskFlow system. I only point out the 🤖 use because it was particularly effective at the task.

haproxy/haproxy-production.cfg
```diff
@@ -33,41 +33,21 @@ defaults
...
-
   # default to airflow web ui
   default_backend airflow
 
 # airflow webservice as provided by a modified copy of airflow
 backend airflow
   option httpchk GET /
   default-server inter 5s maxconn 50
-  server airflow airflow-webserver:8080 check
+  server airflow airflow-apiserver:8080 check
```
frankhereford, line 52: A few new names show up for the major components, and this is the biggest. `webserver` → `apiserver`. It still serves up the web app though, despite the new name.

config/airflow.cfg
frankhereford, line 1: I moved this file to match an [upstream configuration in the example airflow docker-compose.yaml file](https://github.com/apache/airflow/blob/main/airflow-core/docs/howto/docker-compose/docker-compose.yaml#L77). It makes it harder to review which is unfortunant. I took a vanilla one out of the upstream airflow repo and started from there, including bringing in changes from John's PR [here](https://github.com/cityofaustin/atd-airflow/pull/111/changes#diff-56ac061ee2d6ec25719f9ea51ca2bd4ac195239e01c7f6ad8a7f7393c3118691) where he refines and tunes airflow 2 in its early days.

haproxy/haproxy.cfg
frankhereford, line 1: The `haproxy` configurations where particularly simplified by removing celery.

frankhereford (commented):
Going over the PR here before releasing it and here are some notes. Thanks!

docker-compose.yaml
```diff
@@ -1,48 +1,46 @@
...
-    FLOWER_AUTH_USERNAME: ${_AIRFLOW_WWW_USER_USERNAME}
-    FLOWER_AUTH_PASSWORD: ${_AIRFLOW_WWW_USER_PASSWORD}
-    AIRFLOW_CHECKOUT_PATH: ${AIRFLOW_PROJ_DIR}
-    AIRFLOW_CONN_DOCKER_DEFAULT: "docker://${DOCKER_HUB_USERNAME}:${DOCKER_HUB_TOKEN}@https%3A%2F%2Findex.docker.io%2Fv1%2F:80"
-    # Quiet SQLAlchemy warnings
-    SQLALCHEMY_SILENCE_UBER_WARNING: 1
 
-    AIRFLOW__CORE__EXECUTOR: CeleryExecutor
+    # Airflow stock suggested environment variables
+    #AIRFLOW__CORE__EXECUTOR: CeleryExecutor
```
mddilley: could you remove this commented out line in favor of the real one below?

config/airflow.cfg
```diff
@@ -0,0 +1,2314 @@
...
+#
+max_callbacks_per_loop = 20
+
+# Number of seconds after which a DAG file is parsed. The DAG file is parsed every
+# ``[dag_processor] min_file_process_interval`` number of seconds. Updates to DAGs are reflected after
+# this interval. Keeping this number low will increase CPU usage.
+#
+# Variable: AIRFLOW__DAG_PROCESSOR__MIN_FILE_PROCESS_INTERVAL
+#
+min_file_process_interval = 60
```
mddilley: I wonder if we should this [like we did for Airflow 2](https://github.com/cityofaustin/atd-airflow/pull/111/changes#r1232261900)? The end of that comment's thread still feel relevant along with the comments for this parameter that start on line 2183.

dags/utils/slack_operator.py
mddilley: thanks for cleaning this up! feels safe to me since we haven't used it in 5+ year   🧹😅

config/airflow.cfg
mddilley: Nice! It was also nice to see that we have less config w/o CeleryExecutor 😎

config/airflow.cfg
mddilley: This feels right to me! I feel caught off guard when a DAG immediately starts itself. I'm also thinking about future parameterized DAGs where we might want to do a full replace as the first run or something like that.

dags/test_docker_failure.py
mddilley: 🤖 ❤️

dags/test_docker_failure.py
```diff
@@ -67,10 +66,18 @@ def inner_function():
         sys.exit(1)
 """,
         ],
-        docker_conn_id="docker_default",
+        # docker_conn_id="docker_default",
```
mddilley: Do we need to update the default Docker connection like in https://github.com/cityofaustin/atd-airflow/pull/165? Airflow 2 is really feeling like it happened ages ago now!!

toolbox/README.md
mddilley, line 1: 🙌 🙏 Thank you for adding this doc!

docker-compose.yaml
```diff
@@ -1,48 +1,46 @@
...
-  image: atddocker/atd-airflow:production
+  # In order to add custom dependencies or upgrade provider distributions you can use your extended image.
+  # Comment the image line, place your Dockerfile in the directory where you placed the docker-compose.yaml
+  # and uncomment the "build" line below, Then run `docker-compose build` to build the images.
 
-  environment: &airflow-common-env
-    AIRFLOW__CORE__FERNET_KEY: ${_AIRFLOW__CORE__FERNET_KEY}
-    AIRFLOW__WEBSERVER__BASE_URL: ${_AIRFLOW__WEBSERVER__BASE_URL:-http://localhost:8080}
-    # You need to explicitly allow .env vars into the airflow containers's environments here
+  #image: ${AIRFLOW_IMAGE_NAME:-apache/airflow:3.1.0-slim-python3.13}
```
mddilley: This will eventually be `atddocker/atd-airflow:production`, right? We just can't set it yet since this image is still v2. Thanks for sorting out all these small details!

docker-compose.yaml
```diff
@@ -53,30 +51,14 @@ x-airflow-common: &airflow-common
...
-      - 8081:81
-    depends_on:
-      - airflow-webserver
-    volumes:
-      - ${AIRFLOW_PROJ_DIR}/haproxy/ssl:/usr/local/etc/haproxy/ssl
-      - ${AIRFLOW_PROJ_DIR}/haproxy/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg
-
   postgres:
-    image: postgres:13
+    image: postgres:18
```
mddilley, line 54: 🏎️ 😎

mddilley (commented):
I was able to start up this stack locally and test the Slack notifier DAG. I'll go for hard mode later, but I wanted to release some of these comments. Soooo cool to see the new UI and really nice to see the amount of config reduced by implementing the LocalExecutor. 🙌

docker-compose.yaml
frankhereford: You're right, 100%. Here's an update to help clarify -- thank you for pointing that out. https://github.com/cityofaustin/atd-airflow/pull/307/commits/5fd24d3d33dec8535b11fb42cf7c696ecfa50d29

dags/test_docker_failure.py
frankhereford: Yes! Thank you, and thank you to John for installing this in the first place. I had not picked up on the very first thing that John has in the PR you linked. I didn't realize that you could have an invisible connection if you set it via the environment, so I had not realized to bring it over. I only inspected the UI, and when this line had to come out, I thought that was just going to be a airflow 3 thing that either changed and we'd figure it out. I was 😵‍💫 and this comment cleared that up. 🙏 

https://github.com/cityofaustin/atd-airflow/pull/307/changes/5357adb763ac9de91c27f03f959c7e9e303de511 and I put the environment variable in `.env` on `02`.

config/airflow.cfg
frankhereford: Big yes to this! https://github.com/cityofaustin/atd-airflow/pull/307/commits/38df4d1fc4a66468bf05beb6f7ae11b9d982a656 @johnclary's  assessment in his comment [here](https://github.com/cityofaustin/atd-airflow/pull/307/commits/38df4d1fc4a66468bf05beb6f7ae11b9d982a656) is absolutely the way to go. TIL that the re-parser will check more frequently when it's recently detected a change. Subtle!

Also - for future us - `min_file_process_interval` is removed in 3.0 and replaced with `[dag_processor] refresh_interval`.

docker-compose.yaml
frankhereford: You bet. https://github.com/cityofaustin/atd-airflow/pull/307/commits/44fb046f05bc2d56f7a9ffccc325ffd877a37175

frankhereford:
> I was able to start up this stack locally and test the Slack notifier DAG. I'll go for hard mode later, but I wanted to release some of these comments. Soooo cool to see the new UI and really nice to see the amount of config reduced by implementing the LocalExecutor. 🙌

Mike - thank you for your advanced review on this. You solved some mysteries there; everything was super helpful! LMK anything else you see, and I'm looking forward to the next steps. 🙏

dags/utils/slack_operator.py
mddilley: I'm re-reading this, and this is strange because it seems like it was a bug in 3.0.6 ([link](https://www.mail-archive.com/commits@airflow.apache.org/msg434676.html)) where both `duration` and `end_date` were not available from failed tasks - and it was patched up! This works so probably not worth going down the rabbit hole. thanks for the workaround!

config/airflow.cfg
chiaberry: So this means when we git pull updates, we need to turn the dag on? I have the vague memory that it used to work like that, and I am glad to go back that way and feel like i have more control

frankhereford: That's what it means to me, but to be sure, I will check what happens when I move the next DAG back into `./dags/` in the refactoring PR. Thanks for asking. 🙏

chiaberry (commented):
I tried running the test slack notifier and I didnt get a slack notification: 

```
[2026-01-29 15:40:13] INFO - DAG bundles loaded: dags-folder
[2026-01-29 15:40:13] INFO - Filling up the DagBag from /opt/airflow/dags/test_slack_notifier.py
[2026-01-29 15:40:13] ERROR - Task failed with exception
Exception: Task failure test successfully triggered
File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/execution_time/task_runner.py", line 1004 in run

File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/execution_time/task_runner.py", line 1400 in _execute_task

File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/bases/operator.py", line 417 in wrapper

File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/bases/decorator.py", line 252 in execute

File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/bases/operator.py", line 417 in wrapper

File "/home/airflow/.local/lib/python3.12/site-packages/airflow/providers/standard/operators/python.py", line 214 in execute

File "/home/airflow/.local/lib/python3.12/site-packages/airflow/providers/standard/operators/python.py", line 237 in execute_callable

File "/home/airflow/.local/lib/python3.12/site-packages/airflow/sdk/execution_time/callback_runner.py", line 82 in run

File "/opt/airflow/dags/test_slack_notifier.py", line 39 in task_fail

[2026-01-29 15:40:13] WARNING - Slack failure alert failed: The conn_id `slack` isn't defined
```

chiaberry:
But i see that Mike was able to kick off the slack notifier, so I will see what I have different on my setup. This is all looking so great and I am going to try to port over a dag to try hard mode!

frankhereford:
> I tried running the test slack notifier and I didnt get a slack notification:

Thank you for mentioning that @chiaberry -- I forgot to put it in the instructions. If you add a "Connection" named `slack` with this info from 1PW `Airflow - Slack Bot`, that should enable your local stack to post to slack's endpoint. I'll add it to the instructions - thank you!

dags/test_docker_failure.py
```diff
@@ -1,36 +1,35 @@
+from __future__ import annotations
```
mddilley, line 1: i'm not in tune with the latest python stuff but it sounds like this would be useful if we start using type annotations

dags/test_docker_failure.py
```diff
@@ -1,36 +1,35 @@
...
-    dag.byline = f"Test stacked exceptions in a docker container"
-    dag.icon = ":test_tube:"
+    tags=["repo:atd-airflow", "slack"],
+    default_args={
+        "owner": "airflow",
+        "retries": 0,
+        "execution_timeout": pendulum.duration(minutes=5),
+        "on_failure_callback": task_fail_slack_alert,
+    },
+    doc_md="Throws stacked python exceptions from within a docker container",
```
mddilley, line 27: TIL! According to the [docs](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#dag-task-documentation), this is available in the web UI but I can't find it but I'm probably just missing it. 🙃 Were you able to see this anywhere? Maybe we should have the `description` that shows in the Details tab too?

docker-compose.yaml
```diff
@@ -1,48 +1,47 @@
...
-    # Quiet SQLAlchemy warnings
-    SQLALCHEMY_SILENCE_UBER_WARNING: 1
 
-    AIRFLOW__CORE__EXECUTOR: CeleryExecutor
+    # Airflow stock suggested environment variables
+    AIRFLOW__CORE__EXECUTOR: LocalExecutor
+    AIRFLOW__CORE__AUTH_MANAGER: airflow.providers.fab.auth_manager.fab_auth_manager.FabAuthManager
     AIRFLOW__DATABASE__SQL_ALCHEMY_CONN: postgresql+psycopg2://airflow:airflow@postgres/airflow
     AIRFLOW__CELERY__RESULT_BACKEND: db+postgresql://airflow:airflow@postgres/airflow
     AIRFLOW__CELERY__BROKER_URL: redis://:@redis:6379/0
```
mddilley: these two celery lines caught my eye. Can we remove these?

mddilley (approved):
I tested this stack a bunch with the Moped AGOL DAG (dry run) and was able to interact with our 1PW server, and I looked over the config and local/production files. All of the test DAGs worked as expected too. Thanks for making this easy to spin up and for the commands to poke around the DB. 🙏 🚀

dags/log_last_run_time.py
mddilley, line 1: I'm just now realizing that I was mixing this up `prev_end_date_success` with `prev_start_date_success`. So this DAG is showing that we can't get the end date, but we can get the start date which is what we use in DAGs like the [Moped AGOL incremental mode](https://github.com/cityofaustin/atd-airflow/pull/308/changes#diff-cc616d18c8a4ac2adf9dffc798d12ff4b6ce75971271ba676a7ab589ab0150b2). 

I'm just checking because I convinced myself that we needed a different mechanism to get `prev_start_date_success` for a second.

config/airflow.cfg
```diff
@@ -0,0 +1,2314 @@
...
+#
+dags_are_paused_at_creation = True
+
+# The maximum number of active DAG runs per DAG. The scheduler will not create more DAG runs
+# if it reaches the limit. This is configurable at the DAG level with ``max_active_runs``,
+# which is defaulted as ``[core] max_active_runs_per_dag``.
+#
+# Variable: AIRFLOW__CORE__MAX_ACTIVE_RUNS_PER_DAG
+#
+max_active_runs_per_dag = 1
```
mddilley, line 119: Reviewing https://github.com/cityofaustin/atd-airflow/pull/308/ reminded me that I saw this and forgot to comment. Did we use to have this defined for all DAGs in our v2 setup? I can't remember if there is a situation where we would want more one DAG run to be running at the same time. Probably not?

I'm wondering because I had to [explicitly set this](https://github.com/cityofaustin/atd-airflow/blob/6934af9cd75660215d70cd3afe7f7bdc6c6b1bf0/dags/atd_moped_components_to_agol.py#L103) in the DAG definition for the Moped AGOL DAG so it wouldn't run full replace and incremental simultaneously.

chiaberry (approved):
I have tested and moved knack-banner as a test https://github.com/cityofaustin/atd-airflow/pull/321

I am approving, but its my understanding this is collector branch and will be merged to production as the last step of deployment

docker-compose.yaml
frankhereford: You bet, https://github.com/cityofaustin/atd-airflow/pull/307/commits/ecdcbd8ddd45163cca8be7bcedcf545f693dd2b1, they are outta there.

dags/log_last_run_time.py
frankhereford: Thanks again Mike for asking this question. It surfaces not only that I had pulled the end-time of the last success in lieu of the start-time, but also, that I had pulled that data and verified it all within a single task. 

In practice, we call a task (`get_previous_run_date`) out of the `time.py` library and pass it on to the other tasks that need it. In Airflow 3, it looks like the use of `xcom` to pass data needs to be explicit instead of the implicit usage we've used in the past that simulated returning a value to be stored in a variable and then passed as an argument. 

I'm working out the details of that, fixing up the `time.py`, and adding a demonstration of its use in `log_last_run_time.py`.

dags/utils/time.py
frankhereford, line 1: I reworked this file from the ground up to get some more descriptive function names and to add support for both last successful end time and last successful begin time. 

I also made a comment and `import .. as ..` to handle the new function name in the single DAG that was still using it. I didn't realize how much we have almost stopped using this time utility before I put in this work, unfortunately.

dags/log_last_run_time.py
frankhereford: I gave the `utils.time.py` utility and this DAG a total refactor this morning. I've got a very clean demonstration of importing tasks and moving data out of them and into another task using xcom, and having it done such that it looks like the normal `return` value being passed into a function's argument.

I'm curious what you think @mddilley. By my read, there's actually only one file that still uses this library, https://github.com/cityofaustin/atd-airflow/blob/frank/airflow-three/not-yet-upgraded-dags/dts_open_311_scrape.py. The moped DAG you mentioned, if I'm looking at the right one, actually has code in its own tasks to get at that last-run-date, and it's likely it will need to be updated to use the new airflow context task instance's methods or structure to get at that date. The `time.py` library will be a helpful resource on that.

dags/test_docker_failure.py
frankhereford: It's very missable! I did some digging around, and it turns out you can pass a `md_docs` to both the DAG instantiation / decorator and to the task one as well, if you have task specific info. You can find it for the DAG here, and in a similar place (and button) for individual tasks:

Image

config/airflow.cfg
frankhereford: Forgive me if this confusing name has me confused, but I think this controls the number of simultaneous runs that a DAG can be having at once. I think that all of our DAGs are conceptualized around the idea that they are running or not, and they don't have any concept of two of the same DAG running at once. 

I think you'll be able to get rid of that overriding parameter in that DAG you linked, or not, it will just re-set an already set value.

config/airflow.cfg
```diff
@@ -0,0 +1,2314 @@
...
+simple_auth_manager_passwords_file = /opt/airflow/users/simple_auth_manager_passwords.json 
+
+# This defines the maximum number of task instances that can run concurrently per scheduler in
+# Airflow, regardless of the worker count. Generally this value, multiplied by the number of
+# schedulers in your cluster, is the maximum number of task instances with the running
+# state in the metadata database. The value must be larger or equal 1.
+#
+# Variable: AIRFLOW__CORE__PARALLELISM
+#
+parallelism = 2
```
johnclary: This looks a bit low at first blush, but I also don't know how much parallelism is going on in our current setup 👀

config/airflow.cfg
```diff
@@ -0,0 +1,2314 @@
...
+# the number of tasks that is running concurrently for a DAG, add up the number of running
+# tasks for all DAG runs of the DAG. This is configurable at the DAG level with ``max_active_tasks``,
+# which is defaulted as ``[core] max_active_tasks_per_dag``.
+# 
+# An example scenario when this would be useful is when you want to stop a new dag with an early
+# start date from stealing all the executor slots in a cluster.
+#
+# Variable: AIRFLOW__CORE__MAX_ACTIVE_TASKS_PER_DAG
+#
+max_active_tasks_per_dag = 2
```
johnclary, line 105: 👀

config/airflow.cfg
johnclary: Echoing that this sounds desirable to me.

config/airflow.cfg
johnclary: Interesting, I didn't realize or forgot that Mike used this functionality in our v2 setup. I am curious to know what will happen when the scheduler tries to start a second run of a DAG. Does it queue? Does it fail? Does the running instance fail?

config/airflow.cfg
```diff
@@ -0,0 +1,2314 @@
...
+
+# (experimental) The maximum number of consecutive DAG failures before DAG is automatically paused.
+# This is also configurable per DAG level with ``max_consecutive_failed_dag_runs``,
+# which is defaulted as ``[core] max_consecutive_failed_dag_runs_per_dag``.
+# If not specified, then the value is considered as 0,
+# meaning that the dags are never paused out by default.
+#
+# Variable: AIRFLOW__CORE__MAX_CONSECUTIVE_FAILED_DAG_RUNS_PER_DAG
+#
+max_consecutive_failed_dag_runs_per_dag = 0
```
johnclary, line 129: Cool to see this as an option, even though we generally don't want DAGs to pause.

config/airflow.cfg
```diff
@@ -0,0 +1,2314 @@
...
+# Variable: AIRFLOW__CORE__DEFAULT_TASK_RETRIES
+#
+default_task_retries = 0
+
+# The number of seconds each task is going to wait by default between retries. Can be overridden at
+# dag or task level.
+#
+# Variable: AIRFLOW__CORE__DEFAULT_TASK_RETRY_DELAY
+#
+default_task_retry_delay = 300
```
johnclary: Five minutes feels like a long time to wait to retry. But NBD if this value is what our current v2 setup uses.

config/airflow.cfg
Charlie-Henry: I guess I'd rather it'd automatically start DAGs if they're new and git pulled? I do wonder what this means for local dev if when we spin it up it'll start scheduling every DAG we have?

dags/log_last_run_time.py
johnclary: Adding that most of the Knack ETLs use the `get_date_filter_arg` util in `utils/knack.py` which uses `prev_start_date_success`.

haproxy/haproxy-production.cfg
```diff
@@ -33,41 +33,21 @@ defaults
...
 
-# a web application used to monitor the workers' statuses
-backend workers
-  option httpchk GET /
-  default-server inter 5s maxconn 50
-  server workers flower:5555 check
```
johnclary, line 73: 🥀

haproxy/haproxy-production.cfg
johnclary, line 1: 🟥 🟥 🟥 🟥 🚀

docker-compose.yaml
```diff
@@ -245,27 +245,12 @@ services:
...
-      airflow-init:
-        condition: service_completed_successfully
+
 
 volumes:
-  # postgres db data persisted here
-  postgres-db-volume:
+  v3-postgres-db-volume:
   # volume for Knack SSL certs which will be mounted to DockerOperator tasks for 311 integration
   knack-certs:
```
johnclary: @frankhereford — I forgot we had a special setup for these self-signed certs. This can definitely be purged!

johnclary (commented):
This is looking so good, Frank. Thanks for getting us up and running. I have just reviewed code so far but wanted to release these comments.


I'll need to revisit my local setup later because I'm bumping into an issue that needs a closer look.

docker-compose.yaml
chiaberry: 🧹 more knack 311 cleanup!

Charlie-Henry (approved):
I got it spinned up and also got the Slack connection to work. I edited the 1pass item to add some instructions for setting it up in airflow 3. 

I'm curious to learn more about migrating DAGs, it sounds like there's another branch I should check out.

johnclary:
Just reporting back that I was able to get my local stack up and running once I finally realized that I need to `docker compose build`. Looks good!

@frankhereford, with regard to migrating db metadata—specifically the last run dates—is Airflow's [db migration script](https://airflow.apache.org/docs/apache-airflow/stable/installation/upgrading_to_airflow3.html#step-6-deployment-managers-upgrade-your-airflow-instance) an option? I know things are bit tricky with Docker and all, but I am curious if that would save us the headache up trying to orchestrate getting DAG IDs and correct last run dates to set for all the DAGs.

And FYI in case you hadn't seen it, Airflow has a [list of import changes](https://airflow.apache.org/docs/apache-airflow/stable/installation/upgrading_to_airflow3.html#key-import-updates) in the migration guide as well.

dags/utils/knack.py
frankhereford, line 1: I swung through this file today to see how it worked and what it needed to have its functionality restored or maintained. I have:

* Updated it to the latest airflow libraries
* Added detailed logging to assess the functionality of this to the `log_last_run_time.py` DAG
* Run the above DAG multiple times, including after a clean database reset, simulating when we jump to Airflow 3 for our knack DAGs.

Everything seems to be working. I think the logic in the `knack.py` is curious, because it will use todays date if it doesn't have a last-run date, but this PR is about getting things working the way they were, not enhancing our DAGs.

I think this is ready to go. Please let me know if you have other suggestions.

dags/log_last_run_time.py
frankhereford: Thank you both for this conversation and for pointing out where things are and how they work. Big time 🙏 

Re the `knack.py`, I have ported it to the current airflow library and it seems to work. I say that it seems to work because I added support to test it in the `log_last_run_time.py` test DAG fixture, and I got good results, both on a fresh DB and one with previous runs in it. I think we're good.

config/airflow.cfg
frankhereford: 2 does seem overly conservative. I'm going to bump it to `4`, which is plenty safe to me because you know we'll be watching this machine like a hawk for a little bit after airflow 3 fires up. https://github.com/cityofaustin/atd-airflow/pull/307/commits/e23111e7ff1a293ab1b3dd9d8ca8567962542b54

config/airflow.cfg
frankhereford: We generally serialize our DAGs, so I think of this as largely moot. If we did have DAGs that fanned out into a bunch of simultaneous processes, the practical parallelism of the orchestration would be the product of this and the number of DAGs, and we'd have to dial this down, so I set it low to start with.

config/airflow.cfg
frankhereford: It is what we had in there, I just took a peek. But it feels long to me too. I'm thinking 30 seconds. LMK if you think that's off the mark. 

https://github.com/cityofaustin/atd-airflow/pull/307/commits/8be413f87a058b799f00e06cb4801fc32061c52e

docker-compose.yaml
frankhereford: Right on - it's outta there. https://github.com/cityofaustin/atd-airflow/pull/307/commits/f796cbbca0a93dc98b6f8cf92267d3607c68b92a

dags/test_docker_failure.py
chiaberry: I want to find an excuse to use this!

chiaberry (approved):
reviewed the latest updates, 🚢

dags/test_docker_failure.py
mddilley: i just added this to the not-very-well scoped work for after the Airflow 3 migration is complete since we previously discussed how to better document DAGs https://github.com/cityofaustin/atd-data-tech/issues/26322 📚

config/airflow.cfg
mddilley: It has been a while since I've run a full replace, but I think the scheduled runs queue up. I looked at our v2 config, and [we have it set to 16](https://github.com/cityofaustin/atd-airflow/blob/6934af9cd75660215d70cd3afe7f7bdc6c6b1bf0/airflow.cfg#L47) which ~is confusing to me since~ [we saw runs queue up when the extended Docker Hub outage happened](https://austininnovation.slack.com/archives/CHZE6BC6L/p1760983685492079). _Edit: Nvm I just realized that the bright green is confirming that many were running concurrently, i think?_

Max 1 run per DAG still sounds like it lines up with our expectation to me too though.

mddilley (commented):
I re-reviewed the latest commits and added some comments, and it all looks good to me. Commenting instead of approving to help us remember this is the collector for v3. 

I added a link to the [soon to be deprecated imports](https://airflow.apache.org/docs/apache-airflow/stable/installation/upgrading_to_airflow3.html#key-import-updates) that John shared to https://github.com/cityofaustin/atd-data-tech/issues/25457 since it looks like they will deprecate after 3.1 and seem to be mostly in the DAG code.

I'm still thinking how we could automate the DAG run backfill either through SQL or scripting just to make the hand execution when we turn DAGs off/on less tedious. I'm open to any thoughts.

dags/utils/time.py
Charlie-Henry: Hey Frank, these look a lot nicer but I was struggling to migrate my DAG so would be cool just leaving these alone or would you be cool with me reviving `get_previous_run_date`?

Charlie-Henry: Actually, nevermind there are a couple of changes I'll make to these but `get_previous_run_date` is going in the bin 🗑️

PR #308, opened 2025-10-01
Port `atd_airflow_docker_image_prune` & `atd_moped_components_to_agol` to v3

frankhereford:
This PR aims to get into our DAG collector branch which is running on `02`. It initially contained the DAGs I tried to establish patterns on, and one of the recent moped 

### 🔁 Move + modernize existing DAGs
- **Moves `atd_airflow_docker_image_prune` out of `not-yet-upgraded-dags/` and back into `dags/`**
  - This DAG already existed; it’s not new functionality
  - Updated to the Airflow 3-preferred decorator style (`@dag`, `@task.bash`)
  - Uses the new `schedule` arg instead of `schedule_interval`
  - Preserves existing behavior (Slack failure callback, timeouts, prod-only scheduling)
  - Net effect: same job, just made Airflow-3-safe and re-enabled
  - 👍 This one feels safe to run to test, esp locally.

- **Moves and updates `atd_moped_components_to_agol` back into `dags/`**
  - Updates scheduling API usage for Airflow 3
  - Switches to the provider DockerOperator import path
  - Adds support for a **`dry_run` param** (conditionally appends `--dry-run`)
  - Retains incremental behavior based on the previous successful run
  - Slightly increases execution timeout to avoid spurious failures
  - One might coordinate with Mike before firing this up -- just so he's in the loop.


mddilley:
@frankhereford I am seeing AGOL layers with component row counts that aren't adding up quite right so we might need to test Airflow 3 with a different DAG that is more forgiving or has a staging dataset. I can help look for one. I did a full replace on the Moped AGOL dataset to be safe.

Anyways, so nice to see that it isn't a large diff to translate a v2 DAG to v3!!! 😎

dags/atd_moped_components_to_agol.py
```diff
@@ -132,7 +141,7 @@ def branch(params):
         tty=True,
         force_pull=True,
         mount_tmp_dir=False,
-        execution_timeout=duration(minutes=5),
+        execution_timeout=duration(minutes=10),
```
mddilley: Should we leave this at 5 minutes since the schedule is every 5 minutes? I'm thinking about recent [Charlie's observation of scheduled DAGs building up](https://austininnovation.slack.com/archives/CHZE6BC6L/p1760983685492079) and how to avoid. 

But, I am also getting confused because we have `max_active_runs` set to 1 for this DAG to let the full replace mode finish before another run fires off. Maybe this makes needing to line up the schedule and timeout less relevant to this DAG.

dags/atd_moped_components_to_agol.py
mddilley, line 1: Feeling great about moving this one over when ready! I was able to use your SQL from https://github.com/cityofaustin/atd-airflow/pull/307/ to insert a successful DAG run in the table to give the incremental mode a place to pick back up. I did one day in the past, and I saw it pick up 3 project updates that happened since the release yesterday. 

I was also able to delete all DAG runs for `dag_id = 'atd_moped_components_to_agol'` which would give us a way to clear out the mock DAG runs too (if we used the wrong lookback or something like that).

Here is the SQL that I used:

<details>
<summary>Expand me</summary>

```sql
-- Insert successful DAG run that completed one day ago
INSERT INTO dag_run (
  dag_id,
  run_id,
  logical_date,
  data_interval_start,
  data_interval_end,
  run_after,
  queued_at,
  start_date,
  end_date,
  state,
  run_type,
  triggered_by,
  triggering_user_name,
  conf,
  context_carrier,
  span_status
)
VALUES (
  'atd_moped_components_to_agol',
  (now() AT TIME ZONE 'UTC') - interval '1 day',
  (now() AT TIME ZONE 'UTC') - interval '1 day',
  (now() AT TIME ZONE 'UTC') - interval '1 day',
  (now() AT TIME ZONE 'UTC') - interval '1 day',
  (now() AT TIME ZONE 'UTC'),
  (now() AT TIME ZONE 'UTC') - interval '1 day',
  (now() AT TIME ZONE 'UTC') - interval '1 day',
  'success',
  'manual',
  'UI',
  'admin',
  '{}'::jsonb,
  'ended'
);

-- Delete all DAG runs for Moped AGOL only to start over
DELETE FROM dag_run WHERE dag_id = 'atd_moped_components_to_agol';
```

</details>

mddilley (approved):
Looks great, Frank! I was able to give the Moped AGOL DAG a starting point and see it pick up records to update in dry run mode by testing different lookback windows. This makes moving incremental DAGs from Airflow v2 to v3 feel way less intimidating and made the Airflow DB less mysterious! 😎

mddilley:
@frankhereford thanks for merging the latest into this branch. Is this one ready for review from the rest of the team?

frankhereford:
Hi @mddilley, thanks for getting this out for review. I just swung through to do that, and you're ahead of me. I am going to work on comment some to give a little bit more context. I appreciate your help today so much.

dags/atd_moped_components_to_agol.py
frankhereford: I was totally bought into your logic, but I did go back and investigate a little more. Here's some documentation summary on the top:

> You will NOT get two instances running simultaneously. max_active_runs=1 is a hard concurrency cap at the DAG run level. If the current run is still active when the next 5-minute slot fires, the scheduler simply won't create the new run. Combined with catchup=False, Airflow won't accumulate a backlog of missed runs either — it just picks up again at the next interval after the current run terminates (success or failure).

I think that makes this configuration make sense, right?

frankhereford:
@mddilley I think anytime you want to move this into the collector and fire it up, let's do it! I would be confident merging wtih just your ✅ since this is so very moped, and I know you and I would be watching real close to launch. If you're feeling it maybe do it tomorrow to get it going?

chiaberry (approved):
didnt run locally, but I have been doing my dag upgrades off this branch, so I feel comfortable 👍🏼

johnclary:
Nice!!

dags/atd_moped_components_to_agol.py
mddilley: Oh nice! This is good to know! Thanks for finding that part of the docs, Frank! 🙏 🙏 

I think that I'm still confused about our Airflow 2 settings, but this makes 100% going forward with v3 so I'm not going to get bogged down with how it worked in our old config. Won't be around for much longer. 🚀

chiaberry: I am confused too because I swear even though we have catchup=False, i see multiple dags queue up. I am going to blame v2 and optimistically believe that v3 will make more sense!

PR #310, opened 2025-10-24
Open311 Scraping DAG

Charlie-Henry:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/25051

## Associated repo
Related PR: https://github.com/cityofaustin/dts-311-reporting/pull/5

## Testing

**Steps to test:**
1. You DO need to be on VPN if you want to test `dts_csr_report_publishing.py` you do NOT need to be for `dts_open_311_scrape.py`
2. Use the above PR to create a docker image: `docker build . -t atddocker/dts-311-reporting:local`
3. Edit both DAGs to use the your local tag (`atddocker/dts-311-reporting:local`): Line 71 of `dts_open_311_scrape.py` and Line 155 of `dts_csr_report_publishing.py`
4. Comment out `force_pull=True` in both DAGs: Line 85 of `dts_open_311_scrape.py` and Line 170 of `dts_csr_report_publishing.py`
5. Trigger both DAGs and check that that they complete.
6. Retrigger `dts_open_311_scrape.py` and check that it completes and uses the previous successful run date (the script will subtract 10 minutes from this time and retrieve data as normal.)

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/dts_open_311_scrape.py
+++ b/dags/dts_open_311_scrape.py
@@ -7,7 +7,7 @@ from datetime import timedelta
 from airflow.decorators import task
 from airflow.models import DAG
 from airflow.operators.docker_operator import DockerOperator
-from pendulum import datetime, duration
+from pendulum import datetime, duration, now
 
 from utils.onepassword import get_env_vars_task
 from utils.slack_operator import task_fail_slack_alert
@@ -71,7 +71,8 @@ with DAG(
     docker_image = "atddocker/dts-311-reporting:production"
 
     env_vars = get_env_vars_task(REQUIRED_SECRETS)
-    date_arg = get_previous_run_date(fallback_date="2025-10-22")
+    one_day_ago = now("America/Chicago").subtract(days=1)
+    date_arg = get_previous_run_date(fallback_date=one_day_ago.to_iso8601_string())
 
     t1 = DockerOperator(
         task_id="open311_to_socrata",
--- a/dags/dts_open_311_scrape.py
+++ b/dags/dts_open_311_scrape.py
@@ -65,7 +65,7 @@ with DAG(
         "*/5 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None
     ),
     dagrun_timeout=timedelta(minutes=5),
-    tags=["repo:dts-311-reporting", "socrata", "csr", "open311"],
+    tags=["repo:dts-311-reporting", "socrata", "311", "open311"],
     catchup=False,
 ) as dag:
     docker_image = "atddocker/dts-311-reporting:production"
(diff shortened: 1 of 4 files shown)


dags/utils/time.py
```diff
@@ -6,7 +6,7 @@
     task_id="get_previous_run_date",
     multiple_outputs=True,
 )
-def get_previous_run_date(**context):
+def get_previous_run_date(fallback_date="1970-01-01", **context):
```
Charlie-Henry, line 9: A search of this repo shows that no DAGs are currently using this helper function, so we don't have any other scripts to test for backwards compatibility but in theory my changes should not break anything.

dags/dts_open_311_scrape.py
```diff
@@ -0,0 +1,88 @@
...
+        "*/5 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None
+    ),
+    dagrun_timeout=timedelta(minutes=5),
+    tags=["repo:dts-311-reporting", "socrata", "csr", "open311"],
+    catchup=False,
+) as dag:
+    docker_image = "atddocker/dts-311-reporting:production"
+
+    env_vars = get_env_vars_task(REQUIRED_SECRETS)
+    date_arg = get_previous_run_date(fallback_date="2025-10-22")
```
Charlie-Henry: The first time this DAG is run it will use this date. This has to be coordinated with the DAG timeout. Current API limits us to 1000 records per minute, so something recent will complete within the 5 minutes timeout when we merge. This feels a little janky but I'm not sure how else to go about this.

chiaberry (commented):
Tested the dts_311_report_publishing dag and got green squares

Trying to run the dts_open_311_scraping dag resulted in timeouts

Image

 since it was taking more than 5 minutes to backfill all nine days

dags/dts_open_311_scrape.py
Charlie-Henry: After @chiaberry's test I just make the default start date to -1 days in the past 👍

chiaberry (approved):
all green with the updated previous_run_date. If this were to fail for more than a day, we'd just have to run it locally to backfill the data?

Charlie-Henry:
> all green with the updated previous_run_date. If this were to fail for more than a day, we'd just have to run it locally to backfill the data?

No, this is only used when there is no run history. If it fails normally it will use the previous successful run's date which can be more than 1 day in the past 👍 If it's down for a long time we could in theory get a backlog of requests that would go longer than the 5 minute timeout, though. 

Bill S. at 311 seems to think we can get 50 requests/min limit he's just working with the vendor to get it to work which will also help our timeout issue.

johnclary:
I love that the vendor is helping us scrape their API 👍

PR #314, opened 2025-11-21
Create eCAPRIS funding ETL DAG

mddilley:
## Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/25665

## Associated repo
https://github.com/cityofaustin/atd-moped/pull/1722

## Testing

**Steps to test:**
1. Follow the test steps in https://github.com/cityofaustin/atd-moped/pull/1722
2. The changes to ports in this PR are to fix conflicts with the Moped local stack and make testing easier. I will change them back to 8080 before merge. **Done ✅**

---

#### Ship list

- [x] Code reviewed

### DIFF
--- a/dags/atd_moped_ecapris_funding_sync.py
+++ b/dags/atd_moped_ecapris_funding_sync.py
@@ -1,4 +1,4 @@
-# Test locally with: docker compose run --rm airflow-cli dags test atd_moped_ecapris_status_sync
+# Test locally with: docker compose run --rm airflow-cli dags test atd_moped_ecapris_funding_sync
 
 from os import getenv
 
--- a/dags/atd_moped_ecapris_funding_sync.py
+++ b/dags/atd_moped_ecapris_funding_sync.py
@@ -1,0 +1,84 @@
+# Test locally with: docker compose run --rm airflow-cli dags test atd_moped_ecapris_status_sync
+
+from os import getenv
+
+from airflow.models import DAG
+from airflow.operators.docker_operator import DockerOperator
+from pendulum import datetime, duration
+
+from utils.onepassword import get_env_vars_task
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
+    "start_date": datetime(2015, 1, 1, tz="America/Chicago"),
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 0,
+    "retry_delay": duration(minutes=5),
+    "on_failure_callback": task_fail_slack_alert,
+}
+
+REQUIRED_SECRETS = {
+    "SOCRATA_API_KEY_ID": {
+        "opitem": "Socrata Key ID, Secret, and Token",
+        "opfield": "socrata.apiKeyId",
+    },
+    "SOCRATA_API_KEY_SECRET": {
+        "opitem": "Socrata Key ID, Secret, and Token",
+        "opfield": "socrata.apiKeySecret",
+    },
+    "SOCRATA_TOKEN": {
+        "opitem": "Socrata Key ID, Secret, and Token",
+        "opfield": "socrata.appToken",
+    },
+    "SOCRATA_ENDPOINT": {
+        "opitem": "Socrata Key ID, Secret, and Token",
+        "opfield": "socrata.endpoint",
+    },
+    "HASURA_ENDPOINT": {
+        "opitem": "Moped Hasura Admin",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.Endpoint",
+    },
+    "HASURA_ADMIN_SECRET": {
+        "opitem": "Moped Hasura Admin",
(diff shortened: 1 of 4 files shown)


haproxy/haproxy.cfg
mddilley, line 1: I'll switch all these 8082s back to 8080 before merge.

dags/atd_moped_ecapris_funding_sync.py
```diff
@@ -0,0 +1,126 @@
...
+    "HASURA_ENDPOINT": {
+        "opitem": "Moped Hasura Admin",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.Endpoint",
+    },
+    "HASURA_ADMIN_SECRET": {
+        "opitem": "Moped Hasura Admin",
+        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.Admin Secret",
+    },
+    "FUNDING_DATASET_IDENTIFIER": {
+        "opitem": "Moped ETLs",
```
mddilley, line 52: this is a new entry in our vault. I followed the pattern of the **Vision Zero ETLs** entry since we've gotten away without needing a Moped ETL-specific entry so far, but now we need one!

dags/atd_moped_ecapris_funding_sync.py
```diff
@@ -0,0 +1,126 @@
...
+        docker_conn_id="docker_default",
+        auto_remove="force",
+        command=f"python3.14 ecapris_funding_sync.py",
+        environment=env_vars,
+        tty=True,
+        force_pull=True,
+        mount_tmp_dir=False,
+    )
+
+    env_vars >> branch_task >> [ecapris_funding_sync_dry_run, ecapris_funding_sync]
```
frankhereford, line 121: I'm curious what you think Mike -- this isn't a request at all, just me bouncing an idea around.

What would you think about simplifying the branching DAG here and drying up this code by leveraging jinja templating in the command, something like this:

```
    ecapris_funding_sync = DockerOperator(
        task_id="ecapris_funding_sync",
        image=docker_image,
        auto_remove="force",
        command="python3.14 ecapris_funding_sync.py{% if params.dry_run %} -n{% endif %}",
        environment=env_vars,
        tty=True,
        force_pull=True,
        mount_tmp_dir=False,
    )

    env_vars >> ecapris_funding_sync
```

I know it's not the prettiest command line because we don't see these templating delimiters much (it reminds me of PHP 🤢), but it eliminates the requirement that we maintain all the other parameters of the DAG creation to be identical. 

A downside is that you loose some information that was available at a glance because there's no longer two tasks, one which turns pink because it was not executed and one that turns green like normal. In the next image, you can see that there are not two tasks anymore, they are combined, and pls ignore the pink square, that was before i tested this idea.

Image

It's not an obvious slam dunk of an idea, but I do think it's worth exploring because it drys up that code and eliminates the complexity of the branching in here, by eliminating the branch task entirely.

frankhereford (approved):
Thank you for this Mike. I am ✅ because I think this works great, but I have an idea that I think is worth exploring in a comment. Can't wait to hear what you think.

chiaberry (approved):
a breeze to test concurrently with the moped PR

I dont feel super strongly about the two tasks vs one task, but one thing that I think is beneficial from having the dry run task split out from the normal task, is being able to see at a glance what was run and with which parameters

also, big Wicked vibes, that movie marketing is *everywhere* :) 
Image

Charlie-Henry (approved):
All green for me! 🩷 🟢

dags/atd_moped_ecapris_funding_sync.py
mddilley: I agree the repetition isn't great so I like aiming to dry this up too. Like Chia mentioned, I was going for the ability to see whether a run was dry or live without looking at parameters or logging. That has been helpful for the Moped AGOL ETL. But, that doesn't mean that we have to repeat all the code. I think that I could create a shared config dictionary and unpack in the task branches. I'll this out tomorrow! 🙏

```python
common_docker_config = {
        "image": docker_image,
        "docker_conn_id": "docker_default",
        "auto_remove": "force",
        "environment": env_vars,
        "tty": True,
        "force_pull": True,
        "mount_tmp_dir": False,
}

ecapris_funding_sync_dry_run = DockerOperator(
        task_id="ecapris_funding_sync_dry_run",
        command="python3.14 ecapris_funding_sync.py -n",
        **common_docker_config,
)
```

haproxy/haproxy.cfg
johnclary: All good, but we might consider moving the local Moped Hasura container to 8082 permanently. Or at least I think that port is available in a local dev port namespace 🤠

johnclary (approved):
The parameterized DAG is so cool to me. Whichever way yo want to go with potentially cleaning up the dry run task, this looks great and works great 👍 🚢

dags/atd_moped_ecapris_funding_sync.py
frankhereford: Nice solution! ✨

haproxy/haproxy.cfg
mddilley: yup, agreed! this feels error-prone since i'm just now getting back to restoring and could have easily forgotten and merged. i made a Moped issue since this the second time i've done this in an Airflow PR. 🙏 https://github.com/cityofaustin/atd-data-tech/issues/25865

PR #319, opened 2026-01-26
remove several tasks from parking ETL

Charlie-Henry:
Removing several tasks that are no longer needed or always fail (in the case of passport 👎 ). Pretty much just leaves in tact the part that loads parking meter transaction data to the [socrata dataset](https://data.austintexas.gov/Transportation-and-Mobility/Parking-Transactions/5bb2-gtef/about_data). A whole cleanup of the parking-data repo is probably in store as well.

## Associated issues
[Somewhat related](https://github.com/cityofaustin/atd-data-tech/issues/26488)

## Associated repo
https://github.com/cityofaustin/atd-parking-data/

## Testing

**Steps to test:**
Feel free to trigger this DAG but it will take awhile to run due to flowbirds onerous API rate limit. Maybe let people know you are testing as this API limit is global and you can step on the production ETL or others testing. You can also just trust me bro that I tested this and it worked 👍 ✅ 

I believe you have to be on city VPN for it to work.

`docker compose run --rm airflow-cli dags test atd_parking_data`

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/atd_parking_data.py
+++ b/dags/atd_parking_data.py
@@ -106,7 +106,7 @@ REQUIRED_SECRETS = {
 
 with DAG(
     dag_id="atd_parking_data",
-    description="Scripts that download and process parking data for finance reporting.",
+    description="Scripts that download and process parking data.",
     default_args=default_args,
     schedule_interval="35 8 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
     tags=["repo:atd-parking-data", "parking", "socrata", "postgrest"],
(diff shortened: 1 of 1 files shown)


chiaberry (approved):
I trust you!

frankhereford (approved):
I have not run this -- I don't want to get involved in that slow API. Thank you for mentioning that! If you like it, I'm on board.

Looking at the code, I do see 9 removals from the controlling data structure and 9 removals from the `docker_tasks` array that is built. 👍 🚢

johnclary (commented):
You love to see it. Thanks @Charlie-Henry !

PR #322, opened 2026-02-12
Move some knack services dags to v3

chiaberry:
## Associated issues

I was testing airflow 3 and tested by moving these

## Testing

run locally and kick them off if you want
`docker compose up`

### DIFF
--- a/dags/atd_knack_dms.py
+++ b/dags/atd_knack_dms.py
@@ -3,7 +3,7 @@
 from os import getenv
 
 from airflow.models import DAG
-from airflow.operators.docker_operator import DockerOperator
+from airflow.providers.docker.operators.docker import DockerOperator
 from pendulum import datetime, duration
 
 from utils.onepassword import get_env_vars_task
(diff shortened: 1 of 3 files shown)


dags/utils/onepassword.py
```diff
@@ -5,8 +5,8 @@
 
 from os import getenv
 
-from airflow.decorators import task
-from onepasswordconnectsdk.client import Client, new_client
+from airflow.sdk import task
+from onepasswordconnectsdk.client import new_client
```
chiaberry, line 9: Client wasnt being used, so I removed it
I also updated the import because it was showing a deprecation warning

dags/atd_knack_dms.py
chiaberry, line 1: I chose this one because I know its a small dataset and then i realized that it doesnt even take the date filter. which is what i was trying to test.

mddilley (approved):
Looks great, and I saw ✅✅✅✅ across the board for both DAGs. 🚢

PR #323, opened 2026-02-24
Update eCAPRIS sync ETLs to populate staging database tables

mddilley:
## Associated issues
Closes https://github.com/cityofaustin/atd-data-tech/issues/25756

This PR updates the eCAPRIS funding and status DAGs with parameters to target the local database for testing and also the staging database so we can upsert records into the staging environment from either the local or production Airflow dashboard. It isn't crucial for staging to be up-to-date with production so I opted not to schedule them. And, we will now be able to test eCAPRIS features in staging and deploy previews for feature development and release testing.

I decided to branch from the Airflow 3 work so that I could migrate these two DAGs in the process.

## Associated repo
https://github.com/cityofaustin/atd-moped/pull/1782

## Testing

**Steps to test:**
**FYI:** During this testing, the local stacks will be using the Docker images with "development" tags. I pushed the latest production code up to these yesterday since we want to use the latest version of the ETL scripts.

1. Start up your local Moped stack from a production snapshot using the branch in https://github.com/cityofaustin/atd-moped/pull/1782 since that PR makes updates to help run the Moped stack and Airflow stack at the same time local without port conflicts.
2. Start the Airflow stack from this branch: `docker compose up`
3. Connect to the VPN since you need to reach the FSD Oracle DB for the statuses ETL
4. Use your DB explorer to connect to the Moped local DB and run the following statements to clear out existing eCAPRIS funding records and statuses so we can test populating them again with the DAGs
```sql
TRUNCATE ecapris_subproject_statuses;
TRUNCATE ecapris_subproject_funding;
```
5. Now that you have no rows in the statuses table, go to http://localhost:8080/dags/atd_moped_ecapris_status_sync and click the "Trigger" button in the top right
6. Select "development" for the value of `target_database`. This will tell the DAG to use the "development" key when accessing the Hasura endpoint in the **Moped Hasura Admin** secret entry. Click "Trigger" and watch it work on populating statuses into `ecapris_subproject_statuses` table. Check your local DB, and you should have close to 10,000 rows now.
7. Test using the "staging" value from the dropdown, and it will upsert into the staging database. You can run this as many times as you want. There will already be rows in the staging tables since I tested the DAGs already.
8. Now, go to http://localhost:8080/dags/atd_moped_ecapris_funding_sync and repeat the same steps of using the "development" parameter for `target_database`, see new rows in `ecapris_subproject_funding` in you local DB, and the "staging" to run on staging (which can also be run as many times as needed).

---

#### Ship list

- [x] Code reviewed
- ~[ ] Add note to 1PW secrets moved to API vault and check for duplicates~


dags/atd_moped_ecapris_funding_sync.py
```diff
@@ -23,36 +22,40 @@
...
-    },
-    "FUNDING_DATASET_IDENTIFIER": {
-        "opitem": "Moped ETLs",
-        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.FUNDING_DATASET_IDENTIFIER",
-    },
-}
+
+@task
+def get_required_secrets(params):
+    target_database = params["target_database"]
```
mddilley, line 36: this is the big difference between how the old and new DAGs get secrets. This is now a task that uses context to get the target database parameter chosen at runtime.

dags/atd_moped_ecapris_funding_sync.py
```diff
@@ -86,12 +89,25 @@ def branch(params):
...
+    params={
+        "dry_run": Param(default=False, type="boolean"),
+        "target_database": Param(
+            default=DEPLOYMENT_ENVIRONMENT,
+            enum=(
+                ["production", "staging"]
+                if DEPLOYMENT_ENVIRONMENT == "production"
+                else ["staging", "development"]
+            ),
+            description="Target Moped database. Defaults to the current deployment environment.",
```
mddilley, line 110: I initially made the "production" option available no matter what environment the Airflow stack was running in, but I think it makes it too easy to target production from the local stack which we've tried to avoid in the past.

So, this makes targeting the production database available as default for scheduled runs in the production environment and production and staging available as parameters for manually triggered runs. And, it makes production unavailable to target in the development (local) stack. ...unless you modify the local code enough which is true for other DAGs as well.

dags/atd_moped_ecapris_funding_sync.py
chiaberry: Should I update my branch with the updated dags to also follow this pattern?

chiaberry: Or is this only for dags with params?

mddilley: ooo good question. this was to specifically address params so I think the other dags are fine as-is. 

This fixed an issue where I was trying to reference the param value in the dag body, and it was always coming back as the default, "development", since it wasn't reading my chosen parameter correctly at runtime.

dags/atd_moped_ecapris_funding_sync.py
chiaberry: thank you for this safeguard!

chiaberry (approved):
worked like a charm! I tested locally and also to staging. And the dry run flag too.

dags/utils/onepassword.py
```diff
@@ -5,7 +5,7 @@
 
 from os import getenv
 
-from airflow.decorators import task
+from airflow.sdk import task
```
frankhereford, line 8: Thank you for catching this!

dags/atd_moped_ecapris_funding_sync.py
frankhereford: I think this is an elegant way to expose more control over where the DAG targets its output. I like how it builds on the old, simpler way the nonparameterized DAGs figure out which environment to target but gives the whole system a big level up with runtime configurability. It feels like you're working with Airflow's opinions, not against them, and it all just works real good.

frankhereford (approved):
Image

This is great work Mike, big time! Thanks for the detailed testing instructions, that would have been a lot to puzzle out without your super clear directions. I think you're laying down patterns here that will serve us long into the future. 🚢 🚢 🚢

dags/atd_moped_ecapris_funding_sync.py
johnclary: 🙏

dags/atd_moped_ecapris_funding_sync.py
johnclary: Yeah, very nice pattern here 🚀

johnclary (commented):
Just commenting here because I didn't run this but the code looks great 🚀

PR #325, opened 2026-03-09
Migrate some Knack signs and markings DAGs to Airflow 3

mddilley:
## Associated issues

Moves:
- atd_knack_markings_attachments
- atd_knack_signs_markings_reimbursements
- atd_knack_signs_markings_time_logs
- atd_knack_signs_work_order_materials
- atd_knack_signs_work_order_attachments
- atd_knack_signs_work_order_specifications
- atd_knack_signs_work_orders

## Testing

**Steps to test:**
1. Start the local Airflow v3 stack and connect to the VPN which you need to be on to reach the Postgrest endpoint 
2. Get into a psql shell in the Airflow v3 container h/t @frankhereford 🙏 
```shell
docker compose exec --user airflow airflow-dag-processor airflow db shell;
```
3. Run the following statement on the Airflow v3 DB to backfill successful run dates for the DAGs that I migrated in this PR:
```sql
-- Insert successful DAG runs by ids that completed one hour ago
-- We probably can remove some of the columns in this insert but this works well enough
INSERT INTO dag_run (
  dag_id,
  run_id,
  logical_date,
  data_interval_start,
  data_interval_end,
  run_after,
  queued_at,
  start_date,
  end_date,
  state,
  run_type,
  triggered_by,
  triggering_user_name,
  conf,
  context_carrier,
  span_status
)
SELECT
  dag_id,
  (now() AT TIME ZONE 'UTC') - interval '1 hour',
  (now() AT TIME ZONE 'UTC') - interval '1 hour',
  (now() AT TIME ZONE 'UTC') - interval '1 hour',
  (now() AT TIME ZONE 'UTC') - interval '1 hour',
  (now() AT TIME ZONE 'UTC'),
  (now() AT TIME ZONE 'UTC') - interval '1 hour',
  (now() AT TIME ZONE 'UTC') - interval '1 hour',
  'success',
  'manual',
  'UI',
  'admin',
  '{}'::jsonb,
  'ended'
FROM unnest(ARRAY[
  'atd_knack_markings_attachments',
  'atd_knack_signs_markings_reimbursements',
  'atd_knack_signs_markings_time_logs',
  'atd_knack_signs_work_order_materials',
  'atd_knack_signs_work_order_attachments',
  'atd_knack_signs_work_order_specifications',
  'atd_knack_signs_work_orders'
]) AS dag_id;
```
4. Go to http://localhost:8080/dag_runs and you should see the top 7 most recent dag runs match the 7 DAG ids in the insert statement and are all marked as successful
5. Go through the list of DAGs above and you can test one by one by manually triggering them in the Airflow dashboard

---

#### Ship list

- [ ] Code reviewed


mddilley:
@johnclary @chiaberry I already chatted with Chia about this, but I wanted to check with y'all first to make sure the 1 hour backfill of previous successful runs for these DAGs feels right for testing on prod before I request reviews from the rest of the team.

Also, do y'all think there is a blanket lookback window that would work for all Knack Services DAGs? I'm thinking about us avoiding the 1st of the month full replace troubles and the [street_segment_updater](https://austininnovation.slack.com/archives/CHZE6BC6L/p1767973176148089) that we'd want to avoid running a full replace on.

Edit: Feel free to request review from others who would have close knowledge about the Knack services scripts. 🙏

mddilley:
just added @Charlie-Henry too for feedback on the lookback window after I saw some recent PRs with his name in atd-knack-services 🙏

Charlie-Henry:
I think this is a clever way to test it! I was able to see the 7 DAGs with all successful runs for me. 

I guess it _might_ still be possible with some poor timing we could create extra dupe records in Knack with the purchase request copier and signal PM copier scripts by stepping on the toes of the prod DAG, but the these I see here I don't think are a big deal as they should all handle primary keys.

Screenshot 2026-03-11 at 12 00 22 PM

mddilley:
Alright, I tested these with the 1 hour lookback and all looks good. Thanks @Charlie-Henry! I'm going to request the rest of the team for review now and update test instructions.

dags/atd_knack_markings_attachments.py
```diff
@@ -2,14 +2,19 @@
...
 
 from utils.onepassword import get_env_vars_task
 from utils.knack import get_date_filter_arg
 from utils.slack_operator import task_fail_slack_alert
 
+doc_md = """
+## Testing and troubleshooting
+Can run on production with previous successful run or 1 hour look-back window if missing
+"""
+
```
Charlie-Henry, line 24: Might be good for these knack services dags to add that you need to have your IP whitelisted in AWS or be on VPN? At least I think this is the case in order to connect to postgrest

Also, I'm a little confused as to what this means. Maybe a little reword like: **⚠️ Warning:** Running this DAG with no previous run history is not recommended as you'll wipe out thousands of records!! And maybe add a link to some docs on how to add the dummy run history? 

I wonder if these get really long if we should have a docs subdirectory where we'd store these markdown files, and then it'd also be more easily viewable in github?

Charlie-Henry (approved):
I left a few comments about the docs but the DAGs all tested successfully ✅

dags/atd_knack_markings_attachments.py
mddilley: Thank you for this feedback, Charlie, and I think that I was trying to answer two questions at once:
- How do I test this?
- What can I do if this fails in production?

I'll add the note about VPN or allow list and break this into two sections: testing and troubleshooting.

I like the warning and linking to the readme where we could capture the psql command and SQL insert used to test these and jumpstart them. Agree that more detail is better! 🙏

dags/atd_knack_markings_attachments.py
```diff
@@ -2,14 +2,27 @@
...
+doc_md = """
+⚠️ Warning: Running this DAG with no previous run history is not recommended since it will replace thousands of records!
+
+## Troubleshooting
+Trigger the DAG again (as long as there is a previous successful run to pick back up on incremental updates) to address any connection errors or timeouts
+
+## Testing
+**Need VPN access or addition to security group allow list to reach Postgrest**
+
+To insert a previous successful DAG run, see [README](./README.md#inserting-a-previous-dag-run-to-resume-incremental-runs-using-a-look-back-window)
```
mddilley: Just had a thought - this link shows as a hyperlink that 404s when looking at these in the web dashboard, but I had intended to document the path in the repo when developing a dag. Not sure what is best, thoughts? 🤔

chiaberry: maybe omit the link and instead have it say "see atd-airflow readme" or including the section name but not as a link.

dags/atd_knack_signs_materials.py
```diff
@@ -1,13 +1,26 @@
...
+## Troubleshooting
+Trigger the DAG again (as long as there is a previous successful run to pick back up on incremental updates) to address any connection errors or timeouts
+
+## Testing
+**Need VPN access or addition to security group allow list to reach Postgrest**
+
+To insert a previous successful DAG run, see [README](./README.md#inserting-a-previous-dag-run-to-resume-incremental-runs-using-a-look-back-window)
+
+"""
+
```
chiaberry, line 22: now that ive seen them on multiple dags, should this be a constant in a utils file that is imported? 

Or is the point to also see the warning on the dag in the code and not just in the browser?

chiaberry (approved):
Thank you so much for figuring out a path forward for these temperamental/delicate dags. And I am happy to see we are using the docs feature!

🚢

dags/atd_knack_markings_attachments.py
mddilley: cool! thank you for the suggestion, and I'm gonna do it. 🙏

dags/atd_knack_signs_materials.py
mddilley: Good point! Many of the Knack services DAGs are going to need the same info. I'll move to the /utils/knack.py file. 🙏 🙏

PR #327, opened 2026-03-11
Charlie Airflow three migration 

Charlie-Henry:
## Associated issues

## Associated repo
https://github.com/cityofaustin/atd-airflow

## Testing

**Steps to test:**
- Generally gotta be on city VPN
- None of these dags have issues with needing a previous run date.
- `atd_parking_data` can take a while to run and has very onerous rate limit of 1 req/min so you can step on either the prod ETL or someone else running it
- Other than that, feel free to trigger any of these
---

#### Ship list

- [ ] Code reviewed


dags/atd_parking_data.py
```diff
@@ -104,21 +103,27 @@
...
 ) as dag:
     env_vars = get_env_vars_task(REQUIRED_SECRETS)
-
-    # default to the last 14 days of transactions
-    prev_exec = "{{ (prev_start_date_success - macros.timedelta(days=14)).strftime('%Y-%m-%d') if prev_start_date_success else (execution_date - macros.timedelta(days=14)).strftime('%Y-%m-%d')}}"
+    three_days_ago = now("America/Chicago").subtract(days=3)
+    prev = get_previous_success_start_time(
+        fallback_date=three_days_ago.to_iso8601_string()
+    )
+    prev_exec = format_start_date(prev)
```
Charlie-Henry, line 146: Changed this to 3 days in the past as the fallback date when there is no run history. Also now am taking advantage of the time utils

dags/utils/time.py
Charlie-Henry, line 1: A few small changes to these utils. The only DAGs that use time.py are included in this PR. 
- dts_open_311_scrape
- atd_parking_data

not-yet-upgraded-dags/atd_trail_counters.py
Charlie-Henry, line 1: Both of these were paused in prod. Not needed.

dags/atd_finance_data_fdu_program_tagging.py
```diff
@@ -1,10 +1,7 @@
-# test locally with: docker compose run --rm airflow-cli dags test atd_finance_data_fdu_program_tagging
```
Charlie-Henry, line 1: I couldn't get this command to work with airflow 3 👎, kept saying it couldn't find our `utils` directory when importing DAGs.

not-yet-upgraded-dags/atd_trail_counters.py
Charlie-Henry: idk why this says its outdated now, github is AI slop now

dags/atd_finance_data_fdu_program_tagging.py
mddilley: thanks for the reminder about these commands. i also can't run anymore, and I think it has to do with the our [modules configuration](https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/modules_management.html#inspecting-your-pythonpath-loading-configuration). Their docs show the dags folder being available, but I don't see ours listed when I run `docker compose run --rm airflow-cli info`.

I created https://github.com/cityofaustin/atd-data-tech/issues/27358 to fix because I think that we should hold onto our ability to run these commands that we have added to many DAGs. cc @frankhereford

mddilley (approved):
I was able to test all of these (except I skipped atd_parking_data to not mess it up and dts_maximo_reporting due to VPN access), and all looks great! Thanks for flagging the cli command regression. Hoping we can fix and keep. 🚢  

Also wanted to mention that we touched on capturing info about how to troubleshoot these in dev refinement this week (aka "never run this one with a full replace" or similar). I'm going to add something to the [doc_md](https://airflow.apache.org/docs/apache-airflow/stable/tutorial/fundamentals.html#adding-dag-and-tasks-documentation) of the DAGs that I have an open PR for on Monday but not sure what that looks like yet.

dags/atd_finance_data_fdu_program_tagging.py
frankhereford: I totally agree, we need these commands to work as written or have a tool that can run them as written. Something like that. They are a really good testing resource, it's a good pattern we should keep it. I'm looking forward to thinking about this more on Monday -- thanks for surfacing this, it's 100% important.

chiaberry (approved):
reviewed the code and tested running a few of them 🚢 

bummer about the command line behavior changing, thank Mike for opening an issue for it

Charlie-Henry:
Unless anyone wants to check out all the DAG docs I've added, I think we're good to migrate this batch of scripts to airflow 3

chiaberry:
The docs look great!

PR #334, opened 2026-03-30
move 5 more dags to airflow 3

chiaberry:
Another small set of dags to move over. 

I have a thread in slack about my issues running the kits dags locally, putting this out for review in the hopes maybe its just a me problem? 

## Testing

The dags are fine to kick off without previous run times. 

---

#### Ship list

- [ ] Code reviewed


dags/atd_knack_arterial_managment_locations.py
chiaberry, line 1: this one takes a date filter, but even on a full replace its only about 2 minutes

dags/atd_knack_corridor_retiming.py
chiaberry, line 1: another one that takes under 2 minutes on a full replace

dags/atd_kits_dms_message.py
```diff
@@ -53,8 +51,9 @@
 with DAG(
     dag_id="atd_kits_dms_message_pub",
     description="Extract DMS message from KITS database and upload to Data Tracker (Knack).",
+    doc_md="In order to reach kits server, you need special VPN access or run from 8900 Cameron",
```
chiaberry: This phrasing could use some work, but the gist is there?

Charlie-Henry (approved):
On `atd_kits_dms_message_pub` I had to put a new the development api key/secret for a test data tracker app in 1pass as the old app was likely deleted at some point. The DAG config makes it fall back to using dev knack credentials when run locally. After updating 1pass ✅ 

All other DAGs ran without issue ✅

dags/atd_kits_dms_message.py
mddilley: makes sense to me! and I was able to confirm that I don't have the right VPN access to run. What do you think about adding the detail about being able to freely run all these locally without a previous successful run set?

dags/atd_knack_corridor_retiming.py
```diff
@@ -1,8 +1,8 @@
 from os import getenv
 
 from airflow.models import DAG
```
mddilley: @chiaberry just one more `models` to `sdk` that is needed here

mddilley (approved):
Ran these and the two KITS dags fail successfully because of my VPN access ❎, and the three atd-knack-services ones ran with no issues at all! I left a note about a small change to a deprecated import. 🙏 🚀

dags/atd_kits_dms_message.py
frankhereford: I might strike the 8900 part of the address of Cameron and just call it the Cameron Road Office Complex. Something like that. I wouldn't want someone to haul a laptop over to 8900 just to test when they can do it from 8700 just as easy.

mddilley: Also wanted to note that the `get_date_filter_arg` helper returned a date for all my knack services runs in this PR because it falls back to `now()`. I think that I've been overly cautious with the sql to set previous successful runs on similar knack services dags. 😅 

```
[2026-04-03 12:52:52] INFO - Filling up the DagBag from /opt/airflow/dags/atd_knack_arterial_managment_locations.py
[2026-04-03 12:52:52] INFO - Done. Returned value was: -d 2026-04-03T17:52:52.572338+00:00
```

frankhereford (approved):
They all worked great! Thank you Chia for PR here and for all your effort to get us over the Airflow V3 finish line.

mddilley:
Also wanted to note that the `get_date_filter_arg` helper returned a date for all my knack services runs in this PR because it falls back to `now()`. I think that I've been overly cautious with the sql to set previous successful runs on similar knack services dags.

```
[2026-04-03 12:52:52] INFO - Filling up the DagBag from /opt/airflow/dags/atd_knack_arterial_managment_locations.py
[2026-04-03 12:52:52] INFO - Done. Returned value was: -d 2026-04-03T17:52:52.572338+00:00
```

dags/atd_knack_amd_preventative_maintenance.py
```diff
@@ -56,10 +54,10 @@
 
 
 with DAG(
-    dag_id=f"atd_knack_amd_pm",
+    dag_id=f"atd_knack_amd_preventative_maintenance",
     description="Copies primary signal preventive maintenance records to secondary signals. Then, loads preventative maintenance work order (view_3887) records from Knack to Postgrest and Socrata.",
```
Charlie-Henry, line 58: Borrowing from @mddilley 's docs for these knack services DAGs with postgrest:

`**Need VPN access or addition to security group allow list to reach Postgrest**`

dags/atd_knack_corridor_retiming.py
chiaberry: bah thank you

chiaberry: I am finding that I have forgotten to do this in many other places, I am going to fix it here, but then pick up the other ones on my next PR

roseeichelmann (approved):
i was able to test the dags and tasks that didnt require vpn! im excited that i have local airflow dev set up -- now to figure out vpn access

PR #335, opened 2026-04-03
Charlie Airflow 3 second batch

Charlie-Henry:
10 more knack services DAGs

## Associated issues

## Associated repo

## Testing

**Steps to test:**
1. checkout this branch
2. `docker compose up`
3. Go to: http://localhost:8080/dags
4. Trigger these 10 DAGs and see that they run ✅ 

I have added a blurb if VPN should be needed (outside of reaching 1pass API) to each DAG's docs. No need to worry about run history for these.

---

#### Ship list

- [ ] Code reviewed


dags/atd_knack_purchase_request_copier.py
```diff
@@ -24,19 +37,20 @@
 REQUIRED_SECRETS = {
     "KNACK_APP_ID": {
         "opitem": "Knack Finance and Purchasing",
-        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.appId",
+        "opfield": "production.appId",
     },
     "KNACK_API_KEY": {
         "opitem": "Knack Finance and Purchasing",
-        "opfield": f"{DEPLOYMENT_ENVIRONMENT}.apiKey",
+        "opfield": "production.apiKey",
```
Charlie-Henry, line 46: This just made testing this more complicated, so running locally will just run on production rather than a development Knack app which are pretty frequently deleted and API keys go stale.

dags/atd_knack_signal_detection_status_log.py
frankhereford, line 1: FWIW, this one failed on me the first time I ran it, but came through for the second and third times. 

Image

It was an SSL error, where knack dropped the connection, so I think something transient.

frankhereford (approved):
I had one hiccup on one DAG but I don't think it's related to this work at all. These all ran great -- this is ready to 🚢 🚢 🚢. Thanks Charlie!

dags/atd_knack_secondary_signals.py
```diff
@@ -1,14 +1,26 @@
-# test locally with: docker compose run --rm airflow-cli dags test atd_knack_secondary_signals
```
mddilley, line 1: Noting that Frank fixed the stack to support testing DAGs from the command line so this can stay 🙌 

https://github.com/cityofaustin/atd-airflow/pull/336

mddilley (approved):
Ran through all 10 of these. Thanks for making these easier to test and for adding docs that cover troubleshooting really nicely. 🚀 🙏

dags/atd_knack_secondary_signals.py
```diff
@@ -1,14 +1,26 @@
...
+Refreshes primary <-> secondary traffic signal relationships.
+
+## Troubleshooting
+
+You should not need to be on VPN to reach Knack.
+
+Most of the time just re-triggering this DAG will likely resolve any issues automatically.
+
+Further investigation will likely require looking at the supplied Knack view to make sure the required fields are available.
+"""
```
johnclary, line 24: 🙏

johnclary (approved):
Looks extremely good! I already had airflow running when i checked out your branch, but I had to restart airflow in order to bring in these DAGs. Once I did that everything worked smoothly. 🚢 🚢 🚢 🚢 🚢 🚢

chiaberry (approved):
a sea of green checks!

I really like the troubleshooting steps in the docs, I am going to update the knack banner one with steps too.

PR #336, opened 2026-04-06
add PYTHONPATH environment variable to docker-compose for Airflow

frankhereford:
## Intent

This pull request makes a minor update to the Airflow service configuration in `docker-compose.yaml` by setting the `PYTHONPATH` environment variable to `/opt/airflow/dags`. This change ensures that custom Python modules in the `dags` directory can be imported by Airflow tasks.

## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/27358

## Associated repo

Airflow itself

## Testing

Spin up airflow locally and try something like this to kick off a DAG. This one is safe to run under any circumstances.

```
docker compose run --rm airflow-cli dags test log_last_run_time_development
```
## H/T 

Thank you @mddilley -- your instructions in the issue were just exactly the correct fix. I resized the issue down to a 1 from a 2. 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -19,6 +19,7 @@ x-airflow-common: &airflow-common
     AIRFLOW__API__BASE_URL: ${_AIRFLOW__API__BASE_URL:-http://localhost:8080}
 
     # Airflow stock suggested environment variables
+    PYTHONPATH: /opt/airflow/dags
     AIRFLOW__CORE__EXECUTOR: LocalExecutor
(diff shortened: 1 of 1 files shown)


mddilley (approved):
Awesome!! Worked when I tested it and on `atd_knack_amd_preventative_maintenance` as well. Thanks for fixing this so we can test without bringing up the stack and so we don't have to scrap all the nice, existing commands. 🙏 🙌 🎉

johnclary (approved):
I was able to execute the run command without issue 🚀 🚢

PR #341, opened 2026-04-21
Add test DAG and DockerOperatorWithFallback class

frankhereford:
## Note

This PR contains just the enhanced Docker Operator. A PR merges all the DAG refactoring into this, found in #342.

Introduce a new DAG 'test_docker_operator_wrapper' that utilizes the DockerOperatorWithFallback class. This class attempts to pull a Docker image and falls back to a local image if the pull fails. The new files enhance the Airflow setup for testing related Docker operations with a new test fixture.

## Associated issues

## Associated repo

Airflow itself

## Testing

Spin up the stack and kick off the test DAG in this PR. 

---

#### Ship list

- [x] Code reviewed

### DIFF
--- a/dags/test_docker_operator_wrapper.py
+++ b/dags/test_docker_operator_wrapper.py
@@ -29,6 +29,7 @@ def test_docker_operator_wrapper():
         auto_remove="force",
         tty=True,
         mount_tmp_dir=False,
+        force_pull=True,
     )
 
 
--- a/dags/test_docker_operator_wrapper.py
+++ b/dags/test_docker_operator_wrapper.py
@@ -1,0 +1,35 @@
+from __future__ import annotations
+
+import pendulum
+
+from airflow.sdk import dag
+
+from utils.docker_operator import DockerOperatorWithFallback
+
+
+@dag(
+    dag_id="test_docker_operator_wrapper",
+    schedule=None,
+    start_date=pendulum.datetime(2015, 1, 1, tz="America/Chicago"),
+    catchup=False,
+    tags=["repo:atd-airflow"],
+    default_args={
+        "owner": "airflow",
+        "retries": 0,
+        "execution_timeout": pendulum.duration(minutes=5),
+    },
+    doc_md="Hello world DAG using DockerOperatorWithFallback with the python3 image",
+)
+def test_docker_operator_wrapper():
+    DockerOperatorWithFallback(
+        task_id="hello_world",
+        image="python:3",
+        command=["python3", "-c", "print('hello world')"],
+        docker_conn_id="docker_default",
+        auto_remove="force",
+        tty=True,
+        mount_tmp_dir=False,
+    )
+
+
+test_docker_operator_wrapper()
--- a/dags/utils/docker_operator.py
+++ b/dags/utils/docker_operator.py
@@ -1,0 +1,21 @@
+from __future__ import annotations
+
+from airflow.providers.docker.operators.docker import DockerOperator
+
+
+class DockerOperatorWithFallback(DockerOperator):
+    """DockerOperator that attempts a force pull but falls back to a local image on failure."""
+
+    def execute(self, context):
+        self.log.info("Attempting to pull image %s", self.image)
+        try:
+            for output in self.cli.pull(self.image, stream=True, decode=True):
+                if isinstance(output, dict) and "status" in output:
+                    self.log.info("%s", output.get("status", ""))
+        except Exception as e:
+            self.log.warning(
+                "Could not pull image %s (%s). Using local image if available.",
+                self.image,
+                e,
+            )
+        return self._run_image()
--- a/dags/utils/docker_operator.py
+++ b/dags/utils/docker_operator.py
@@ -7,15 +7,18 @@ class DockerOperatorWithFallback(DockerOperator):
     """DockerOperator that attempts a force pull but falls back to a local image on failure."""
 
     def execute(self, context):
+        original_force_pull = getattr(self, "force_pull", False)
         self.log.info("Attempting to pull image %s", self.image)
         try:
-            for output in self.cli.pull(self.image, stream=True, decode=True):
-                if isinstance(output, dict) and "status" in output:
-                    self.log.info("%s", output.get("status", ""))
+            self.force_pull = True
+            return super().execute(context)
         except Exception as e:
             self.log.warning(
                 "Could not pull image %s (%s). Using local image if available.",
                 self.image,
                 e,
             )
-        return self._run_image()
+            self.force_pull = False
+            return super().execute(context)
+        finally:
+            self.force_pull = original_force_pull
--- a/dags/utils/docker_operator.py
+++ b/dags/utils/docker_operator.py
@@ -110,9 +110,11 @@ class DockerOperatorWithFallback(DockerOperator):
 
         self._require_run_image()
 
-        # Base DockerOperator would also pull inside _run_image when force_pull=True.
-        # We handle pull behavior here to provide fallback semantics and avoid a second pull.
-        should_pull = bool(self.force_pull)
+        # Upstream DockerOperator.execute() pulls when force_pull=True (or image is
+        # missing locally), then calls _run_image(). _run_image() itself does not read
+        # force_pull. We handle pull here to add fallback semantics, then call
+        # _run_image() directly instead of super().execute() to avoid a second pull.
+        should_pull = bool(self.force_pull or not self.cli.images(name=self.image))
 
         if should_pull:
             self.log.info("Attempting to pull image %s", self.image)
@@ -135,6 +137,9 @@ class DockerOperatorWithFallback(DockerOperator):
                 self.image,
             )
 
+        # Start the container via upstream _run_image(). Clear force_pull first as a
+        # guard: current providers only honor it in execute(), but this prevents an
+        # accidental double-pull if that logic ever moves or we delegate to super().
         original_force_pull = self.force_pull
         try:
             self.force_pull = False
--- a/dags/utils/docker_operator.py
+++ b/dags/utils/docker_operator.py
@@ -44,8 +44,8 @@ class _DockerHookWithLoginFallback(DockerHook):
 
 class DockerOperatorWithFallback(DockerOperator):
     """
-    DockerOperator that attempts a force pull but falls back to a local image on failure.
-    All other functionality is the same as DockerOperator, `force_pull` is ignored.
+    DockerOperator that can force pull with fallback to local image on pull failures.
+    All other functionality is the same as DockerOperator.
     """
 
     _RUN_IMAGE_METHOD = "_run_image"
@@ -109,21 +109,38 @@ class DockerOperatorWithFallback(DockerOperator):
         """
 
         self._require_run_image()
-        self.log.info("Attempting to pull image %s", self.image)
-        try:
-            for output in self.cli.pull(self.image, stream=True, decode=True):
-                if isinstance(output, dict) and "status" in output:
-                    self.log.info("%s", output.get("status", ""))
-        except Exception as e:
-            if not self._should_fallback_to_local_image(e):
-                raise
 
-            self.log.warning(
-                "Could not pull image %s (%s). Using local image if available.",
+        # Base DockerOperator would also pull inside _run_image when force_pull=True.
+        # We handle pull behavior here to provide fallback semantics and avoid a second pull.
+        should_pull = bool(self.force_pull)
+
+        if should_pull:
+            self.log.info("Attempting to pull image %s", self.image)
+            try:
+                for output in self.cli.pull(self.image, stream=True, decode=True):
+                    if isinstance(output, dict) and "status" in output:
+                        self.log.info("%s", output.get("status", ""))
+            except Exception as e:
+                if not self._should_fallback_to_local_image(e):
+                    raise
+
+                self.log.warning(
+                    "Could not pull image %s (%s). Using local image if available.",
+                    self.image,
+                    e,
+                )
+        else:
+            self.log.info(
+                "Skipping image pull for %s because force_pull is disabled.",
                 self.image,
-                e,
             )
-        return self._run_image()
+
+        original_force_pull = self.force_pull
+        try:
+            self.force_pull = False
+            return self._run_image()
+        finally:
+            self.force_pull = original_force_pull
 
 
 DockerOperatorWithFallback._require_run_image()
--- a/dags/utils/docker_operator.py
+++ b/dags/utils/docker_operator.py
@@ -7,18 +7,15 @@ class DockerOperatorWithFallback(DockerOperator):
     """DockerOperator that attempts a force pull but falls back to a local image on failure."""
 
     def execute(self, context):
-        original_force_pull = getattr(self, "force_pull", False)
         self.log.info("Attempting to pull image %s", self.image)
         try:
-            self.force_pull = True
-            return super().execute(context)
+            for output in self.cli.pull(self.image, stream=True, decode=True):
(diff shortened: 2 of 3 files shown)


dags/utils/docker_operator.py
```diff
@@ -0,0 +1,54 @@
...
+from airflow.providers.docker.hooks.docker import DockerHook
+from airflow.providers.docker.operators.docker import DockerOperator
+
+
+class _DockerHookWithLoginFallback(DockerHook):
+    """DockerHook that warns instead of raising when registry login fails."""
+
+    def _DockerHook__login(self, client, conn):
+        try:
+            DockerHook._DockerHook__login(self, client, conn)
```
mddilley: i did a search on Airflow's `_DockerHook__login` naming here and TIL about [name mangling](https://algomaster.io/learn/python/name-mangling) 💡 

I was thinking about if using a private method could be brittle so i'm wondering if there is a way to assert that it exists before calling and then fail loudly so we would know if a new version renamed it or changed it in some way?

That way we would know immediately that something changed before promoting a new version in our maintenance cycle.

Charlie-Henry (approved):
This test worked for me. The code is pretty hard for me to understand, but that's what we are dealing with here. 💯

dags/utils/docker_operator.py
```diff
@@ -0,0 +1,54 @@
...
+        )
+        return _DockerHookWithLoginFallback(
+            docker_conn_id=self.docker_conn_id,
+            base_url=self.docker_url,
+            version=self.api_version,
+            tls=tls_config,
+            timeout=self.timeout,
+        )
+
+    def execute(self, context):
```
mddilley, line 106: I was poking around the source code, and I think it could be valuable to link to the [original methods](https://github.com/apache/airflow/blob/dc939331f4cd90892fd201931c466dffff977a4f/providers/docker/src/airflow/providers/docker/operators/docker.py#L487) (in case we need to trace back to them later)

dags/utils/docker_operator.py
```diff
@@ -0,0 +1,54 @@
...
+        try:
+            DockerHook._DockerHook__login(self, client, conn)
+        except Exception as e:
+            self.log.warning(
+                "Registry login failed (%s). Proceeding without authentication.", e
+            )
+
+
+class DockerOperatorWithFallback(DockerOperator):
+    """DockerOperator that attempts a force pull but falls back to a local image on failure."""
```
mddilley: Might be worth noting in this comment that this supports all parameters of the DockerOperator except for `force_pull` although that might be apparent enough through the code in the custom class 🤷

dags/test_docker_operator_wrapper.py
```diff
@@ -0,0 +1,35 @@
...
+        "owner": "airflow",
+        "retries": 0,
+        "execution_timeout": pendulum.duration(minutes=5),
+    },
+    doc_md="Hello world DAG using DockerOperatorWithFallback with the python3 image",
+)
+def test_docker_operator_wrapper():
+    DockerOperatorWithFallback(
+        task_id="hello_world",
+        image="python:3",
```
mddilley, line 26: thanks to the 🤖s, i was able to simulate a failed registry pull by running the following on the command line:
```bash
docker pull python:3
docker tag python:3 mikes.cool.registry/python-is-fun:3
```
after that i triggered the DAG in the UI, and i saw the logging work just as we would expect during an outage:
Image

mddilley (commented):
Thanks for digging into DockerOperator - it was fun to read the internals and learn more about the built-in behavior. The tests all worked as expected, and I left some comments about adding a few things to help us make sure we don't have a regression with a new version. Let me know what you think!

dags/test_docker_operator_wrapper.py
frankhereford: I would totally move my images to `mikes.cool.registry`. Thanks for this TIL!

dags/utils/docker_operator.py
frankhereford: This is a really good idea. When you're messing with the parts your not supposed to, you're deep in the danger zone.

I've added a new class method `_require_base_login_method` which gets at the login method more generically and also which verifies that it's there and callable. We'll get a runtime error if the required bits and pieces are not where they need to be. 

https://github.com/cityofaustin/atd-airflow/pull/341/commits/92a26a52949a0e617cfbbb73fafa6bcf4f1a8c71

dags/utils/docker_operator.py
frankhereford: https://github.com/cityofaustin/atd-airflow/pull/341/commits/50a8b99ba59f0993c5bf3487f1142a9cd4f750c5

Great call - thanks!

dags/utils/docker_operator.py
frankhereford: Great idea, all breadcrumbs are good breadcrumbs. https://github.com/cityofaustin/atd-airflow/pull/341/commits/b68c3428f47cd128a7ec2665a6298d171dd5bf4b

chiaberry (commented):
Kicked off the test dag ✅ and it works

Why are we ignoring force_pull, does this mean we lose the option to opt in to pulling docker images and will pull a new one every time no matter what?

chiaberry (approved):
I asked my question posted above during dev sync and Frank replied that if we want to not force pull an image, we can just use the normal DockerOperator

dags/utils/docker_operator.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Original Method: https://github.com/apache/airflow/blob/dc939331f4cd90892fd201931c466dffff977a4f/providers/docker/src/airflow/providers/docker/operators/docker.py#L487
+        """
+
+        self.log.info("Attempting to pull image %s", self.image)
+        try:
+            for output in self.cli.pull(self.image, stream=True, decode=True):
+                if isinstance(output, dict) and "status" in output:
+                    self.log.info("%s", output.get("status", ""))
+        except Exception as e:
```
johnclary: What do you think about narrowing this a bit? I'm not sure we want to suppress anything besides a 500 error from Docker hub. A 4xx error, i think, would definitely be something we want to know about.

dags/utils/docker_operator.py
```diff
@@ -0,0 +1,79 @@
...
+                f"`DockerHook.{cls._BASE_LOGIN_METHOD}` is missing or not callable. "
+                "Update _DockerHookWithLoginFallback to match the provider implementation."
+            )
+        return method
+
+    def _DockerHook__login(self, client, conn):
+        base_login = self._require_base_login_method()
+        try:
+            base_login(self, client, conn)
+        except Exception as e:
```
johnclary, line 29: Ditto here, can we narrow this down to 5xx errors? This way we'll know about a misconfigured docker credential.

dags/utils/docker_operator.py
```diff
@@ -0,0 +1,79 @@
...
+            for output in self.cli.pull(self.image, stream=True, decode=True):
+                if isinstance(output, dict) and "status" in output:
+                    self.log.info("%s", output.get("status", ""))
+        except Exception as e:
+            self.log.warning(
+                "Could not pull image %s (%s). Using local image if available.",
+                self.image,
+                e,
+            )
+        return self._run_image()
```
johnclary: I think the stakes are pretty low here—if this operator breaks we can revert to the built-in one—but I'm curious what you think about Copilot's [concern](https://github.com/cityofaustin/atd-airflow/pull/341/changes#r3118932234) about depending on private methods. 

A bot I chatted with gave a similar opinion, and suggested we consider using proper subclassing with `super()` and a custom `pre_execute` [method](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/operators.html). 

I don't want to pick nits with your implementation, but, if it's all bot code anyway, i'm wondering if it's worth trying to avoid depending on the operator internals.

johnclary (commented):
Very cool—I was able to test this and it worked totally smoothly. Thanks for writing up a new test DAG for this ✨

dags/utils/docker_operator.py
johnclary: I guess there are other HTTP errors (connection timeout, others?) that we need to consider during a Docker outage. My primary concern here is this failing silently when it shouldn't be.

dags/utils/docker_operator.py
frankhereford: Good call, this should only let 5xx errors though now. Anything else or some other non-response type failure will get an exception raised. https://github.com/cityofaustin/atd-airflow/pull/341/commits/93fbc6236e9799dfd32f7fb81d8ed7dd8613780b

dags/utils/docker_operator.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+
+        self.log.info("Attempting to pull image %s", self.image)
+        try:
+            for output in self.cli.pull(self.image, stream=True, decode=True):
+                if isinstance(output, dict) and "status" in output:
+                    self.log.info("%s", output.get("status", ""))
+        except Exception as e:
+            if not self._should_fallback_to_local_image(e):
+                raise
```
frankhereford: In this private method, we do some more detailed analysis of our situation, explained above.

dags/utils/docker_operator.py
```diff
@@ -0,0 +1,125 @@
...
+            return status_code
+
+        response = getattr(exc, "response", None)
+        return getattr(response, "status_code", None)
+
+    @classmethod
+    def _should_fallback_to_local_image(cls, exc: Exception) -> bool:
+        status_code = cls._extract_status_code(exc)
+        if isinstance(status_code, int):
+            return 500 <= status_code < 600
```
frankhereford, line 77: extract our HTTP code and return it. The number is a true value, and when we return true from this function, airflow will not raise an exception and fall back to our existing copy of the image on disk. We continue, even though our pull failed. ✅

dags/utils/docker_operator.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if isinstance(exc, (TimeoutError, ConnectionError)):
+            return True
```
frankhereford, line 80: Check for two specific exceptions that we are OK with.

dags/utils/docker_operator.py
frankhereford: I'm going to give it a shot. I like the take that if we're using tools that should enable high quality solutions, we should hold our solutions to a higher standard.

frankhereford: I would prefer to avoid internals and generally am 100% on board with it being brittle and suboptimal, but I think we want to leave this largely alone, with some added runtime checks. Here's my reasoning:

* We use internals in two places:
  1) Name-mangling `__login` and defining `_DockerHook__login` and we guard that what we expect to be where we expect it with a `RuntimeError` around line 18. It'll bark if the provider internals change, and that feels 'enough' assurance to bend the rules here for me.
  2) Using the single-underscore `_run_image()` instead of trying to fit into the exposed pattern of `pre_execute()`, `execute()`, and `post_execute()`. Generally, I would love to fix into this pattern, but the way the `DockerOperator` works is that it does both the pull and execute (`self._run_image()`) portion of the process inside it's `execute()`. It views the pulling as part of the main execution portion of the life-cycle. If we super().execute() after our own custom pull logic, we'll end up executing the original pull logic that we're trying to avoid. Pulling twice feels like it's a path to hard to debug issues and seems less ideal than using the internal to pick up on the execution path the original DockerOperator takes.

For the _run_image() usage, I think if we add [an explicit guard](https://github.com/cityofaustin/atd-airflow/pull/341/commits/d1cdf1cb62db45993898301cc509dcebd9348158) around it that can detect if it's not there, I am hopeful that this is enough safety margin to make it feel safe enough using the internal. 

I hope this is helpful information. I feel like I'm out on the limits of my python experience here, so please forgive me if I'm misunderstanding. I'm always game to sync on any of this and continue to refine.

frankhereford:
@mddilley Thanks for the nudge today to look at this and thank you @johnclary for your thoughtful questions. I've done my best to answer them, and I think this is ready for y'all to look at again if that's something you'd like to do. I appreciate all y'all's help, as always.

mddilley (approved):
Tested this out with the updates, and everything works just like before the latest updates. Don't want to jinx us but can't wait for the next DockerHub outage so we can keep service desk and other frequent jobs going. 😎

dags/utils/docker_operator.py
johnclary: Thanks for walking me through your reasoning here that we are constrained DockerOperator's design choices. 

I'm hoping to get to testing this later today or tomorrow. Any suggestions for making sure we test all of the code paths?

johnclary: Well, I was able to test 403 and 404 errors successfully while devving on https://github.com/cityofaustin/atd-airflow/pull/345. 

Not sure how to test `TimeoutError` and `ConnectionError` though 🤔

johnclary: @frankhereford does it make sense to update the test DAG to try to improve code coverage? or the PR test instructions? There's been a decent amount of churn since the first go at this.

dags/utils/docker_operator.py
```diff
@@ -0,0 +1,146 @@
...
+    def execute(self, context):
+        """
+        Original Method: https://github.com/apache/airflow/blob/dc939331f4cd90892fd201931c466dffff977a4f/providers/docker/src/airflow/providers/docker/operators/docker.py#L487
+        """
+
+        self._require_run_image()
+
+        # Base DockerOperator would also pull inside _run_image when force_pull=True.
+        # We handle pull behavior here to provide fallback semantics and avoid a second pull.
+        should_pull = bool(self.force_pull)
```
mddilley: @frankhereford since this code is overriding the `execute` method now, I noticed that Airflow's method checks if the image is present as a backup to the `force_pull` param here like this: 
```python
if self.force_pull or not self.cli.images(name=self.image):
```

do you think that we should do the same for the case when we've first deployed a DAG with `force_pull=false`? I don't _think_ that we would ever do that or want to pull and then run a stale image from then on, but I think this code would jump to the `else` block and then never pull the image.

dags/utils/docker_operator.py
```diff
@@ -0,0 +1,146 @@
...
+                "Skipping image pull for %s because force_pull is disabled.",
+                self.image,
+            )
+
+        original_force_pull = self.force_pull
+        try:
+            self.force_pull = False
+            return self._run_image()
+        finally:
+            self.force_pull = original_force_pull
```
mddilley, line 148: @frankhereford could you add a comment about what these blocks are doing? I was looking at Airflow's `_run_image` method, and I don't see it reading the `force_pull` value. I'm probably missing something!

I did test my fake registry method and noticed it doesn't fallback to the disk image without these lines so it is definitely doing something!

dags/utils/docker_operator.py
mddilley: I am also curious about this since the current DAG tests the ideal state of the registry being reachable. Some other scenarios that I could think of are:
- fake registry route with image already on disk like i did with:
```bash
docker pull python:3
docker tag python:3 mikes.cool.registry/python-is-fun:3
```
- fake registry with no image already on disk
- the other errors that John mentioned that Claude is suggesting that we could possibly monkey patch? 🐒 

What do y'all think about beefing up the test DAG a little more, shipping, and then refining over time like we did with the Slack error notifications?

mddilley (commented):
Very cool stuff 😎 , and I was able to test this with or without a cached local image and see it work. I asked a couple questions, and I also chained off one of John's earlier comments https://github.com/cityofaustin/atd-airflow/pull/341#discussion_r3235993996.

frankhereford:
Thanks y'all for your help and reviews. This pattern didn't work out due to complications around authentication. I am closing this and hope to solve this issue with https://github.com/cityofaustin/atd-airflow/pull/368 instead.

PR #342, opened 2026-04-21
Refactor DAGs to use DockerOperatorWithFallback

frankhereford:
## Note

Reviewing this PR is 😵‍💫 -- i've tried to make it as easy as possible, but it's a thing.

This PR merges into the one which contains just the new Docker Operator with custom behavior, found in #341. This is all the DAG refacatoring to use the new operator.

This updated multiple DAG files to replace the deprecated DockerOperator with DockerOperatorWithFallback. This change enhances the robustness of Docker operations by allowing fallback to local images if pulling fails. Additionally, removed the force_pull parameter from DockerOperator calls to streamline the configuration.

## Associated issues

## Associated repo

airflow itself

## Testing

* Read the code, and ask yourself if you like this pattern
* Fire up a few DAGs, the more the merrier. You're testing when docker hub is up.
* Simulating a docker hub failure is tricky, I don't have a good suggestion on how. 
  * I tried giving the docker hub important hostnames nonroutable IPs in my `/etc/hosts` but the containers have their own DNS systems and don't depend on the OS running on my local hardware. I'm open to any ideas!
---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/atd_cost_of_service_fees.py
+++ b/dags/atd_cost_of_service_fees.py
@@ -1,6 +1,6 @@
 from os import getenv
 
-from airflow.providers.docker.operators.docker import DockerOperator
+from utils.docker_operator import DockerOperatorWithFallback
 from airflow.sdk import dag
 from pendulum import datetime, duration
 
@@ -81,7 +81,6 @@ def atd_cost_of_service_fees():
         environment=env_vars,
         tty=True,
         docker_conn_id="docker_default",
-        force_pull=True,
         mount_tmp_dir=False,
     )
 
--- a/dags/atd_cost_of_service_fees.py
+++ b/dags/atd_cost_of_service_fees.py
@@ -72,7 +72,7 @@ def atd_cost_of_service_fees():
 
     env_vars = get_env_vars_task(REQUIRED_SECRETS)
 
-    load_fees_to_knack = DockerOperator(
+    load_fees_to_knack = DockerOperatorWithFallback(
         task_id="atd_cost_of_service_fees_to_knack",
         image="atddocker/atd-cost-of-service:production",
         auto_remove="force",
--- a/dags/atd_cost_of_service_fees.py
+++ b/dags/atd_cost_of_service_fees.py
@@ -73,6 +73,7 @@ def atd_cost_of_service_fees():
     env_vars = get_env_vars_task(REQUIRED_SECRETS)
 
     load_fees_to_knack = DockerOperatorWithFallback(
+        force_pull=True,
         task_id="atd_cost_of_service_fees_to_knack",
         image="atddocker/atd-cost-of-service:production",
         auto_remove="force",
--- a/dags/atd_finance_data_fdu_program_tagging.py
+++ b/dags/atd_finance_data_fdu_program_tagging.py
@@ -116,7 +116,7 @@ with DAG(
         mount_tmp_dir=False,
     )
 
-    t2 = DockerOperatorWithFallback(
+    t2 = DockerOperator(
         task_id="tagging_fdus",
         image="atddocker/atd-finance-data:production",
         docker_conn_id="docker_default",
--- a/dags/atd_finance_data_fdu_program_tagging.py
+++ b/dags/atd_finance_data_fdu_program_tagging.py
@@ -1,7 +1,7 @@
 from os import getenv
 
 from airflow.sdk import task, DAG
-from airflow.providers.docker.operators.docker import DockerOperator
+from utils.docker_operator import DockerOperatorWithFallback
 from pendulum import datetime, duration
 
 from utils.onepassword import get_env_vars_task
@@ -113,7 +113,6 @@ with DAG(
         command="python3 upload_to_s3.py fdu_expenses_obligated",
         environment=env_vars,
         tty=True,
-        force_pull=True,
         mount_tmp_dir=False,
     )
 
@@ -125,7 +124,6 @@ with DAG(
         command="python3 fdu_program_tagging.py",
         environment=env_vars,
         tty=True,
-        force_pull=False,
         mount_tmp_dir=False,
     )
 
--- a/dags/atd_finance_data_fdu_program_tagging.py
+++ b/dags/atd_finance_data_fdu_program_tagging.py
@@ -1,7 +1,6 @@
 from os import getenv
 
 from airflow.sdk import task, DAG
-from airflow.providers.docker.operators.docker import DockerOperator
 from utils.docker_operator import DockerOperatorWithFallback
 from pendulum import datetime, duration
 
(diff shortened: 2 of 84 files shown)


dags/atd_signal_comms.py
```diff
@@ -110,132 +110,122 @@ def atd_signal_comms():
...
         image=docker_image,
         docker_conn_id="docker_default",
         auto_remove="force",
         command=f"python atd-signal-comms/run_comm_check.py camera --env {deployment_stage_abbreviation}",
         environment=env_vars,
         tty=True,
-        force_pull=True,
         mount_tmp_dir=False,
         network_mode="bridge",
     )
```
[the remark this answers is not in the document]
frankhereford: Fixed

dags/atd_signal_comms.py
```diff
@@ -110,132 +110,122 @@ def atd_signal_comms():
...
         image=docker_image,
         docker_conn_id="docker_default",
         auto_remove="force",
         command=f"python atd-signal-comms/run_comm_check.py detector --env {deployment_stage_abbreviation}",
         environment=env_vars,
         tty=True,
-        force_pull=False,
         mount_tmp_dir=False,
         network_mode="bridge",
     )
```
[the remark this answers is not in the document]
frankhereford: Super good call. 

* Added from `airflow.providers.docker.operators.docker import DockerOperator` import.
* `cameras_s3` stays as `DockerOperatorWithFallback` — it pulls the image at the start of the chain (with graceful fallback to local cache if the registry is unreachable).
* The remaining 9 tasks (`detectors_s3`, `dms_s3`, `battery_backup_s3`, `signal_monitors_s3`, and all 5 Socrata tasks) are now plain `DockerOperator` with `force_pull=False`, so they reuse the image already on disk.

dags/atd_knack_purchase_request_copier.py
```diff
@@ -62,15 +62,14 @@
...
         task_id="purchase_request_copier",
         image=docker_image,
         docker_conn_id="docker_default",
         auto_remove="force",
         command=f"./atd-knack-services/services/purchase_request_copier.py -a {app_name} -c {container}",
         environment=env_vars,
         tty=True,
-        force_pull=False,  # atd_knack_signals pulls this image every 5 minutes
         mount_tmp_dir=False,
     )
```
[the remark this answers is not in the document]
frankhereford: I did end up reverting this back to the way it was. Custom situation, existing custom solution - nothing to fix there.

dags/atd_cost_of_service_fees.py
frankhereford, line 1: I'm going to call this a "normal" pattern that you see in this PR. I'll call out all DAGs which are not normal and speak to what makes them special or unique.

dags/atd_knack_development_services.py
frankhereford, line 1: OK - here's one that's not "normal." We enhance the template to be able to specify what operator we want to use and then enumerate over the tasks using the index to know when to use the new one vs all the rest which are the regular docker operator.

dags/atd_knack_purchase_request_copier.py
frankhereford, line 1: This is a non-edit, but this is a special, minutely-running DAG. It just doesn't pull, ever. It relies on another DAG somewhere else to pull this on a more slow-paced interval.

dags/atd_moped_ecapris_funding_sync.py
frankhereford, line 1: Normal operation here, this is just a little fancy with the splatted out common config. I don't need the blame button to know who wrote this elegant work✨

dags/atd_signal_comms.py
```diff
@@ -131,9 +131,9 @@ def atd_signal_comms():
         command=f"python atd-signal-comms/run_comm_check.py detector --env {deployment_stage_abbreviation}",
         environment=env_vars,
         tty=True,
-        force_pull=False,
         mount_tmp_dir=False,
         network_mode="bridge",
+        force_pull=False,
```
frankhereford, line 135: I waffled around on how I wanted to do this and ended up moving this parameter around in this file, but it's just relocated, with only the first task getting the new enhanced operator. We don't want to pull over and over and over ..

dags/dts_row_reporting.py
frankhereford, line 1: Similar to a previous pattern above.

Charlie-Henry (approved):
ty for these @frankhereford. I can't think of any of these that we would not want this? Like are there any DAGs we can think of where if we are not running the latest version undesirable things could happen?

dags/atd_finance_data_fdu_program_tagging.py
```diff
@@ -105,27 +105,25 @@
...
         auto_remove="force",
         command="python3 upload_to_s3.py fdu_expenses_obligated",
         environment=env_vars,
         tty=True,
-        force_pull=True,
         mount_tmp_dir=False,
     )
 
-    t2 = DockerOperator(
+    t2 = DockerOperatorWithFallback(
```
chiaberry: This should be switched back to the vanilla DockerOperator right?

mddilley: That sounds right to me so that the first task tries to reach DockerHub, and, if that fails, it will uses the image on disk and then so will the rest of the tasks without trying to pull.

dags/atd_finance_data_fdus.py
```diff
@@ -142,27 +142,25 @@
...
         auto_remove="force",
         command="python3 upload_to_s3.py fdus",
         environment=data_tracker_env,
         tty=True,
-        force_pull=True,
         mount_tmp_dir=False,
     )
 
-    t2 = DockerOperator(
+    t2 = DockerOperatorWithFallback(
```
mddilley: Same here, I think that this should switch back to vanilla DockerOperator since we are not pulling on the second task.

mddilley (changes requested):
Looking good! Thanks for sweeping through all these DAGs. I'm requesting changes because I want to check the same thing that Chia flagged. For DAGs that have multiple tasks that use the same image we should use the `DockerOperatorWithFallback` on the first task and the plain `DockerOperator` for the rest so we don't pull the same image over and over.

There is also the question about whether we should explicitly set `force_pull=false` on the plain `DockerOperator`s since I saw we have a mix since it defaults to `false` if not set. I think the docs are easy enough to find and the special operator is straightforward to trace back to but i'm curious to hear other thoughts.

dags/atd_finance_data_fdus.py
frankhereford: Good call thank you! https://github.com/cityofaustin/atd-airflow/pull/342/commits/dbfec5d60733b405b82d16d4697fd8038df57fac

dags/atd_finance_data_fdu_program_tagging.py
frankhereford: Thanks for the review and for catching this 🙏  https://github.com/cityofaustin/atd-airflow/pull/342/commits/dbfec5d60733b405b82d16d4697fd8038df57fac

frankhereford:
We've changed our approach twice since this PR was opened, and it makes the review process just that much harder to sort out. I'm going to close this PR, and open a new one.

PR #347, opened 2026-05-18
Fix Health Checks in Airflow Compose Stack

frankhereford:
## Associated issues

This PR aims to close https://github.com/cityofaustin/atd-data-tech/issues/28476.

## Associated repo

Airflow itself

## Testing

* Spin up your local stack
* `docker compose ps` 
* Is everything healthy, similar to below?

```
STATUS
Up 58 minutes (healthy)
Up 47 minutes (healthy)
Up About an hour (healthy)
Up 24 minutes (healthy)
Up About a minute (healthy)
Up About an hour (healthy)
```

* If unhealthy, please share the output of the following:
```
docker inspect --format='{{json .State.Health}}' <container name from `docker ps`>
```

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/docker-compose-production.yaml
+++ b/docker-compose-production.yaml
@@ -9,3 +9,13 @@ services:
       - ${AIRFLOW_PROJ_DIR}/haproxy/haproxy-production.cfg:/usr/local/etc/haproxy/haproxy.cfg
     depends_on:
       - airflow-apiserver
+    healthcheck:
+      test: # health check to see if the configuration is valid and if the webserver is returning a 301 as expected on port 8080. 
+        [
+          "CMD-SHELL",
+          "haproxy -c -q -f /usr/local/etc/haproxy/haproxy.cfg && bash -ec 'exec 3<>/dev/tcp/127.0.0.1/8080; printf \"GET /api/v2/version HTTP/1.1\\r\\nHost: localhost\\r\\nConnection: close\\r\\n\\r\\n\" >&3; read -r _ code _ <&3; [[ \"$$code\" == \"301\" ]]'",
+        ]
+      interval: 30s
+      timeout: 10s
+      retries: 5
+      start_period: 30s
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -71,7 +71,10 @@ services:
     image: haproxy:lts
     restart: unless-stopped
     healthcheck:
-      test: # health check to see if the configuration is valid and if the api is reachable. No curl or wget, so a raw TCP connection is used to check the HTTP response code.
+      # This health check is to see if the configuration is valid and if the api is reachable. 
+      # No curl or wget is available in the haproxy image, so a raw TCP connection is used to 
+      # check the HTTP response code.
+      test: 
         [
           "CMD-SHELL",
           "haproxy -c -q -f /usr/local/etc/haproxy/haproxy.cfg && bash -ec 'exec 3<>/dev/tcp/127.0.0.1/80; printf \"GET /api/v2/version HTTP/1.1\\r\\nHost: localhost\\r\\nConnection: close\\r\\n\\r\\n\" >&3; read -r _ code _ <&3; [[ \"$$code\" == \"200\" ]]'",
@@ -122,6 +125,7 @@ services:
     <<: *airflow-common
     command: dag-processor
     healthcheck:
+      # We run our stack as root, so we need to use gosu to run the airflow jobs check command as the airflow user.
       test:
         [
           "CMD-SHELL",
@@ -141,6 +145,7 @@ services:
     <<: *airflow-common
     command: triggerer
     healthcheck:
+      # We run our stack as root, so we need to use gosu to run the airflow jobs check command as the airflow user.
       test:
         [
           "CMD-SHELL",
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -74,7 +74,7 @@ services:
       test: # healthy to see if the configuration is valid and if the api is reachable. No curl or wget, so a raw TCP connection is used to check the HTTP response code.
         [
           "CMD-SHELL",
-          "haproxy -c -q -f /usr/local/etc/haproxy/haproxy.cfg && bash -ec 'exec 3<>/dev/tcp/127.0.0.1/8080; printf \"GET /api/v2/version HTTP/1.1\\r\\nHost: localhost\\r\\nConnection: close\\r\\n\\r\\n\" >&3; read -r _ code _ <&3; [[ \"$$code\" == \"200\" ]]'",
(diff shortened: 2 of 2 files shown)


docker-compose.yaml
```diff
@@ -70,6 +70,16 @@ services:
...
+    healthcheck:
+      test: # healthy to see if the configuration is valid and if the api is reachable. No curl or wget, so a raw TCP connection is used to check the HTTP response code.
+        [
+          "CMD-SHELL",
+          "haproxy -c -q -f /usr/local/etc/haproxy/haproxy.cfg && bash -ec 'exec 3<>/dev/tcp/127.0.0.1/80; printf \"GET /api/v2/version HTTP/1.1\\r\\nHost: localhost\\r\\nConnection: close\\r\\n\\r\\n\" >&3; read -r _ code _ <&3; [[ \"$$code\" == \"200\" ]]'",
+        ]
+      interval: 30s
+      timeout: 10s
+      retries: 5
+      start_period: 10s
```
frankhereford: Adding a health check here for HAProxy. We're constrained to this very raw check of the HTTP response code because we don't have curl or wget on the vanilla haproxy image and we don't really want to build our own image just so we can health check more succinctly.

docker-compose.yaml
```diff
@@ -115,7 +125,7 @@ services:
       test:
         [
           "CMD-SHELL",
-          'airflow jobs check --job-type DagProcessorJob --hostname "$${HOSTNAME}"',
+          'gosu airflow airflow jobs check --job-type DagProcessorJob --hostname "$${HOSTNAME}"',
```
frankhereford, line 132: `gosu` is a docker oriented `su` command written in go for this specific purpose. We're using it to say "run this command as `airflow` which is needed because we run our stack as `root`. 

A bit of background on the utility:

>In the world of containers and system administration, `gosu` is a lightweight utility written in Go that allows a user to execute a command as a specific user (similar to `sudo` or `su`).

> It is primarily used in Docker container entrypoint scripts to drop root privileges safely.

docker-compose.yaml
```diff
@@ -70,6 +70,16 @@ services:
   haproxy:
     image: haproxy:lts
     restart: unless-stopped
+    healthcheck:
+      test: # healthy to see if the configuration is valid and if the api is reachable. No curl or wget, so a raw TCP connection is used to check the HTTP response code.
```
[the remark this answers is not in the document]
frankhereford: Typo - thanks!

docker-compose.yaml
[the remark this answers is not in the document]
frankhereford: The stacks are low (zero) on this health check, so i'm comfortable relying on bash sticking around in the LTS image. The check is pulling the health check status page that the apiserver uses, and it's not an interactive page, it's just the json health check endpoint -- meaning we can count on the very specific behavior we're testing for here, in this case, a 200 HTTP response code.

docker-compose.yaml
```diff
@@ -134,7 +144,7 @@ services:
       test:
         [
           "CMD-SHELL",
-          'airflow jobs check --job-type TriggererJob --hostname "$${HOSTNAME}"',
+          'gosu airflow airflow jobs check --job-type TriggererJob --hostname "$${HOSTNAME}"',
```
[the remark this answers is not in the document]
frankhereford: For various reasons, we run our stack as root, and the airflow python libraries are not available to root by default. Running the script as the airflow user does make those available, and the inner airflow command can execute. This is a hallucinated problem here, imho.

docker-compose.yaml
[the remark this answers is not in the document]
frankhereford: I bumped this up to 30 seconds, but I also find that it goes healthy nearly instantly, so I think I'm chasing a problem we don't have here.

docker-compose-production.yaml
```diff
@@ -9,3 +9,13 @@ services:
       - ${AIRFLOW_PROJ_DIR}/haproxy/haproxy-production.cfg:/usr/local/etc/haproxy/haproxy.cfg
     depends_on:
       - airflow-apiserver
+    healthcheck:
+      test: # health check to see if the configuration is valid and if the webserver is returning a 301 as expected on port 8080. 
+        [
+          "CMD-SHELL",
+          "haproxy -c -q -f /usr/local/etc/haproxy/haproxy.cfg && bash -ec 'exec 3<>/dev/tcp/127.0.0.1/8080; printf \"GET /api/v2/version HTTP/1.1\\r\\nHost: localhost\\r\\nConnection: close\\r\\n\\r\\n\" >&3; read -r _ code _ <&3; [[ \"$$code\" == \"301\" ]]'",
```
frankhereford, line 16: This is the production version of the haproxy health check -- where it checks port 8080 for a 301, as specified in the haproxy config.

docker-compose.yaml
```diff
@@ -70,6 +70,16 @@ services:
   haproxy:
     image: haproxy:lts
     restart: unless-stopped
+    healthcheck:
+      test: # health check to see if the configuration is valid and if the api is reachable. No curl or wget, so a raw TCP connection is used to check the HTTP response code.
+        [
+          "CMD-SHELL",
+          "haproxy -c -q -f /usr/local/etc/haproxy/haproxy.cfg && bash -ec 'exec 3<>/dev/tcp/127.0.0.1/80; printf \"GET /api/v2/version HTTP/1.1\\r\\nHost: localhost\\r\\nConnection: close\\r\\n\\r\\n\" >&3; read -r _ code _ <&3; [[ \"$$code\" == \"200\" ]]'",
```
frankhereford, line 80: This is the local development version. Here, we're looking for the API Server's health check coming through via the haproxy service on plain HTTP.

chiaberry (approved):
`docker ps` returns healthy containers now. Do you think it would be worth adding a note about the usage of `gosu` in the readme? or do you think it needs to documented at all, is it obvious why you used it?

docker-compose.yaml
mddilley, line 1: thanks for the reminder that we run our Airflow stack as root. I can't remember if we've captured why we do that in docs anywhere. It is making me think that it would be worth capturing since we are having to work around it for the health checks.

mddilley (approved):
I spun this up earlier and just checked back in on the health checks to finish this review. Everything is still up and running and showing as healthy! I added one comment about stashing some context about these changes somewhere (maybe [GitBook](https://app.gitbook.com/o/-LzDQOVGhTudbKRDGpUA/s/-M4Ul-hSBiM-3KkOynqS/servers/dts-int-data-p01-on-prem-server/airflow)?). 🚢 🚀

frankhereford:
@chiaberry wrote:

> `docker ps` returns healthy containers now. Do you think it would be worth adding a note about the usage of `gosu` in the readme? or do you think it needs to documented at all, is it obvious why you used it?

Hi Chia, thanks for asking. I think documentation is always helpful, I'll add a little blurb as an inline comment. Thanks for asking.

docker-compose.yaml
frankhereford: I think this is a good idea, I like the idea of putting a more detailed explanation of why we do this in gitbook instead of the public repo. I'll do that.

PR #349, opened 2026-05-20
Wait for date filter in knack development services

chiaberry:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/28613

## Testing
compare whats running in airflow3

Screenshot 2026-05-18 at 4 21 46 PM

to

Screenshot 2026-05-20 at 9 14 32 AM

**Steps to test:**

Spin up local airflow and check that the dags import correctly. 
I tested by commenting out all of the tasks in commands except "tia_mitigations_to_postgrest" and "tia_mitigations_to_socrata", just to minimize then number of tasks I was testing. 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/atd_knack_development_services.py
+++ b/dags/atd_knack_development_services.py
@@ -187,7 +187,7 @@ def atd_knack_development_services():
             )
         )
 
-    chain(*tasks)
+    chain(date_filter_arg, *tasks)
 
 
 atd_knack_development_services()


dags/atd_knack_development_services.py
```diff
@@ -187,7 +187,7 @@ def atd_knack_development_services():
             )
         )
 
-    chain(*tasks)
+    chain(date_filter_arg, *tasks)
```
chiaberry, line 190: considered putting `env_vars` also in the chain, but the dependency graph seems fine and we haven't encountered the same race condition with environment variables

Charlie-Henry (approved):
Looks good ty for fixing this bug @chiaberry 💯

johnclary (approved):
Looks good—thanks!

PR #353, opened 2026-06-01
Update slack notifier to check if summary is defined on the `dag` object when checking the timetable

frankhereford:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/28749

This intends to fix the broken notifications in slack we have been encountering since I updated the stack to 3.2.1 on Friday.

## Associated repo

Airflow itself

## Testing

Observe the organic notification in production [here](https://austininnovation.slack.com/archives/C013G4RNJ01/p1780338106589719).  This happened today after the fix was put in place via an _ad hoc_ change that will be replaced by the same change found here in this PR. If you agree that notifications are working again: ✅ 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/utils/slack_operator.py
+++ b/dags/utils/slack_operator.py
@@ -138,8 +138,8 @@ def task_fail_slack_alert(context):
 
     schedule_description = (
         format_schedule(dag.timetable.summary)
-        if dag and hasattr(dag, "timetable")
-        else "Not available"
+        if dag and hasattr(getattr(dag, "timetable", None), "summary")
+        else "Manual Run / None"
     )
 
     all_exceptions = extract_all_exceptions(context)


mddilley (approved):
Looks great, and I think the replacement for "Not available" is a nice improvement! Thanks for getting catching this regression and fixing. I saw the error notifications start working again earlier. 🚢

chiaberry (approved):
✅

johnclary:
interesting—so there was a breaking change to an airflow internal that we relied on for this. @frank i noticed we have a section in the doc called [Update process](https://github.com/cityofaustin/atd-airflow#update-process)—what do you think about listing out our test DAGs that can be run to verify an upgrade, since `test_slack_notifier` notifier would have caught this?

mddilley:
@johnclary I like that idea a lot! Another place that I was thinking that we could capture is being a little more specific about the Airflow upgrade testing in the [monthly maintenance issues](https://github.com/cityofaustin/atd-data-tech/issues/28611) cc @frankhereford

frankhereford:
@johnclary & @mddilley Thanks for this discussion 🙏 I think we're all aligned, and I've added something to refinement today WRT airflow updates. I think y'all's suggestion of here of some testing DAGs, at the very least, the notification system test DAGs are a good idea to include, and I also want to back-off a little from the bleeding edge, and run the latest patch of the second to latest minor release. 

All in all, a little less frequent on the updating and more testing when we do update -- seems like low hanging fruit to help minimize disruptions like this one.

PR #358, opened 2026-06-09
Add a dag to check citybase health

chiaberry:
## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/26146

## Testing

**Steps to test:**

1. start airflow locally
2. From the lefthand menu choose Admin, Connections. Use the button to open the add connection modal. Add one that looks like
Image
3. Trigger the "citybase_postback_healthcheck" dag. You must be on the city network to access the healthcheck. 

I have added the https connection to prod

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/citybase_healthcheck.py
+++ b/dags/citybase_healthcheck.py
@@ -3,7 +3,7 @@ from airflow import DAG
 from airflow.providers.http.sensors.http import HttpSensor
 from pendulum import datetime, duration
 
-from utils.slack_operator import task_fail_slack_alert
+from utils.slack_operator import task_fail_slack_alert, slack_member_ids
 
 DEPLOYMENT_ENVIRONMENT = getenv("ENVIRONMENT", "development")
 
@@ -13,7 +13,7 @@ DEFAULT_ARGS = {
     "start_date": datetime(2026, 1, 1, tz="America/Chicago"),
     "retries": 0,
     "execution_timeout": duration(seconds=30),
-    # "on_failure_callback": task_fail_slack_alert,
+    "on_failure_callback": task_fail_slack_alert,
 }
 
 doc_md = ""
--- a/dags/citybase_healthcheck.py
+++ b/dags/citybase_healthcheck.py
@@ -16,7 +16,14 @@ DEFAULT_ARGS = {
     "on_failure_callback": task_fail_slack_alert,
 }
 
-doc_md = ""
+doc_md = """
+Checks that the citybase postback service is running and reachable.
+The application runs on the bastion.
+
+See the atd-citybase readme for more details on deployment and how to restart.
+
+If the postback is down for an extended period of time, inform Hanna so she can check Revenue Managment for missed transactions.
+"""
 
 with DAG(
     dag_id="citybase_postback_healthcheck",
@@ -33,7 +40,7 @@ with DAG(
 
     check_citybase_endpoint = HttpSensor(
         task_id="check_citybase_postback_endpoint",
-        http_conn_id="check_citybase",
+        http_conn_id="http_default",
         endpoint="citybase.austinmobility.io",
         request_params={},
         response_check=lambda response: response.status_code == 200,
--- a/dags/citybase_healthcheck.py
+++ b/dags/citybase_healthcheck.py
@@ -30,7 +30,7 @@ with DAG(
     description="Checks citybase postback for 200 response",
     doc_md=doc_md,
     default_args=DEFAULT_ARGS,
-    schedule="0 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    schedule="*/10 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
     tags=["repo:atd-citybase", "citybase"],
     catchup=False,
 ) as dag:
--- a/dags/citybase_healthcheck.py
+++ b/dags/citybase_healthcheck.py
@@ -1,0 +1,42 @@
+from os import getenv
+from airflow import DAG
+from airflow.providers.http.sensors.http import HttpSensor
+from pendulum import datetime, duration
+
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = getenv("ENVIRONMENT", "development")
+
+DEFAULT_ARGS = {
+    "owner": "airflow",
+    "depends_on_past": False,
+    "start_date": datetime(2026, 1, 1, tz="America/Chicago"),
+    "retries": 0,
+    "execution_timeout": duration(seconds=30),
+    # "on_failure_callback": task_fail_slack_alert,
+}
+
+doc_md = ""
+
+with DAG(
+    dag_id="citybase_postback_healthcheck",
+    description="Checks citybase postback for 200 response",
+    doc_md=doc_md,
+    default_args=DEFAULT_ARGS,
+    schedule="0 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    tags=["repo:atd-citybase", "citybase"],
+    catchup=False,
+) as dag:
+
+    dag.byline = f"Citybase postback healthcheck failed, {slack_member_ids['Chia']}"
+
+
+    check_citybase_endpoint = HttpSensor(
+        task_id="check_citybase_postback_endpoint",
+        http_conn_id="check_citybase",
+        endpoint="citybase.austinmobility.io",
+        request_params={},
+        response_check=lambda response: response.status_code == 200,
+    )
+
+    check_citybase_endpoint
--- a/dags/citybase_healthcheck.py
+++ b/dags/citybase_healthcheck.py
@@ -41,8 +41,8 @@ with DAG(
         task_id="check_citybase_postback_endpoint",
         http_conn_id="http_default",
         endpoint="citybase.austinmobility.io",
-        request_params={"verify": True, "timeout": 30},
-        response_check=lambda response: response.status_code == 200,
+        extra_options={"verify": True, "timeout": 30},
+        response_check=lambda response: response.json().get("status") == "OK",
     )
 
     check_citybase_endpoint
--- a/dags/citybase_healthcheck.py
+++ b/dags/citybase_healthcheck.py
@@ -1,6 +1,6 @@
 from os import getenv
 from airflow import DAG
-from airflow.providers.http.sensors.http import HttpSensor
+from airflow.providers.http.operators.http import HttpOperator
 from pendulum import datetime, duration
 
 from utils.slack_operator import task_fail_slack_alert, slack_member_ids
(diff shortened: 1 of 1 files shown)


dags/citybase_healthcheck.py
```diff
@@ -0,0 +1,49 @@
...
+    dag.byline = f"Citybase postback healthcheck failed, {slack_member_ids['Chia']}"
+
+
+    check_citybase_endpoint = HttpSensor(
+        task_id="check_citybase_postback_endpoint",
+        http_conn_id="http_default",
+        endpoint="citybase.austinmobility.io",
+        request_params={},
+        response_check=lambda response: response.status_code == 200,
+    )
```
frankhereford: I was concerned initially because I didn't see anything in this code or in the `http_default` connection that indicated the request should be made with SSL/TLS, but when I `curl`ed to investigate, the behavior we've configured for `HTTP` requests does indeed not return a 200. Kind of makes you wish they called it `HttpsSensor`.

```
$ curl -s -o /dev/null -w "%{http_code}\n" http://citybase.austinmobility.io
301
$ curl -s -o /dev/null -w "%{http_code}\n" https://citybase.austinmobility.io
200
```

dags/citybase_healthcheck.py
```diff
@@ -0,0 +1,49 @@
...
+
+    dag.byline = f"Citybase postback healthcheck failed, {slack_member_ids['Chia']}"
+
+
+    check_citybase_endpoint = HttpSensor(
+        task_id="check_citybase_postback_endpoint",
+        http_conn_id="http_default",
+        endpoint="citybase.austinmobility.io",
+        request_params={},
+        response_check=lambda response: response.status_code == 200,
```
frankhereford: I like this very airflow way of checking this, but I wonder if you might want to pull the file yourself in python and parse the JSON that is returned by the endpoint looking for that `"status":"OK"` KV pair. The idea being that you're testing further into the stack, down to the flask app and not the reverse proxy or web-server we have in the deployment. 

I'm torn on this, and I do like the airflow way too. I'm curious what others think. I'm 👍 on this shipping.

dags/citybase_healthcheck.py
```diff
@@ -0,0 +1,49 @@
...
+
+If the postback is down for an extended period of time, inform Hanna so she can check Revenue Managment for missed transactions.
+"""
+
+with DAG(
+    dag_id="citybase_postback_healthcheck",
+    description="Checks citybase postback for 200 response",
+    doc_md=doc_md,
+    default_args=DEFAULT_ARGS,
+    schedule="0 * * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
frankhereford: To your question about timing; I'd say hourly is very conservative; I'd have gone for something like every 5 minutes or every 10. The overhead of serving up that status page is practically zero; I say let's find out when it breaks just as soon as we can.

frankhereford (approved):
Thank you for this Chia, this works really well. It's neat to see a new pattern in our airflow dags too.

```
[2026-06-11 09:35:10] INFO - Poking: citybase.austinmobility.io
[2026-06-11 09:35:10] INFO - Success criteria met. Exiting.
```

🚢 🚢 🚢

dags/citybase_healthcheck.py
chiaberry: Thank you for thinking about this, I was thinking this dag would just be in charge of checking "can I reach the endpoint", is the healthcheck healthy, but I get your observation that it isnt testing further down into the stack. 

This is the example that came up when looking up the airflow way to see that a site is up and running.

johnclary: I assume it's not possible for the current API code it to return `200` without the JSON response along with it, but I agree with Frank—inspecting the JSON response would strengthen this and future proof it in case the behavior changed.

dags/citybase_healthcheck.py
johnclary: Is the implication that this health check would not verify the SSL cert?

johnclary: I misread your comment, but I am still wondering if we're certain that this will throw on an SSL error, and what error it will throw.

johnclary: Sorry for so much noise.  I see now that [the docs](https://airflow.apache.org/docs/apache-airflow-providers-http/stable/_api/airflow/providers/http/sensors/http/index.html) mention that you can pass `extra_options` as params to the `requests` library. nice! Two things come to mind:

* `timeout` - some timeout setting (30 seconds?). I think the request might hang forever without a timeout setting.
* `verify=True` - default behavior but make it clear that we want to check SSL?

mddilley: i was having similar thoughts because i think the intent is to verify SSL and to verify that the endpoint is healthy. 

100% beyond the scope of this issue but it is also making me wonder if we would want to hand this DAG a list that includes some of our other endpoints since we are DIYing some of the functionality of a service like [Uptime](https://uptime.com/). Could be reusable!

frankhereford: I was initially concerned that it might be possible to get back a 200, but after a little research, I feel pretty 👍 that this will catch SSL/TLS errors because if TLS fails to establish a connection, there isn't a connection for a HTTP response code to come back over, 200 or not. 

We could generate a self signed cert and stick it in there to see what happens if y'all wanted -- might be interesting, but maybe overkill too -- i'm not sure which.

chiaberry: While I was researching how to do this, I did come across a post saying that if the cert is expired, this dag would fail. I wish I had kept that link so I could share here, but I will go ahead and be explicit in saying `verify=True` even though its the default and doing that anyway. 

The dag has a timeout of 30 seconds, i think the request will be closed if the dag times out, but i will also add it to the parameters.

chiaberry:
I added the 30 second timeout to the request, the task timed out right before it hit that, but the either way the task wont run forever if the healthcheck fails
Screenshot 2026-06-15 at 9 06 56 AM

dags/citybase_healthcheck.py
mddilley: +1 increasing the frequency seems totally reasonable to me. i'm thinking that having a smaller window would benefit us during office hours but not as much outside that since we would trace back to the first Slack message to tell when the service went down

dags/citybase_healthcheck.py
```diff
@@ -0,0 +1,48 @@
...
+    catchup=False,
+) as dag:
+
+    dag.byline = f"Citybase postback healthcheck failed, {slack_member_ids['Chia']}"
+
+    check_citybase_endpoint = HttpSensor(
+        task_id="check_citybase_postback_endpoint",
+        http_conn_id="http_default",
+        endpoint="citybase.austinmobility.io",
+        request_params={"verify": True, "timeout": 30},
```
mddilley: @chiaberry i think these might need to be passed as `extra_options` https://airflow.apache.org/docs/apache-airflow-providers-http/stable/_modules/airflow/providers/http/sensors/http.html#HttpSensor.extra_options.

this PR is a cool TIL about HttpSensor 🙏

mddilley (commented):
I left one comment about the request params might needing to be passed in a different place. Sorry for letting this one hang around for so long! 🙈

dags/citybase_healthcheck.py
chiaberry: Thank you for catching this, you are so right.

chiaberry: Which I see John correctly mentioned this when he suggested it and I flubbed it somewhere along the way

dags/citybase_healthcheck.py
```diff
@@ -0,0 +1,48 @@
...
+) as dag:
+
+    dag.byline = f"Citybase postback healthcheck failed, {slack_member_ids['Chia']}"
+
+    check_citybase_endpoint = HttpSensor(
+        task_id="check_citybase_postback_endpoint",
+        http_conn_id="http_default",
+        endpoint="citybase.austinmobility.io",
+        extra_options={"verify": True, "timeout": 30},
+        response_check=lambda response: response.json().get("status") == "OK",
```
johnclary, line 46: i noticed that if i modify the response check so that it fails, the DAG eventually times out but we never see the response check error message. Is that desirable? 

I didn't see a way to have the HTTPSensor throw on error, because it does seem to be designed to just keeping poking at a regular interval. I think the use case for this operator is to actually block later tasks until an endpoint is awake 😴 

The `HttpOperator` is an alternative thing that will one-shot try and fail. Setting it up with HTTPs is kind of annoying though—you have to configure HTTPs in the connection settings itself. So here's what that looked like for me.

```python
    check_citybase_endpoint = HttpOperator(
        task_id="check_citybase_postback_endpoint",
        http_conn_id="citybase_https",
        endpoint="/",
        method="GET",
        extra_options={"verify": True, "timeout": 1},
        response_check=lambda response: response.json().get("status") == "OK",
    )
```

Image

Is this more in line with what we want?

johnclary (commented):
The changes look great but I have started to have doubts about this operator.

dags/citybase_healthcheck.py
chiaberry: >i noticed that if i modify the response check so that it fails, the DAG eventually times out but we never see the response check error message. Is that desirable?

So the DAG does fail, but the issue is that it won't have a helpful error message? What error message does the HttpOperator show?

I just want something that will alert us if the citybase postback is down, but I'd rather is try once and just tell us something is amiss instead of spending 30 seconds poking something that isnt working, so I'm inclined to modify the dag to use the HttpOperator. Thanks for digging into this so much.

johnclary: The HttpOperator returns `AirflowException: Response check returned False.`([example](https://austininnovation.slack.com/archives/C013G4RNJ01/p1782751135798169)) — which is at least marginally better than the task timeout error. that's problably as good as we can get without writing a little python operator instead.

chiaberry: Out of curiosity, what do you mean by "I didn't see a way to have the HTTPSensor throw on error" ? when I was testing the previous [iteration of this ](https://github.com/cityofaustin/atd-airflow/pull/358/changes/dae4d1ec0a3d251adcdbcbc8faa122fb6a489f44), I definitely had the task fail if the url was incorrect / or if I wasnt able to reach the healthcheck.

johnclary: I meant that I didn't see a way make the `response_check` failures throw an error.

johnclary (approved):
Working good with the HttpOperator! Thanks for going through the multiple iterations. Will be great to have this in place 🚢

frankhereford (approved):
The DAG looks good, thanks Chia. I spun it up and got a successful run using your instructions to test if production citybase was up ✅ and a ❌ run when I put in an entry for `citybase.austinmobility.io` with a bad IP in my `/etc/hosts`. I think this is 🚢 🚢 🚢.

mddilley (approved):
Very cool to see where this landed! I was able to test with bad SSL, changing the port, pointing to a url that timed out, and updating the DAG to test a status that doesn't exist. Failures across the board just liked we'd want! Thank you, Chia! 🚀 🙌

PR #361, opened 2026-06-22
Adds `--no-files-pass` flag to CAD incident import DAG

johnclary:
## Associated issues

* https://github.com/cityofaustin/atd-data-tech/issues/29006
* See also: https://github.com/cityofaustin/vision-zero/pull/2064

This PR fixes the problem of the various tasks of this ETL being blocked when no files are available to process. 

For example, if the first task (incidents_to_s3) succeeds but the incident import fails, it will not be possible to re-trigger this because the first task will fail when it does not find new files to process.

This update makes use of the `--no-files-pass` flag so that this DAG is effectively idempotent and can be triggered as needed at any time and all tasks will run.

## Associated repo

- vision-zero

## Testing

**Steps to test:**

1. Edit L74 of this DAG by replacing "/your/path/here" to the absolute path to your vision zero repo's CAD test data folder:

```
/your/path/to/vision-zero/etl/cad_incidents_import/test_data
```

2. If you have any CSV files in your test data folder, remove them.

3. Start Airflow and trigger this DAG. Enable the new **No files, no problem** toggle in the trigger modal before you proceed.

4. Observe that the `cad_incidents_to_s3` and `cad_incident_import` succeed despite the logs showing there are no files to process.

✅ 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/vz_cad_import.py
+++ b/dags/vz_cad_import.py
@@ -103,6 +103,17 @@ def get_incident_link_limit(params):
         return ""
 
 
+@task(
+    task_id="get_no_files_pass",
+)
+def get_no_files_pass(params):
+    """Return ` --no-files-pass` if the no_files_pass param has been set"""
+    if bool(params["no_files_pass"]):
+        return " --no-files-pass"
+    else:
+        return ""
+
+
 @dag(
     dag_id="vz-cad-incidents-import",
     description="A DAG which imports CAD records into the Vision Zero database.",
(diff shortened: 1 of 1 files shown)


mddilley (approved):
Tested locally with no files and no problem! ✅ ✅ ✅

chiaberry (approved):
no files, no problem!

PR #362, opened 2026-06-24
Rework DockerOperatorWithFallback and test DAG

### LINKED ISSUE #25196: Create task to handle Docker image pulls and handle outage
This task would handle Docker pulls instead of the current our current parameter, `force_pull=True`, which does not handle errors reaching Docker Hub gracefully.

https://austininnovation.slack.com/archives/CHZE6BC6L/p1760967953524189
https://austininnovation.slack.com/archives/CHZE6BC6L/p1760983685492079

frankhereford:
## Note

This PR contains a new DockerOperator wrapper named DockerOperatorWithFallback. It also contains a test DAG which can be used to test the wrapper. This PR closes and replaces #341 because it implements the same feature in a less brittle, more understandable way. 

🤖 This set of changes was created using Claude Code and was heavily influenced by the first attempt in #341 and even more so by the feedback that that PR received. 
✅ I think the code in this PR is of reasonable complexity and can be worked on without a requirement of AI tooling.

I believe this new PR is worth starting the review process over with because it is both much easier to read and understand. I feel it more completely addresses the concerns and advice given in the last PR's reviews.

## Refactoring the DAGs

This PR works with the existing refactor PR, no changes needed. That existing refactor PR is #342.

## Associated issues

This PR closes https://github.com/cityofaustin/atd-data-tech/issues/25196.

## Associated repo

Airflow Itself

## Testing

The goal here is to test the behavior of the docker wrapper in the four combinations of conditions:

* Having, or not having, a cached copy of an image you want to run
* If there is, or is not, a transient error preventing you from communicating with docker hub

Examples of transient/temporary errors are network problems, 500s from docker hub, it being unable to respond to a request when reached, etc. An non-example is when docker hub says "400 - image does not exist" -- that's a permanent error that should have an exception raised. 

### Testing Operations

Here's some stuff you're going to need to be able to do during the test.

Mask or inhibit docker hub from being reachable:

```
sudo sh -c 'echo "127.0.0.1 registry-1.docker.io" >> /etc/hosts'
```

Undo or uninhibit docker hub from being reachable, making it reachable again:

```
sudo sed -i '' '/registry-1.docker.io/d' /etc/hosts
```

Cache the test image:

```
docker pull python:3-slim

# Having a DAG's task pull the image caches it as well, BTW
```

Remove cache of docker image:

```
docker rmi python:3-slim
```

#### Testing Information

image

☝️ The test DAG will detect those two run time conditions and run appropriate tasks based on what it finds as shown in this graph.

#### Test Steps

Now that you have the background and the tools to set the state on your machine, work through the numbered list below.

The result of each testing instruction row should match the image where the columns match up to the numbered tasks. Be sure to click on the individual tasks of each tests and read the logs for a complete understanding of what happened. The `check_image_cached` and `check_registry_reachable` tasks' logs will tell you what the DAG detected for each of those run time conditions.

image

1) Remove image from local cache in case you just happen to have it laying around and, do not inhibit docker hub DNS, then trigger DAG.
2) Trigger DAG again. You're testing now that you have it cached.
3) Inhibit the docker hub DNS via /etc/hosts. Trigger the DAG.
4) Remove the image from local cache, and leave DNS inhibited. Trigger the DAG.
5) Remove the masking entry from your /etc/hosts. The image remains uncached. This is the same test as 1. Trigger the DAG.
6) You've got it cached now, so trigger it and expect the same results as 2. 

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/test_docker_operator_wrapper.py
+++ b/dags/test_docker_operator_wrapper.py
@@ -10,7 +10,7 @@ _TEST_IMAGE = "python:3-slim"
 doc_md = """
 ## DockerOperatorWithFallback test DAG
 
-Hint: This is much easier to read by closing this modal and examining the code tab below.
+Hint: This documentation is much easier to read by closing this modal and examining the code tab below.
 
 ---
 
--- a/dags/test_docker_operator_wrapper.py
+++ b/dags/test_docker_operator_wrapper.py
@@ -10,6 +10,10 @@ _TEST_IMAGE = "python:3-slim"
 doc_md = """
 ## DockerOperatorWithFallback test DAG
 
+Hint: This is much easier to read by closing this modal and examining the code tab below.
+
+---
+
 Validates all failure modes of `DockerOperatorWithFallback`
 (`dags/utils/docker_operator_with_fallback.py`).
 
--- a/dags/test_docker_operator_wrapper.py
+++ b/dags/test_docker_operator_wrapper.py
@@ -1,44 +1,86 @@
-from airflow.sdk import DAG
+import docker
+import requests
+from airflow.sdk import DAG, task
 from pendulum import datetime, duration
 
 from utils.docker_operator_with_fallback import DockerOperatorWithFallback
 
+_TEST_IMAGE = "python:3-slim"
+
 doc_md = """
 ## DockerOperatorWithFallback test DAG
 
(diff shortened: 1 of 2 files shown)


frankhereford:
Thanks y'all for your help and reviews. This pattern didn't work out due to complications around authentication. I am closing this and hope to solve this issue with https://github.com/cityofaustin/atd-airflow/pull/368 instead.

PR #365, opened 2026-07-08
Signal analytics orchestration 

Charlie-Henry:
## Associated issues
[#28432](https://github.com/cityofaustin/atd-data-tech/issues/28432) - Extract the movement column for the INRIX data in Datahub

## Associated repo
https://github.com/cityofaustin/dts-traffic-signal-metrics

## Testing

**Steps to test:**
1. You do NOT need to be on VPN
2. `docker compose up`
3. Go to the DAG called `dts_traffic_signal_metrics`
4. Try running it with the dry run parameter toggled on
5. Try running it without the dry run parameter toggled
6. Make sure everything ran successfully

---

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/dts_traffic_signal_metrics.py
+++ b/dags/dts_traffic_signal_metrics.py
@@ -1,0 +1,150 @@
+from os import getenv
+
+from airflow.sdk import DAG, task, Param
+from airflow.providers.docker.operators.docker import DockerOperator
+from pendulum import datetime, duration
+
+from utils.onepassword import get_env_vars_task
+from utils.slack_operator import task_fail_slack_alert
+
+doc_md = """
+## Traffic Signal Metrics ETL
+
+Downloads signal metrics from INRIX and uploads them to two socrata datasets:
+- [Signal metrics](https://datahub.austintexas.gov/dataset/INRIX-Signal-Metrics/bfmq-ijru)
+- [Signal metrics by movement](https://datahub.austintexas.gov/dataset/INRIX-Signal-Metrics-by-Movement/8qqy-h6xg/about_data)
+
+This ETL is scheduled to get the metrics from the last time this DAG was successful minus three days to today.
+
+Default behavior without any run history will get metrics for the last three days.
+
+### Troubleshooting
+
+The most common issue with this script might be periodic timeouts with Socrata as this DAG is upserting ~100k records per date.
+
+If that happens, a retry will usually be successful. Check [socrata status page](https://status.socrata.com/) as well.
+
+If there is a large backlog of dates to process you may need to manually remove the run history of this DAG to get it to complete before the 45 minute timeout.
+
+This DAG is configured to run daily so that should not happen...
+
+"""
+
+DEPLOYMENT_ENVIRONMENT = getenv("ENVIRONMENT", "development")
+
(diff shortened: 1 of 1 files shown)


dags/dts_traffic_signal_metrics.py
```diff
@@ -0,0 +1,151 @@
...
+
+@task(task_id="get_start_date")
+def get_start_date(**context):
+    # Returns the start date of this ETL. If there is no run history it will get the date 3 days in the past.
+    # If there is run history it will get the date it was last run succesfully, minus 3 days.
+    from pendulum import now
+
+    prev_start_date = context.get("prev_start_date_success") or now()
+    prev_start_date = prev_start_date.subtract(days=3)
+    return prev_start_date.strftime("%Y-%m-%d")
```
Charlie-Henry, line 98: Important date logic. 

Basically every time this will run it will fetch the last 3 days (or more accurately 3 days minus the last successful run date).

I'm not super sure what time INRIX posts their daily metrics updates so this should give us enough buffer to capture the most recent data without losing any history. If there is some downtime it should also catch back up with the `prev_start_date_success`

dags/dts_traffic_signal_metrics.py
```diff
@@ -0,0 +1,151 @@
...
+        ),
+    },
+) as dag:
+    docker_image = "atddocker/dts-traffic-signal-metrics:production"
+
+    env_vars = get_env_vars_task(REQUIRED_SECRETS)
+
+    start_date = get_start_date()
+
+    dry_run_arg = get_is_dry_run_arg()
```
mddilley, line 135: the DAG worked great for me and never had problems getting the right arg, but should this task be included in the task order declared at the end of the DAG definition?

i can't remember if this is one of those race conditions that we've run into in production or not

mddilley (approved):
Tested locally in both modes, and I saw it run just as expected. 🙏 🚢

dags/dts_traffic_signal_metrics.py
```diff
@@ -0,0 +1,151 @@
...
+    default_args=DEFAULT_ARGS,
+    schedule="30 5 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
+    tags=["repo:dts-traffic-signal-metrics", "socrata", "inrix", "signals", "metrics"],
+    catchup=False,
+    params={
+        "dry_run": Param(
+            title="Dry run",
+            default=False,
+            type="boolean",
+            description_md="Applies the dry-run flag. Only tests authenticating with the INRIX/Socrata APIs",
```
johnclary, line 125: Nice detail 🙏

dags/dts_traffic_signal_metrics.py
johnclary: Yep, i believe the bug is that Airflow doesn't auto-detect the dependency in string interpolation.

Image

johnclary (approved):
worked for me! +1 what Mike observed about the missing dependency in the DAG flow.

dags/dts_traffic_signal_metrics.py
Charlie-Henry: nice thanks for catching that. definitely an annoying bug

PR #367, opened 2026-07-09
Adds DAG for the VZ incident ETL 

johnclary:
## Associated issues

* Closes https://github.com/cityofaustin/atd-data-tech/issues/29289

## Associated repo

* vision-zero

## Testing

1. Checkout VZ `main`, start the VZ stack and apply migrations + metadata
2. Start airflow. Trigger this DAG with a **Dry run** and make sure to use the **Incident link limit** param when you trigger the DAG. `100` should be good.
3. Trigger the DAG without the dry run flag. Make sure to still use the **Incident link limit** param again.

✅ ✅ ✅ ✅ ✅ 

---

## Before we turn on this DAG

I need to run a manual backfill (similar to https://github.com/cityofaustin/atd-data-tech/issues/28941). I think I might truncate the whole VZ incidents table (purging the already-matched CAD records) and start with a clean slate 🤔 

#### Ship list

- [ ] Code reviewed

### DIFF
--- a/dags/vz_incidents.py
+++ b/dags/vz_incidents.py
@@ -76,7 +76,7 @@ def get_incident_link_limit(params):
 
 
 @dag(
-    dag_id="vz-cad-incidents-import",
+    dag_id="vz-incidents",
     description="A DAG which creates VZ incident records in the Vision Zero database.",
     doc_md=doc_md,
     # the CAD file export happens daily at 5a CT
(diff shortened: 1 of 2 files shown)


dags/vz_incidents.py
```diff
@@ -0,0 +1,157 @@
...
+    else:
+        return ""
+
+
+@dag(
+    dag_id="vz-incidents",
+    description="A DAG which creates VZ incident records in the Vision Zero database.",
+    doc_md=doc_md,
+    # the CAD file export happens daily at 5a CT
+    schedule="* 7 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
johnclary, line 83: This starts up the minute the CAD incident DAG will have timed out. 

In a future chunk of work I think it would be a good idea to bundle all the VZ ETLs into a single DAG, so the timing and dependencies are super clear. Right?

chiaberry (approved):
Image

ignore the first set of failures where i forgot to apply migrations to a fresh download of prod data!

dags/vz_incidents.py
mddilley: that sounds like a good plan to me! i have a vague memory of this being a wishlist item from before the revamp to help us tie a couple processes together.

mddilley: @johnclary should this be `0 7 * * *`? i can't quite remember the context of the run but it seems to be long running. might be a typo

dags/vz_incidents.py
```diff
@@ -0,0 +1,157 @@
+from os import getenv
+
+from airflow.sdk import dag, task, Param
+from airflow.providers.docker.operators.docker import DockerOperator
+from docker.types import Mount
```
mddilley, line 5: 🍝 think it is just copypasta but wanted to flag this import and slack_member_ids not being referenced, and the doc_md not matching up with this DAG

mddilley (approved):
Tested this and I am now part of the club of failing the first run - forgot to pull the latest from `main` 🙈 It worked fine after that! I commented on a few things that don't affect this working - just some small details and a question about the schedule. Thanks John! 🚢 
Screenshot 2026-07-09 at 6 28 38 PM

dags/vz_incidents.py
johnclary: yes! thank you for catching that!! i'm also going to take another look at our other ETL schedules and see if we can shift this all a bit earlier in the day. We're bumping into Xavier's working hours at this point. See also https://github.com/cityofaustin/atd-data-tech/issues/29314

🙏

frankhereford (approved):
I did the same thing as the other reviewers - switched to main but didn't pull! But, once I got my setup right, this ran just like I expected, and the code looks great. Ship it! Thank you John.

johnclary:
Closing this! https://github.com/cityofaustin/atd-data-tech/issues/30060

PR #346, opened 2026-05-14
Refactor DAGs to use DockerOperatorWithFallback

frankhereford:
## Seems Familiar?

This is a second-try PR following in the footsteps of https://github.com/cityofaustin/atd-airflow/pull/342. We had reviewed a different approach (#1, see [here](https://austininnovation.slack.com/archives/CHZE6BC6L/p1778685159614799?thread_ts=1778683746.321059&cid=CHZE6BC6L)) and it was impossible to tell what comments referred to what. 

## Associated issues

https://github.com/cityofaustin/atd-data-tech/issues/25196

## Associated repo

Airflow itself

## Testing

* Read the code, and ask yourself if you like this pattern
* Fire up a few DAGs, the more the merrier. You're testing when docker hub is up.
* Simulating a docker hub failure is tricky, I don't have a good suggestion on how.

---

#### Ship list

- [ ] Code reviewed
- [ ] Product manager approved
- [ ] Add note to 1PW secrets moved to API vault and check for duplicates

PR #368, opened 2026-07-13
DAG to pull images on a schedule and remove `force_pull` from our DAG pattern

frankhereford:
## Associated issues

## Previous work on this issue

This PR is a follow on to a series of PRs where I have attempted to write a `DockerOperatorWithFallback` operator which would trap exceptions gracefully for upstream issues at DockerHub or other network events. Apache's DockerOperator treats authentication in an unconventional way, and it complicates making a subclass significantly. I don't think it's reasonably possible with Airflow's current operator to accomplish the task without taking on responsibility for way more of the operator than we want. 

I'm going to close the following: https://github.com/cityofaustin/atd-airflow/pull/362, https://github.com/cityofaustin/atd-airflow/pull/346, & https://github.com/cityofaustin/atd-airflow/pull/341.

## What's in this PR

This is a much simpler solution to this problem. This is a DAG which pulls a list of images every 6 hours. And, If you need something pulled immediately, you can trigger this DAG and our airflow will be up to date. 

This PR also removes the `force_pull` configurations from all of other DAGs that used it.

## Associated repo

Airflow Itself

## Testing

* Fire up your local Airflow stack and trigger this DAG in dry run mode. 
  * You can check out that it correctly generates and serializes the task of pulling all the images in the list. 
  * You can also see the the dry-run logging.
* Spin it up again, this time for real, and observe that it pulls images, one by one. 
  * If you're on a mac, you'll see that we have a handful of tasks which fail, and that's because we don't bundle them with an ARM image. This is OK and expected.
  * You can also see here that if a pull fails, the DAG continues on

---

#### Ship list

- [ ] Code reviewed


dags/airflow_docker_image_pull.py
```diff
@@ -0,0 +1,187 @@
...
+from airflow.sdk import Param, chain, dag, task
+from pendulum import datetime, duration
+
+from utils.slack_operator import task_fail_slack_alert
+
+DEPLOYMENT_ENVIRONMENT = getenv("ENVIRONMENT")
+logger = logging.getLogger(__name__)
+
+# Docker images used by our DAGs, kept warm so task runs don't have to pull on demand
+DOCKER_IMAGES = [
```
frankhereford, line 19: I made the list initially with a 🤖, but I went through by hand correlating all of DAGs to which images we have in this list. So, the comments of which DAGs belong to which image are done by hand as an audit of the 🤖 work.

dags/airflow_docker_image_pull.py
```diff
@@ -0,0 +1,187 @@
...
+    "atddocker/vz-run-sql:production",
+    
+    # vz_moped_component_crashes.py, vz_socrata_export.py
+    "atddocker/vz-socrata-export:production",
+]
+
+
+@dag(
+    dag_id="airflow_docker_image_pull",
+    schedule="0 */6 * * *" if DEPLOYMENT_ENVIRONMENT == "production" else None,
```
frankhereford: I felt like 4X a day was pretty frequent but not over-the-top. Do you all think it's too much - should we do daily?

dags/cctv_service_healthcheck.py
```diff
@@ -46,7 +46,7 @@ def healthcheck(env_vars):
     try:
         res = requests.get(url, timeout=10)
     except requests.exceptions.RequestException as e:
-        # A ConnnectionError would raise here, for example
+        # A ConnectionError would raise here, for example
```
frankhereford, line 49: I merged `production` into the branch to get up to date and found a typo, this is a tag-along change.

dags/dts_row_reporting.py
```diff
@@ -140,7 +140,6 @@ def knack_services_task_template(task_id, image, command, env_vars, pull=False):
         command=command,
         environment=env_vars,
         tty=True,
-        force_pull=pull,
         mount_tmp_dir=False,
         trigger_rule="all_done",
```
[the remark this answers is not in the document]
frankhereford: Great call, https://github.com/cityofaustin/atd-airflow/pull/368/commits/2251f8c1e1ba714dfa77eaad75f46a3771febd06

dags/atd_knack_development_services.py
```diff
@@ -64,7 +64,6 @@ def knack_services_task_template(task_id, image, command, env_vars, pull=False):
         command=command,
         environment=env_vars,
         tty=True,
-        force_pull=pull,
         mount_tmp_dir=False,
         trigger_rule="all_done",
```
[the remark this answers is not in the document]
frankhereford: Great call, https://github.com/cityofaustin/atd-airflow/pull/368/commits/2251f8c1e1ba714dfa77eaad75f46a3771febd06

chiaberry (approved):
Nary a `force pull` left! Saw the reds and greens when running locally, and none of the failures blocked subsequent task runs. Ran it again a second time and as expected, it went much faster. Interestingly, I got some green squares on previous failed pulls and a failure on a previous green. 

What do you think about adding to the pull request template a checkbox to remind us to add new docker image names to the list?

dags/airflow_docker_image_pull.py
johnclary: Not a strong opinion, but I think daily would be fine.

dags/airflow_docker_image_pull.py
johnclary: Thanks for doing the audit! I'd prefer we don't try to continue to maintain the commented DAG names in this file—it seems like something that will just fall out of date. It's easy enough to search the codebase for an image name, so I don't think it's needed beyond this initial review.

dags/airflow_docker_image_pull.py
```diff
@@ -0,0 +1,187 @@
...
+            logger.info("Would pull %s", image)
+            return
+
+        logger.info("Pulling %s", image)
+        subprocess.run(["docker", "pull", image], check=True)
+
+    pull_tasks = []
+    for image in DOCKER_IMAGES:
+        # atddocker/atd-airflow:production -> pull_atd-airflow
+        image_name = image.split("/")[-1].split(":")[0]
```
johnclary: Might be better to just reach for a regex sub here to replace non-alphanumerics?

johnclary (approved):
This looks really good, Frank! Thank you for all your working getting us to this point.

I left some minor feedback and am curious what others think about maintaining the list of dependent DAGs in this code.

dags/airflow_docker_image_pull.py
mddilley: I think daily is good too. Feels like we'll get in the habit of manually triggering the DAG when launching new stuff so I think that schedule is frequent enough.

dags/airflow_docker_image_pull.py
```diff
@@ -0,0 +1,187 @@
...
+            type="boolean",
+            description_md="Log images that would be pulled without running docker pull.",
+        ),
+    },
+)
+def airflow_docker_image_pull():
+    """
+    Pulls the docker images used by our ETLs.
+
+    This DAG runs every 6 hours in production to keep local docker images
```
mddilley: Maybe you could remove the schedule reference here so that it doesn't fall out of sync if we need to adjust it in the future for some reason.

README.md
mddilley, line 1: Maybe it would be worth a section that explains the method of keeping images fresh with this DAG?

We kind of have similar things in https://github.com/cityofaustin/atd-airflow#utilities

dags/airflow_docker_image_pull.py
mddilley: Yeah, i'm thinking that it would probably fall out of sync too. I could totally imagine myself forgetting to remove a DAG from this list if it was decommissioned.

As for maintaining the list of images, I like Chia's suggestion to add an item to the PR template checklist. I'm sure we would find out soon enough if we forgot since it would complain loudly, but it would be a nice touch to have a reminder.

mddilley (approved):
Thanks Frank! I tested both modes locally, and I saw the expected output. 🚢

dags/airflow_docker_image_pull.py
frankhereford: These are great assessments and suggestions, I agree totally. I have removed the inline comments and added an entry to the PR checklist.

dags/airflow_docker_image_pull.py
frankhereford: Daily it is - thanks y'all. I set it for 6AM.

dags/airflow_docker_image_pull.py
frankhereford: Good call, thank you. I tightened up this comment.

dags/airflow_docker_image_pull.py
frankhereford: Thank you John. I think I have addressed your question, but please let me know if I didn't interpret it correctly. It reads now:

```
image_name = re.sub(r"[^a-zA-Z0-9-]", "", image.split("/")[1].split(":")[0])
```

README.md
frankhereford: Great idea. I added a section to the utilities that documents what the DAG does and how it fits in.

